PR #41, opened 2017-03-01
Missing spec tests

### LINKED ISSUE #21: Missing Spec Tests
## Fetching Data
- [x] Request for an empty collection should return an empty collection
> A server MUST respond to a successful request to fetch a resource collection with an array of resource objects or an empty array ([]) as the response document’s primary data.
```http
HTTP/1.1 200 OK

{
  "links": {
    "self": "http://example.com/articles"
  },
  "data": []
}
```
- [x] Relationship requests should return null if the relationship has not been set
```http
HTTP/1.1 200 OK

{
  "links": {
    "self": "http://example.com/articles/1/author"
  },
  "data": null
}
```

- [x]  Server MUST return 404 Not Found when processing a request to fetch a relationship link URL that does not exist.

> Note: This can happen when the parent resource of the relationship does not exist. For example, when /articles/1 does not exist, request to /articles/1/relationships/tags returns 404 Not Found.

- [x] Inclusion of multiple relationships:

> The value of the include parameter MUST be a comma-separated (U+002C COMMA, “,”) list of relationship paths. A relationship path is a dot-separated (U+002E FULL-STOP, “.”) list of relationship names.

- [x] 400 response if relationship does not exist

> If a server is unable to identify a relationship path or does not support inclusion of resources from a path, it MUST respond with 400 Bad Request.

- [x] Pagination links MUST appear in the links object that corresponds to a collection

> The following keys MUST be used for pagination links:
> - first: the first page of data
> - last: the last page of data
> - prev: the previous page of data
> - next: the next page of data
>
> Keys MUST either be omitted or have a null value to indicate that a particular link is unavailable.

## Creating, Updating and Deleting Resources

- [x] 403 Forbidden for client generated Ids

> A server MUST return 403 Forbidden in response to an unsupported request to create a resource with a client-generated ID.

- [x] Creation: The response SHOULD include a Location header identifying the location of the newly created resource.

- [x] Creation: The response MUST also include a document that contains the primary resource created.

- [x] 409: 

> A server MUST return 409 Conflict when processing a POST request in which the resource object’s type is not among the type(s) that constitute the collection represented by the endpoint.

- [x] 404:

> A server MUST return 404 Not Found when processing a request that references a related resource that does not exist.

## Updating Relationships

- [x] Can update to-many relationships through relationship link
- [x] Can update to-one relationships through relationship link

> A server MUST respond to PATCH, POST, and DELETE requests to a URL from a to-many relationship link as described below.

```http
PATCH /articles/1/relationships/tags HTTP/1.1

{
  "data": [
    { "type": "tags", "id": "2" },
    { "type": "tags", "id": "3" }
  ]
}
```

## Deleting

- [x] 404

> A server SHOULD return a 404 Not Found status code if a deletion request fails due to the resource not existing.

### LINKED ISSUE #40: Optional relationships
Hi. I have optional relationships in my models. 
```csharp
class Model : Identifiable<int>
{
  public override int Id {get;set;}
  public int? FooId {get;set;}
  public int? BarId {get;set;}
  public int? BazId {get;set;}

  public virtual A Foo {get;set;}
  public virtual B Bar {get;set;}
  public virtual C Baz {get;set;}
}
```
For every model only one of this relationships can exists. When i do query like this
```
http://localhost:5000/models?include=foo,bar,baz
```
i catch exception:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Object.GetType()
   at JsonApiDotNetCore.Builders.DocumentBuilder._getRelationship(Object entity, String relationshipName)
   at JsonApiDotNetCore.Builders.DocumentBuilder.<>c__DisplayClass8_0.<_addRelationships>b__0(Relationship r)
   at System.Collections.Generic.List`1.ForEach(Action`1 action)
   at JsonApiDotNetCore.Builders.DocumentBuilder._getData(ContextEntity contextEntity, IIdentifiable entity)
   at JsonApiDotNetCore.Builders.DocumentBuilder.Build(IEnumerable`1 entities)
   at JsonApiDotNetCore.Serialization.JsonApiSerializer._serializeDocuments(Object entity, IJsonApiContext jsonApiContext)
   at JsonApiDotNetCore.Formatters.JsonApiOutputFormatter.<WriteAsync>d__1.MoveNext()
```

As i understand, problem is here - https://github.com/Research-Institute/json-api-dotnet-core/blob/master/src/JsonApiDotNetCore/Builders/DocumentBuilder.cs#L209 cause `entity` is null.

Should there be check for null? I can make PR if needed.

jaredcnance:
Closes #21 
Closes #40

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,6 @@ dotnet: 1.0.0-preview2-1-003177
 branches:
   only:
     - master
+    - staging
 script:
   - ./build.sh
--- a/README.md
+++ b/README.md
@@ -76,9 +76,7 @@ Your models should inherit `Identifiable<TId>` where `TId` is the type of the pr
 
 ```csharp
 public class Person : Identifiable<Guid>
-{
-    public override Guid Id { get; set; }
-}
+{ }
 ```
 
 #### Specifying Public Attributes
@@ -89,8 +87,6 @@ add the `AttrAttribute` and provide the outbound name.
 ```csharp
 public class Person : Identifiable<int>
 {
-    public override int Id { get; set; }
-    
     [Attr("first-name")]
     public string FirstName { get; set; }
 }
@@ -104,8 +100,6 @@ they should be labeled as virtual.
 ```csharp
 public class Person : Identifiable<int>
 {
-    public override int Id { get; set; }
-    
     [Attr("first-name")]
     public string FirstName { get; set; }
 
@@ -119,8 +113,6 @@ For example, a `TodoItem` may have an `Owner` and so the Id attribute should be 
 ```csharp
 public class TodoItem : Identifiable<int>
 {
-    public override int Id { get; set; }
-    
     [Attr("description")]
     public string Description { get; set; }
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,7 +19,7 @@ deploy:
     secure: 6CeYcZ4Ze+57gxfeuHzqP6ldbUkPtF6pfpVM1Gw/K2jExFrAz763gNAQ++tiacq3
   skip_symbols: true
   on:
-    branch: master
+    branch: staging
 - provider: NuGet
   name: production
   api_key:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,7 @@ pull_requests:
 branches:
   only:
   - master
+  - staging
 nuget:
   disable_publish_on_pr: true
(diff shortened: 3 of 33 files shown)


jaredcnance:
Closed in favor of #42 

Will deploy to master when [P2](https://github.com/Research-Institute/json-api-dotnet-core/projects/2) is completed

Issue #47, opened 2017-03-04
Upgrade to use msbuild (csproj)


jaredcnance:
Wait until SDK reaches RTM (currently at RC4)

https://github.com/dotnet/core/tree/master/release-notes

PR #48, opened 2017-03-13
Use 204 on empty delete response

medokin:
See: http://jsonapi.org/format/#crud-deleting

### DIFF
--- a/src/JsonApiDotNetCore/Controllers/JsonApiController.cs
+++ b/src/JsonApiDotNetCore/Controllers/JsonApiController.cs
@@ -201,7 +201,7 @@ namespace JsonApiDotNetCore.Controllers
             if (!wasDeleted)
                 return NotFound();
 
-            return Ok();
+            return NoContent();
         }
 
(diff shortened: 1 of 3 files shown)


medokin:
No idea why the test is failing

jaredcnance:
@medokin thanks for the PR! I'll dig into it this afternoon.

jaredcnance:
the test you updated was intended to verify that the server returns a 404 if a `DELETE` request is received for an entity that doesn't exist. I've updated the appropriate tests.

Currently, the organization of the tests is less than desirable and I hope to change that in the near future

medokin:
Perfect, thanks. It works fine now. Would be easier to contribute if #47 was finished.

jaredcnance:
#51 is open to resolve #47. However, it looks like AppVeyor has yet to upgrade their version of the SDK. Once that is fixed, I'll merge into master. Until then, if you'd like to contribute, you can work off the `develop` branch

Issue #53, opened 2017-03-15
PostAsync entity id check fails for Guid type


jfhs:
Since `Guid` type does not have null value, it is always initialized to `Guid.Empty` and [this check](https://github.com/Research-Institute/json-api-dotnet-core/blob/master/src/JsonApiDotNetCore/Controllers/JsonApiController.cs#L142) in `JsonApiController` fails (string value of Guid.Empty is "000-000...").

Looking around I was not able to find any generic way to check for empty values. So I would suggest adding method `IsIdEmpty` or something similar to `IIdentifiable` with default implementation maybe as you have it right now, or simply checking for null value. Then users could implement that method with whatever logic their ids have.

If that seems reasonable for you, I might make a pull request with that change tomorrow.

PR #54, opened 2017-03-15
fix: non-generic id check in PostAsync

jfhs:
Proposed fix for #53

### DIFF
--- a/src/JsonApiDotNetCore/Models/Identifiable.cs
+++ b/src/JsonApiDotNetCore/Models/Identifiable.cs
@@ -12,5 +12,10 @@ namespace JsonApiDotNetCore.Models
             get { return Id; }
             set { Id = (T)value; }
         }
+
+        public bool IsIdEmpty()
+        {
+            return Id != null;
+        }
     }
 }
--- a/src/JsonApiDotNetCore/Controllers/JsonApiController.cs
+++ b/src/JsonApiDotNetCore/Controllers/JsonApiController.cs
@@ -137,9 +137,8 @@ namespace JsonApiDotNetCore.Controllers
                 _logger?.LogInformation($"Entity cannot be null returning 422");
                 return UnprocessableEntity();
             }
-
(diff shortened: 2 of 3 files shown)


src/JsonApiDotNetCore/Models/Identifiable.cs
```diff
@@ -12,5 +12,10 @@ object IIdentifiable.Id
             get { return Id; }
             set { Id = (T)value; }
         }
+
+        public bool IsIdEmpty()
```
jaredcnance: This will need to be virtual so users can override it. Also, since this won't work "out-of-the-box" for a lot of types. I'd like to add support for common, non-nullable types. I can work on this tonight.

jaredcnance:
@jfhs thanks for the PR! I've added a few comments. I can take it from here or you can keep working on it if you want?

jfhs:
@jaredcnance you are welcome :) However I've stumbled upon another problem, again related to Id type. I will describe it here, because I think solution for both problem is linked.

I got `InvalidCastException` in `DeSerializer` apparently because `string` can't be casted to `Guid`, which is called from [here](https://github.com/Research-Institute/json-api-dotnet-core/blob/master/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs#L63)

Given that and problem for checking id for empty/null, maybe a better solution would be to have methods like `SetIdFromString`/`GetIdAsString`, or just a property `StringId`, so that you can safely safely serialize/deserialize ids from string. Then in controller you would actually check this string representation for emptiness.

Also, I hurried a bit with PR, probably it would break your samples (at least one from the video) because for `ApplicationUser` you implemented interface, and now it has one more method.

I'm not sure I will be able to work any longer today, so if you want, you can just close my PR and do it yourself. If not, I will come back tomorrow, and fix it, because I need this working ASAP :)

jaredcnance:
Closing in favor of #55

Issue #52, opened 2017-03-14
HasMany relationship treated as HasOne in deserialization


jfhs:
Having a model
```c#
    public class Country: Identifiable<Guid>
    {
        [Attr("name")]
        public string Name { get; set; }

        [HasMany("cities")]
        public virtual List<City> Cities { get; set; }
    }
```
And trying to post
```json
{"data": {
    "attributes": {
      "name":"qwe"
    },
    "relationships": {
      "cities": {
        "data": []
      }
    },
    "type":"countries"
  }
}
```
`JsonApiDeSerializer` tries to find `CitiesId` property in [here](https://github.com/Research-Institute/json-api-dotnet-core/blob/master/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs#L101) which it of course doesn't.

I believe there should be separate code for handling HasMany relationships.

PR #55, opened 2017-03-15
Fix/id type support

### LINKED ISSUE #53: PostAsync entity id check fails for Guid type

jaredcnance:
Closes #53


jaredcnance:
@jfhs this **should** fix the issues. I will be writing tests to validate it in the morning. Let me know what you think

jfhs:
@jaredcnance Looks good. However I checked [specification](http://jsonapi.org/format/#crud-creating-client-ids), and it says that server may accept client generated ids which is explicitly forbidden in current implementation. That is not the same issue you are solving by this PR, but it requires change in same place, so maybe you can do it all at once.

I would suggest adding a flag in controller for disabling this verification, something like `bool AllowClientGeneratedId` which will be false by default, and can be re-enabled by users. Although that's not required for my current use-case, I think in general that would be good to have.

jaredcnance:
You're right in that a server MAY support client generated ids. I'll include a test this morning that checks whether or not it will work without much modification. I suspect this may not be so simple, but I'll dig into it.

src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs
```diff
@@ -98,29 +111,68 @@ private static object DataToObject(DocumentData data, IJsonApiContext context)
...
+
+                var newValue = data["id"];
+                var convertedValue = TypeHelper.ConvertType(newValue, entityProperty.PropertyType);
+                entityProperty.SetValue(entity, convertedValue);
             }
 
             return entity;
         }
 
+        private static object _setHasManyRelationship(object entity,
```
jaredcnance: Closes #52

jaredcnance:
You can get these updates on [MyGet](https://www.myget.org/feed/research-institute/package/nuget/JsonApiDotNetCore) (version 1.1.1-alpha1-0069)

PR #61, opened 2017-03-17
Feature/request meta

### LINKED ISSUE #56: Support for generic request meta
Currently the only (documented) way to add metadata is to implement `IHasMeta` in entity class. However, it might be good to have some generic way to add it to any request, regardless of the entity type, i.e. for pagination, performance metrics or some caching information.

I am not sure what would be the right place to have it, maybe as an option for `DocumentBuilder` somehow?

jaredcnance:
Closes #56

### DIFF
--- a/test/JsonApiDotNetCoreExampleTests/Acceptance/Extensibility/RepositoryOverrideTests.cs
+++ b/test/JsonApiDotNetCoreExampleTests/Acceptance/Extensibility/RepositoryOverrideTests.cs
@@ -57,7 +57,7 @@ namespace JsonApiDotNetCoreExampleTests.Acceptance.Extensibility
             // act
             var response = await client.SendAsync(request);
             var responseBody = await response.Content.ReadAsStringAsync();
-            var deserializedBody = JsonApiDeSerializer.DeserializeList<TodoItem>(responseBody, jsonApiContext, _fixture.GetService<AppDbContext>());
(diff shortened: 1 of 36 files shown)


test/JsonApiDotNetCoreExampleTests/Acceptance/Extensibility/RepositoryOverrideTests.cs
```diff
@@ -57,7 +57,7 @@ public RepositoryOverrideTests(DocsFixture<Startup, JsonDocWriter> fixture)
             // act
             var response = await client.SendAsync(request);
             var responseBody = await response.Content.ReadAsStringAsync();
-            var deserializedBody = JsonApiDeSerializer.DeserializeList<TodoItem>(responseBody, jsonApiContext, _fixture.GetService<AppDbContext>());
+            var deserializedBody = _fixture.GetService<IJsonApiDeSerializer>().DeserializeList<TodoItem>(responseBody);
```
jaredcnance: The only "breaking" changes here are Serialization/DeSerialization APIs. I would consider this a breaking change because it may be used in tests. This is a much needed change that helps to de-couple concrete types.

I implemented it in this PR because to allow a`IMetaRequest` service to be injected and used by the `DocumentBuilder` would require an extra call to `HttpContext.RequestServices.GetService<TService>()`.

These changes reduce the number of times we have to go back to the container, and define 3 distinct composition roots:

- `JsonApiInputFormatter`
- `JsonApiController`
- `JsonApiOutputFormatter`

PR #72, opened 2017-03-23
Fix #71 Hard dependency to MVC setup

### LINKED ISSUE #71: Hard dependency to MVC setup
Currently, MVC is setup inside the JSONAPI Framework's own calls and there is no possibility to customize the MVC setup e.g. with custom options. There should be a possibility to customize the calls or do them on one's own.

Particularly: [IApplicationBuilderExtension](https://github.com/Research-Institute/json-api-dotnet-core/blob/2c7d07f58f408c8081cbeb271b1b8fcce05db6b1/src/JsonApiDotNetCore/Extensions/IApplicationBuilderExtensions.cs#L12) and [IServiceCollectionExtensions](https://github.com/Research-Institute/json-api-dotnet-core/blob/2c7d07f58f408c8081cbeb271b1b8fcce05db6b1/src/JsonApiDotNetCore/Extensions/IServiceCollectionExtensions.cs#L36-L40)

... working on it...

JanMattner:
Offers decoupling of the dependency to MVC setup in IApplicationBuilderExtensions. Default behavior is still to use MVC.

### DIFF
--- a/src/JsonApiDotNetCore/Extensions/IApplicationBuilderExtensions.cs
+++ b/src/JsonApiDotNetCore/Extensions/IApplicationBuilderExtensions.cs
@@ -10,9 +10,7 @@ namespace JsonApiDotNetCore.Routing
             app.UseMiddleware<RequestMiddleware>();
 
             if (useMvc)
-            {
                 app.UseMvc();
-            }
 
             return app;
         }
--- a/src/JsonApiDotNetCore/Extensions/IApplicationBuilderExtensions.cs
+++ b/src/JsonApiDotNetCore/Extensions/IApplicationBuilderExtensions.cs
@@ -5,11 +5,14 @@ namespace JsonApiDotNetCore.Routing
 {
     public static class IApplicationBuilderExtensions
     {
-        public static IApplicationBuilder UseJsonApi(this IApplicationBuilder app)
+        public static IApplicationBuilder UseJsonApi(this IApplicationBuilder app, bool useMvc = true)
         {
             app.UseMiddleware<RequestMiddleware>();
 
-            app.UseMvc();
+            if (useMvc)
+            {
+                app.UseMvc();
+            }
 
             return app;
         }
--- a/src/JsonApiDotNetCore/JsonApiDotNetCore.csproj
+++ b/src/JsonApiDotNetCore/JsonApiDotNetCore.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <VersionPrefix>1.2.0</VersionPrefix>
+    <VersionPrefix>1.2.2</VersionPrefix>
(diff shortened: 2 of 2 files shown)


src/JsonApiDotNetCore/Extensions/IApplicationBuilderExtensions.cs
```diff
@@ -5,11 +5,14 @@ namespace JsonApiDotNetCore.Routing
...
     public static class IApplicationBuilderExtensions
     {
-        public static IApplicationBuilder UseJsonApi(this IApplicationBuilder app)
+        public static IApplicationBuilder UseJsonApi(this IApplicationBuilder app, bool useMvc = true)
         {
             app.UseMiddleware<RequestMiddleware>();
 
-            app.UseMvc();
+            if (useMvc)
+            {
```
jaredcnance: at this point I would question whether or not it should be the responsibility of this library to even call `.UseMvc()`

JanMattner: I think so too. However, that would be a breaking change.

jaredcnance: Thanks for the PR. I'm fine with the proposed solution. I'd just like to see if there is maybe a better one.

Since the library won't even work without the MVC middleware, what do you think of just adding an overload with a `Action<IRouteBuilder> configureRoutes` parameter which we can pass into the [UseMvc method](https://github.com/aspnet/Mvc/blob/d2430f0f4ef869762834f794ec30662adf2c5232/src/Microsoft.AspNetCore.Mvc.Core/Builder/MvcApplicationBuilderExtensions.cs#L64)?

JanMattner: Yes, I thought about that too. Currently, that is the only available overload for UseMvc, so that would be fine. As soon as there is another overload, you'd need to make this available as well...

Furthermore, I think it matters in which order the Use-Calls occur, so to be completely flexible, another `Action<IApplicationBuilder>` would be necessary that is called between the `app.UseMiddleware<RequestMiddleware>()` and `app.UseMvc()` statements.

jaredcnance: @JanMattner all good points. I can't think of any circumstances where someone would need to apply middleware in between the two, but that doesn't mean there isn't a use case. That would result in a method signature like:

```csharp
IApplicationBuilder UseJsonApi(this IApplicationBuilder app, 
  Action<IRouteBuilder> configureRoutes, 
  Action<IApplicationBuilder> intermediateAppBuilderCalls);
```

which I can't say I'm a huge fan of. I don't think we should further couple it to the `MvcApplicationBuilder` api. Ideally I'd like to see this deprecated in the future. But, I think this PR is a good start.

PR #88, opened 2017-04-07
Feat/allow non dasherized routes

### LINKED ISSUE #86: Add possibility to add custom non-JSONAPI controllers
Currently, via the `DasherizedRoutingConvention`, all controllers are considered as JSONAPI controllers and the routing convention is applied making it impossible to use that convention besides other non-JSONAPI controllers with custom attribute routing.

It should be possible to explicitly define which controllers are considered as JSONAPI controllers. Easiest way would be a simple attribute that's attached to a controller and the `DasherizedRoutingConvention` applies the convention only to those controllers with the attribute.

jaredcnance:
Closes #86 
- ~Still needs tests~

Limitations:

- Limits the use of the dasherized route convention to JsonApiControllers
- All JsonApiController routes will be dasherized (to enable the opposite would require some extra work and breaking changes)


JanMattner:
The changes look fine, but at least one test with a custom non-dasherized controller should be added.

JanMattner:
Great! I think that is sufficient for a first extensibility for custom controllers.

PR #89, opened 2017-04-08
De-Couple DAL from Entity Framework

### LINKED ISSUE #84: RFC: De-Couple data access from EF
### Problem
Currently, the controller leans on `IEntityRepository` directly which is tightly coupled to `IQueryable`. However, this doesn't make much sense for ORMs, like Dapper, that do not use `IQueryable`. Also, it is entirely possible that resources may be retrieved from non SQL storage solutions, such as another HTTP API or RPC.

**Tightly Coupled Areas:**
- [JsonApiController](https://github.com/Research-Institute/json-api-dotnet-core/blob/master/src/JsonApiDotNetCore/Controllers/JsonApiController.cs)
- [ContextGraph](https://github.com/Research-Institute/json-api-dotnet-core/blob/master/src/JsonApiDotNetCore/Internal/ContextGraph.cs#L9)
- [ContextGraphBuilder](https://github.com/Research-Institute/json-api-dotnet-core/blob/master/src/JsonApiDotNetCore/Internal/ContextGraphBuilder.cs)
- [IServiceCollectionExtensions](https://github.com/Research-Institute/json-api-dotnet-core/blob/master/src/JsonApiDotNetCore/Extensions/IServiceCollectionExtensions.cs)
- [Internal.Generics](https://github.com/Research-Institute/json-api-dotnet-core/tree/master/src/JsonApiDotNetCore/Internal/Generics)

### Proposed Solution
1. Extract controller logic into `IResourceService`. Reduces controller methods to something like the following. Would allow any intermediate service to perform the data access.

```csharp
public class JsonApiController<T, TId>
    : JsonApiControllerMixin where T : class, IIdentifiable<TId>
{
    private readonly IResourceService<T, TId> _service;

    public JsonApiController(IResourceService<T, TId> resourceService)
    {
        _service = resourceService;
    }

    [HttpGet]
    public virtual async Task<IActionResult> GetAsync()
    {
        var entities = await _service.GetAsync();
        return Ok(entities);
    }
}
```

2. Refactor the context graph: 
 - Rename `ContextGraphBuilder` to `EntityContextGraphBuilder`, since it uses the EF `DbContext`.
 - Can `ContextGraph`<T> be used without the `DbContext` generic? If not, perform rename `Entity...`
 - Define API to either inject custom `IContextGraph` implementations or an API to build a generalized context graph

3. Refactor service extensions or add methods for using the new APIs defined by (2)

### Considerations

- Need to take a closer look at json:api operations and how this would fit in. 
  - https://github.com/json-api/json-api/issues/795
- Since this may introduce a variety of breaking changes, perhaps we should consider further de-coupling from the MVC middleware in this solution as well (see discussion in #71)?

jaredcnance:
Closes #84 

As written, this will break all controllers inheriting from `JsonApiController`. You would need to update your controller constructor parameters. One approach (as lightly discussed in #84) would be to provide a deprecated constructor method and possibly a `LegacyJsonApiController` that would contain the current logic. `JsonApiController` would then pass calls into the legacy controller if it were constructed using the `IEntityRepository` constructor instead of the `IResourceService` param. This would be removed at a later date. I would like to hear any thoughts on this and whether it is necessary, or if we should accept the breaking change. Also, does this constitute a major version change?

**Please feel free to tear this apart or raise any concerns/suggestions/complaints**

### TODO
- [x] Replace `IEntityRepository` with `IResourceService` in `JsonApiController`
- [x] Determine how best to maintain backwards compatibility, if at all
- [x] Acceptance tests to validate extensibility at the service layer
- [x] Custom ContextGraphBuilder API
- [x] Service Collection Extension (`services.AddJsonApi()`) without the `DbContext` generic
- [x] Complete documentation
- [x] Bump project version
- [x] Add `ResourceAttribute` to allow custom type names for DbSets


src/JsonApiDotNetCore/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -73,6 +80,12 @@ public static void AddJsonApi<TContext>(this IServiceCollection services, Action
             this IServiceCollection services,
             JsonApiOptions jsonApiOptions)
         {
+            if(!jsonApiOptions.ContextGraph.UsesDbContext)
+            {
```
jaredcnance: I really **really** don't like this, but as stated in the commit message:

> as far as I can tell, .Net's native DI will check all pathways for resolution even if they shouldn't be resolved. Specifically, the Internal.Generics classes require DbContext to be injected. However, they would never be resolved if the app does not actually use EF and implements their own IResourceService

Would love to see if anyone has any extra insight into this. I may open up an issue on the dotnet side.

src/NoEntityFrameworkExample/Startup.cs
```diff
@@ -34,7 +34,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddJsonApi(options => {
                 options.Namespace = "api/v1";
                 options.BuildContextGraph((builder) => {
-                    builder.AddResource<TodoItem>("todo-items");
+                    builder.AddResource<TodoItem>("TodoItems");
```
jaredcnance: Currently, this simulates the `DbSet` name. I don't really like the way this works, especially since the format provided is not even the public format. But, it will currently break in the de-serialization. Need to do some more work on this.

src/JsonApiDotNetCore/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -15,64 +15,86 @@ namespace JsonApiDotNetCore.Extensions
...
+
+        public static void AddJsonApiInternals(
+            this IServiceCollection services,
+            JsonApiOptions jsonApiOptions)
+        {
+            if(!jsonApiOptions.ContextGraph.UsesDbContext)
+            {
+                services.AddScoped<DbContext>();
+                services.AddSingleton<DbContextOptions>(new DbContextOptionsBuilder().Options);
+            }
```
jaredcnance: I really **really** don't like this, but as stated in the commit message:

> as far as I can tell, .Net's native DI will check all pathways for resolution even if they shouldn't be resolved. Specifically, the Internal.Generics classes require DbContext to be injected. However, they would never be resolved if the app does not actually use EF and implements their own IResourceService

Would love to see if anyone has any extra insight into this. I may open up an issue on the dotnet side.

jaredcnance:
@JanMattner @jfhs this is mostly ready. I'll spend some time this weekend adding extra tests around it. Wanted to get your opinions on it so far. I've updated the Readme, and there is a sample project that does not use EF.

JanMattner:
Thanks, great work! However, what first comes in my mind: If I want to use the IResourceService as an interface to some HTTP service, I won't be able to fulfill `Task<IEnumerable<T>> GetAsync();` simply because of the huge amount of items.

I need at least some kind of paging mechanism with pageSize and pageNumber.

jaredcnance:
@JanMattner you should be able to implement paging however you like. You can get the `IJsonApiContext` from your service constructor. From that, you can get the page query like so:

```csharp
var pageNumber = _jsonApiContext.PageManager.CurrentPage;
var route = $"{EXTERNAL_API_HOST}/items?page={pageNumber}";
var request = new HttpRequestMessage(new HttpMethod("GET"), route);
var response = await client.SendAsync(request);
```

Is this what you're talking about?

JanMattner:
My bad! Yes, that was what I was looking for. I'll have a closer look at the rest later and give feedback.

src/NoEntityFrameworkExample/Startup.cs
JanMattner: For the DbSets, the name is dasherized and there is no opt-out. Why are you not dasherizing here as well?

src/JsonApiDotNetCore/Configuration/JsonApiOptions.cs
```diff
@@ -6,5 +11,31 @@ public class JsonApiOptions
...
+            if(builder != null)
+                builder(contextGraphBuilder);
+
+            ContextGraph = contextGraphBuilder.Build();
+        }
+
+        public void BuildContextGraph(Action<IContextGraphBuilder> builder)
+        {
+            if(builder == null)
+                throw new ArgumentException("Cannot build non-EF context graph without a IContextGraphBuilder action", nameof(builder));
```
JanMattner: typo: "without a**n** IContextGraphBuilder"

src/JsonApiDotNetCore/Builders/ContextGraphBuilder.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                var dbSetType = property.PropertyType;
+
+                if (dbSetType.GetTypeInfo().IsGenericType
+                    && dbSetType.GetGenericTypeDefinition() == typeof(DbSet<>))
+                {
+                    var entityType = dbSetType.GetGenericArguments()[0];
+                    entities.Add(new ContextEntity
+                    {
+                        EntityName = property.Name.Dasherize(),
```
jaredcnance: I'm inclined to agree with allowing for an optional attribute per DbSet that would allow for a custom name. 

- When defining names manually (either through the ContextGraphBuilder or using a ResourceAttribute), whatever input is provided will be accepted as the resource name
- If no `[Resource("my-resource")]` is specified on the DbSet, then it will be dasherized by default

src/JsonApiDotNetCore/Builders/ContextGraphBuilder.cs
```diff
@@ -0,0 +1,114 @@
...
+            };
+            return graph;
+        }
+
+        public void AddResource<TResource>(string pluralizedTypeName) where TResource : class
+        {
+            var entityType = typeof(TResource);
+            Entities.Add(new ContextEntity
+            {
+                EntityName = pluralizedTypeName,
```
jaredcnance: See comment below. I'd like to avoid mutating any manual input so we can avoid inconsistencies as described in #93

src/JsonApiDotNetCore/Builders/ContextGraphBuilder.cs
JanMattner: That sounds good! And yes you are right, any manual explicit input should not be altered.

jaredcnance:
I think this is ready to go (other than the goofy DI stuff mentioned above -- but i dont think I want that to hold up a pre-release). Last call for comments. If there are no other comments, I'll merge this tomorrow afternoon. Thanks @JanMattner for your help!

JanMattner:
I have found still a buggy usage of the dasherization when creating links. I have added a couple of tests that currently fail.

jaredcnance:
@JanMattner thanks for the tests. So, the main issue revolves around [this line](https://github.com/Research-Institute/json-api-dotnet-core/blob/feat/decouple-dal/src/JsonApiDotNetCore/Builders/LinkBuilder.cs#L29). Where I made the assumption that APIs should use semantic endpoints (maybe the wrong name for it?). That is, a request to "/todo-items" will return "todo-items" and a request to "/custom-todo-items" should return "custom-todo-items" not "todo-items". This breaks the link builder as it assumes, the base route structure will end with the resource name. I'd be curious to hear if you currently need this support.

JanMattner:
Yes, you are right, in general a server implementation should follow the guidelines and provide semantic endpoints. However, the user is not bound to that and the current implementation is just another hidden dasherization. Furthermore, if I interpret the code correctly, all path segments will be dasherized. But what if I want to host (even my proper semantic) endpoint at:
https://host.com/my/customPath/todo-items

It would dasherize the whole path. I just think that this is not really intuitive if done hidden / implicitly.

jaredcnance:
Agreed that all path segments should not be dasherized. My [last commit](https://github.com/Research-Institute/json-api-dotnet-core/pull/89/commits/4480e6b7e64ac1a04a678e39ca71d4e40ea937fe) should have fixed this. 

The way the logic is intended to work is to identify the "base route/namespace" from the current route and build links from there:

```
                 resource-name |--------|
https://host.com/my/customPath/todo-items
|-----base-route-------------|
```

You're right that a user can define routes however they want and nothing prevents that. If someone defines a controller as:

```csharp
[Route("custom/route/blueberries")]
public class TodoItemsController : JsonApiController<TodoItem>
{
   // ...
}
```

There is no great way (that I am aware of) to determine "TodoItems" links should actually point to "Blueberries". In this scenario, the link builder assumes /blueberries is actually part of the base route and would build links like:

```
https://host.com/custom/route/blueberries/todo-items
```

I'm unsure how to proceed with this. Perhaps the link builder show throw if the resource cannot be found in the path? But I'm not convinced that we should support routing like this.

jaredcnance:
There is one other issue that we need to address and that's the `DasherizedRoutingConvention` that assumes all JsonApiControllers should have dasherized names.

```csharp
var template = $"{_namespace}/{controller.ControllerName.Dasherize()}";
```

I propose a few options:
- introduce a flag on `JsonApiOptions` that would allow users to disable the `DasherizedRoutingConvention` globally
- provide attributes that could decorate the controller to disable the routing convention
- check if the user has supplied a specific route through the `RouteAttribute` and if so, then disable the convention.

However, I think this is more about routing choices and would like to close this PR and punt routing into another. Although, to be clear, I'd like to see both of these go out in the 2.0.0 release.

jaredcnance:
@JanMattner i'm going to merge this into develop and move any additional routing concerns into a new PR

PR #92, opened 2017-04-11
v1.3.1

### LINKED ISSUE #85: Filter by attribute of related entity
Maybe this is possible already but it would be great to be able to filter by an attribute of a related entity
Maybe, have something like 
bookings?include=status&filter[starting-at]=le:2017-01-01&filter[status.name]=eq:active

### LINKED ISSUE #86: Add possibility to add custom non-JSONAPI controllers

jaredcnance:
Closes #85 
Closes #86


JanMattner:
I have added tests for the deeper inheritance case and fixed it with a simpler check for the subtype. I think that should be sufficient.

jaredcnance:
@JanMattner looks great! Thanks! 

I want to let this sit for another day or so to give anyone else a chance to comment before merging this in. Until then, these changes are available on MyGet.

PR #94, opened 2017-04-25
v2.0.0

### LINKED ISSUE #84: RFC: De-Couple data access from EF

### LINKED ISSUE #93: Inconsistency for dasherized attributes
If I have a client entity definition as follows:

```csharp
{
        [Attr("clientName")]
        public string ClientName { get; set; }
}
```

I expect that the attribute is everywhere referenced as "clientName". However, when POSTing a new client, I am forced to use the dasherized version of it:

```json
{
  "data": {
    "type": "clients",
    "attributes": {
      "client-name": "Foo Bar"
  }
}
```

However, the response of that and a normal GET request uses the non-dasherized version:

```json
{
  "data": {
    "type": "clients",
    "id": "8",
    "attributes": {
      "clientName": "Foo Bar"
    },
  }
}
```

I am forced to do that due to this [line in the JsonApiDeSerialization.cs](https://github.com/Research-Institute/json-api-dotnet-core/blob/master/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs#L98)

I expect that the attribute names defined via the `Attr` attribute are used throughout the framework.

### LINKED ISSUE #95: PATCH request with nullable attributes does not work
Hi, 
Thanks for this great library! 

When you are trying to PATCH a model with nullable attributes, it does not work. 

Model example: 

```csharp

public class TodoItem : Identifiable
{
    Attr("description")]
    public string Description { get; set; }

    [Attr("created-date")]
    public DateTime CreatedDate { get; set; }

    [Attr("achieved-date")]
    public DateTime? AchievedDate { get; set; }
}
```

1st PATCH Request example: Field is not sent 
PATCH http://localhost:5000/api/todoitem/1
```json
{
  "description": "New description"
}
```

2nd PATCH Request example: Field is sent 
PATCH http://localhost:5000/api/todoitem/1
```json
{
  "description": "New description",
  "achieved-date": "2017-04-25T21:41:29.033Z"
}
```

Actual behavior: Exception is returned from the API. 
Expected behavior:
- Field is ignored if not sent (1st example)
- Field is updated with new value if sent (2nd example)

_Gerfaut_

jaredcnance:
Closes #84 
Closes #93 
Closes #95


jaredcnance:
I'm going to let this sit for another day or two and then release

src/JsonApiDotNetCore/Internal/DasherizedRoutingConvention.cs
```diff
@@ -17,21 +19,36 @@ public DasherizedRoutingConvention(string nspace)
...
             foreach (var controller in application.Controllers)
-            {                
-                if (IsJsonApiController(controller))
+            {
+                var template = string.Empty;
+                
+                if (IsDasherizedJsonApiController(controller))
+                    template = $"{_namespace}/{controller.ControllerName.Dasherize()}";
+                else 
+                    template = GetTemplate(controller);
```
JanMattner: Why should anything be changed if the dasherized routing convention does not apply? Why not just leave it as is? E.g. imagine a controller with several stacked routes:

```csharp
[Route("/api/v1/MyController")]
[Route("/api/v2/MyController")]
public class MyController(...) { ... }
```

This is a valid use case: I want the same controller to be hosted for various routes. (not even restricted to API controllers, this also applies to static content or MVC controllers for documentation, login, ...)

Currently in this case, line 47 `var routeAttr = type.GetCustomAttribute<RouteAttribute>();` throws an exception that multiple attributes of the same type are found. Why not just leaving everything as is if the dasherized convention is disabled or the controller base type does not match? Of course then the user is responsible to actually define the routing and this is not done by the dasherized routing convention. Actually, if you think about it: even if you explicitly disable the dasherized routing convention, it still does apply a routing convention i.e. just the name of the controller. As a developer, I don't expect that component to do anything if I disable it.

jaredcnance: Hey thanks for bringing this to my attention and you're absolutely right. I hadn't considered multiple route attributes which is totally valid. 

The ideal scenario would be to not apply the convention at all for a particular controller. But, I'm not sure if this is possible (I'll be digging deeper this week). The line you commented on is more of a hack to try and determine what the intended route should be. In my limited experimentation, the simple fact that a [convention is added](https://github.com/Research-Institute/json-api-dotnet-core/blob/master/src/JsonApiDotNetCore/Extensions/IServiceCollectionExtensions.cs#L92) seems to alter the behavior. Even if I detect that the convention should not be applied and don't [alter the AttributeRouteModel](https://github.com/Research-Institute/json-api-dotnet-core/pull/94/files#diff-b6d62225c2ec19e65a68eb2b45a05092R30) it still doesn't seem to recognize attribute defined routes. 

I'll explore this further this week and hopefully come up with a better solution. I would expect this PR to be on hold till next weekend.

jaredcnance: disregard ^^ i had bad tests. 
opening PR to fix it now

PR #96, opened 2017-04-25
Allow PATCH objects with nullable attribute (e.g. DateTime?)

### LINKED ISSUE #95: PATCH request with nullable attributes does not work

Gerfaut:
Closes #95.

Ready to merge, tests are running. 

Problem was located in the AttrAttribute class, while setting the new value on model attributes. It was crashing on nullable types (e.g DateTime?). 

Fixed as http://stackoverflow.com/questions/3531318/convert-changetype-fails-on-nullable-types. 

Thank you!


src/JsonApiDotNetCore/Models/AttrAttribute.cs
```diff
@@ -26,10 +26,15 @@ public void SetValue(object entity, object newValue)
             var propertyInfo = entity
                 .GetType()
                 .GetProperty(InternalAttributeName);
-                
-            var convertedValue = Convert.ChangeType(newValue, propertyInfo.PropertyType);
```
jaredcnance: Thanks for the PR!

We do have a method called `TypeHelper.ConvertType` which is where type conversions should be processed, but looks like I missed this here. Can you try using that method instead and see if your tests still pass?

src/JsonApiDotNetCore/Models/AttrAttribute.cs
```diff
@@ -26,10 +26,15 @@ public void SetValue(object entity, object newValue)
...
-            propertyInfo.SetValue(entity, convertedValue);
+
+			if (propertyInfo != null)
+			{
+				Type t = Nullable.GetUnderlyingType(propertyInfo.PropertyType) ?? propertyInfo.PropertyType;
+
+				var convertedValue = (newValue == null) ? null : Convert.ChangeType(newValue, t);
+
+				propertyInfo.SetValue(entity, convertedValue, null);
+			}
```
jaredcnance: some funkiness with your tab spacing throughout...not a big deal just a note

Gerfaut: Thanks for review! I fixed it 😄

src/JsonApiDotNetCore/Models/AttrAttribute.cs
Gerfaut: Hi @jaredcnance, I tested with `TypeHelper.ConvertType` and all tests are green! 

While reading this method, I was wondering if it was not a good idea to replace 

```
if (type.GetTypeInfo().IsGenericType && type.GetGenericTypeDefinition().Equals(typeof(Nullable<>)))
                type = Nullable.GetUnderlyingType(type);
```

by 

`type = Nullable.GetUnderlyingType(type) ?? type;`

It changes nothing about the result (I think), but it is more easy to read (maybe subjective), WDYT? 

Thanks a lot for review!

jaredcnance: Yeah that looks fine to me and I agree the intent is clearer. Thanks again for the PR!

PR #97, opened 2017-04-27
Fix/#93

### LINKED ISSUE #93: Inconsistency for dasherized attributes

jaredcnance:
Closes #93

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ JsonApiDotnetCore provides a framework for building [json:api](http://jsonapi.or
 		- [Non-Integer Type Keys](#non-integer-type-keys)
 	- [Routing](#routing)
 		- [Namespacing and Versioning URLs](#namespacing-and-versioning-urls)
+        - [Disable Convention](#disable-convention)
 	- [Defining Custom Data Access Methods](#defining-custom-data-access-methods)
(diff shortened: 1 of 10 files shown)


jaredcnance:
@JanMattner this should handle your routing concerns. Let me know what you think. After this I think 2.0.0 is ready

PR #132, opened 2017-06-14
fix(deserialization): deserialize enums

jasolko:
fix exception thrown when attributes are of an enum type during
deserialization


jaredcnance:
looks good. can you bump the version in csproj to `2.0.9`? we'll get this merged into master as a hotfix and then I'll bring it back down to develop and unstable branches.

PR #187, opened 2017-11-10
Add SerializerSettings to JsonApiOptions

### LINKED ISSUE #185: Allow overriding more JsonSerializerSettings
Currently it is already possible to change the default `JsonContractResolver`. However it is not possible to change other settings (like adding the StringEnumConverter) for just JsonApiDotNetCore (it is possible to change the global default settings via `JsonConvert.DefaultSettings`)

Is there a reason why this is not possible? If not, I am happy to provide a PR where it is possible to change all settings.

dnperfors:
Closes #185 

#### FEATURE
- [X] fulfill the feature requirements
- [ ] bump package version

Open question: Should `JsonContractResolver` property be made `Obsolete` in favor of the new `SerializerSettings` property?


src/JsonApiDotNetCore/Configuration/JsonApiOptions.cs
```diff
@@ -3,47 +3,65 @@
...
-        {
-            if (builder == null) return;
-
-            builder(ContextGraphBuilder);
-
-            ContextGraph = ContextGraphBuilder.Build();
-        }
-    }
-}
+	public class JsonApiOptions
```
jaredcnance: Please fix formatting here. There is no need to change the size of the whitespace, nor introduce line spaces between properties. It also makes diffs more difficult.

src/JsonApiDotNetCore/Configuration/JsonApiOptions.cs
```diff
@@ -3,47 +3,65 @@
...
+
+		public IContractResolver JsonContractResolver
+		{
+			get => SerializerSettings.ContractResolver;
+			set => SerializerSettings.ContractResolver = value;
+		}
+
+		internal IContextGraphBuilder ContextGraphBuilder { get; } = new ContextGraphBuilder();
+
+		public JsonSerializerSettings SerializerSettings { get; } = new JsonSerializerSettings()
```
jaredcnance: :+1:
nit: let's move this to the line directly beneath the `JsonContractResolver` since they are related. Also, I agree that we should mark the `JsonContractResolver` as `Obsolete`

src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs
```diff
@@ -11,216 +11,211 @@
...
-            foreach (var attr in contextEntity.Relationships)
-            {
-                entity = attr.IsHasOne
-                    ? SetHasOneRelationship(entity, entityProperties, attr, contextEntity, relationships)
-                    : SetHasManyRelationship(entity, entityProperties, attr, contextEntity, relationships);
-            }
-
-            return entity;
-        }
+	public class JsonApiDeSerializer : IJsonApiDeSerializer
```
jaredcnance: Please fix whitespace formatting. I won't try to review this diff until it is absolutely clear what has been changed.

src/JsonApiDotNetCore/Configuration/JsonApiOptions.cs
dnperfors: I really thought I had my settings on spaces, but perhaps that is because I am used to using an editorconfig file... The extra lines where added by the CodeMaid plugin I use a lot, but I will disable that.

jaredcnance: That's fair. TBH I thought we had an .editorconfig :oops: thanks for mentioning it and I'll get that added.

dnperfors:
I made the changes, you want met to squash the commits together before it will be merged?

jaredcnance:
LGTM, please squash and then we're good to go. Thanks!

PR #204, opened 2017-12-08
200: fixed check of element type for IEnumerable<T>

### LINKED ISSUE #200: DocumentBuilder crashes when building Arrays
Method `DocumentBuilder.Build(IEnumerable<IIdentifiable> entities)`:

When `entities` is an Array (and thus not a generic type) .GenericTypeArguments[0] crashes with IndexOutOfRangeException.

Does anything speak against the following simple fix?

```csharp
var entitiesType = entities.GetType();
var entityType = entitiesType.IsArray ? entitiesType.GetElementType() : entitiesType.GenericTypeArguments[0];
```

shuebner:
Closes #200

#### BUG FIX
- [x] fix issue
- [ ] bump package version

### DIFF
--- a/src/JsonApiDotNetCore/Extensions/TypeExtensions.cs
+++ b/src/JsonApiDotNetCore/Extensions/TypeExtensions.cs
@@ -1,0 +1,35 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace JsonApiDotNetCore.Extensions
+{
+    internal static class TypeExtensions
+    {
+        public static Type GetElementType(this IEnumerable enumerable)
+        {
+            var enumerableTypes = enumerable.GetType()
+                .GetInterfaces()
+                .Where(t => t.IsGenericType == true && t.GetGenericTypeDefinition() == typeof(IEnumerable<>))
+                .ToList();
+
+            var numberOfEnumerableTypes = enumerableTypes.Count;
+
+            if (numberOfEnumerableTypes == 0)
+            {
+                throw new ArgumentException($"{nameof(enumerable)} of type {enumerable.GetType().FullName} does not implement a generic variant of {nameof(IEnumerable)}");
+            }
+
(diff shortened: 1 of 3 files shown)


src/JsonApiDotNetCore/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+            var enumerableTypes = enumerable.GetType()
+                .GetInterfaces()
+                .Where(t => t.IsGenericType == true && t.GetGenericTypeDefinition() == typeof(IEnumerable<>));
+
+            if (!enumerableTypes.Any())
+            {
+                throw new ArgumentException($"{nameof(enumerable)} of type {enumerable.GetType().FullName} does not implement a generic variant of {nameof(IEnumerable)}");
+            }
+
+            if (enumerableTypes.Count() > 1)
```
jaredcnance: Thanks for taking care of this! Only comment I have is that we could just call `SingleOrDefault` rather than calling `GetEnumerator` twice with `Count` and `Any`. We can catch the `InvalidOperationException`, do a `null` check  and throw the `ArgumentException`s in the appropriate cases.

Great work!

shuebner: I too felt it was a bit convoluted under the hood, but left it that way because of the readability and expressiveness of LINQ.
I am not a fan of catching exceptions that are easily avoidable and whose conditions are perfectly transparent.
So, I am going to suggest a middle ground without calling `GetEnumerator` multiple times, but also without catching provoked exceptions.

shuebner:
My force push seems to have removed your comment from this pull request chat.
Sorry, this is my first pull request and I thought you might not want to have multiple commit for such a simple issue.

jaredcnance:
No worries! Comments are still there:
image

It looks good to me. I'm going to merge this a little later today. I need to PR this back to v2.1 as well so I want to allow myself some time to manage the releases. Thanks again!

PR #220, opened 2018-01-03
Feat/#211

### LINKED ISSUE #211: Make service layer independent of persistence framework
I am currently writing an IEntityRepository implementation for MongoDB, for which there is no EF support.
The code for `Filter` and `Sort` is the same and `PageAsync` only differs in its query realization in that instead of `ToListAsync()` it is `ToAsyncEnumerable().ToList()`.

To avoid code duplication for those methods I suggest adding an abstract class that handles `Filter`, `Sort` and `PageAsync` for all DB drivers that can provide an `IQueryable<>`, with an abstract method for DB-specific query realization.
That abstract class would only depend on IJsonApiContext.

Alternatively the query realization could be in its own interface for DI.

What do you think?

shuebner:
Closes #211 

#### FEATURE

### DIFF
--- a/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
+++ b/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
@@ -157,5 +157,20 @@ namespace JsonApiDotNetCore.Data
                     .Take(numberOfElementsInPage)
                     .ToListAsync();
         }
+
+        public async Task<int> CountAsync(IQueryable<TEntity> entities)
+        {
+            return await entities.CountAsync();
+        }
+
+        public Task<TEntity> FirstOrDefaultAsync(IQueryable<TEntity> entities)
+        {
+            return entities.FirstOrDefaultAsync();
+        }
+
+        public async Task<IReadOnlyList<TEntity>> ToListAsync(IQueryable<TEntity> entities)
+        {
+            return await entities.ToListAsync();
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


shuebner:
Alright. I will try to fix those test failures.

shuebner:
I added some unit tests and also took the liberty of simplifying the paging acceptance tests by reducing the amount of logic in the assert sections.

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
```diff
@@ -156,26 +140,36 @@ public virtual IQueryable<TEntity> Include(IQueryable<TEntity> entities, string
...
-                        .ToListAsync();
-                else // page from the end of the set                   
-                    return (await entities
-                        .OrderByDescending(t => t.Id)
-                        .Skip((Math.Abs(pageNumber) - 1) * pageSize)
-                        .Take(pageSize)
-                        .ToListAsync())
-                        .OrderBy(t => t.Id)
-                        .ToList();
+                return await entities.PageForward(pageSize, pageNumber).ToListAsync();
```
jaredcnance: why are you calling the `ToListAsync()` extension method over your own `this.ToListAsync(/*...*/)` ?

shuebner: No particular reason. I thought, the Mongo extension methods are an implementation detail anyway. We can use the repo's own method if you like that better.

jaredcnance: That's fair, I suppose the query realization methods exposed here are really only required for the service layer's use.

PR #227, opened 2018-01-25
Feat/#226: Add DocumentBuilderOptions allowing omission of null attributes

### LINKED ISSUE #226: Ability to omit null valued attributes from responses?
Hey,

Nice project! Just a quick question - we have a lot of optional attributes in our data, is there a mechanism that we could use to omit the null valued attributes from responses?

lheiskan:
Closes #226

#### FEATURE
- [ ] bump package version

Looked into adding support for omitting null valued attributes from responses. This can be enabled globally by setting option 'OmitNullValuedAttributesFromResponses' to true. Default value for the option is false.

To consider before the merge: JsonApi is spec does not specify anything about omitting null values in responses - least as far as I can tell. It may be better to allow clients to opt-in for this functionality instead of having a global configuration, so only compliant clients can opt-in. What do you think?

### DIFF
--- a/src/JsonApiDotNetCore/Builders/DocumentBuilder.cs
+++ b/src/JsonApiDotNetCore/Builders/DocumentBuilder.cs
@@ -13,18 +13,14 @@ namespace JsonApiDotNetCore.Builders
         private readonly IJsonApiContext _jsonApiContext;
         private readonly IContextGraph _contextGraph;
         private readonly IRequestMeta _requestMeta;
+        private readonly DocumentBuilderOptions _documentBuilderOptions; 
 
-        public DocumentBuilder(IJsonApiContext jsonApiContext)
-        {
-            _jsonApiContext = jsonApiContext;
-            _contextGraph = jsonApiContext.ContextGraph;
-        }
-
-        public DocumentBuilder(IJsonApiContext jsonApiContext, IRequestMeta requestMeta)
+        public DocumentBuilder(IJsonApiContext jsonApiContext, IRequestMeta requestMeta=null, IDocumentBuilderOptionsProvider documentBuilderOptionsProvider=null)
         {
             _jsonApiContext = jsonApiContext;
             _contextGraph = jsonApiContext.ContextGraph;
             _requestMeta = requestMeta;
+            _documentBuilderOptions = documentBuilderOptionsProvider?.GetDocumentBuilderOptions() ?? new DocumentBuilderOptions(); ;
         }
 
         public Document Build(IIdentifiable entity)
@@ -118,8 +114,11 @@ namespace JsonApiDotNetCore.Builders
 
             contextEntity.Attributes.ForEach(attr =>
             {
-                if(ShouldIncludeAttribute(attr))
-                    data.Attributes.Add(attr.PublicAttributeName, attr.GetValue(entity));
+                var attributeValue = attr.GetValue(entity);
+                if (ShouldIncludeAttribute(attr, attributeValue))
+                {
+                    data.Attributes.Add(attr.PublicAttributeName, attributeValue);
+                }
             });
 
             if (contextEntity.Relationships.Count > 0)
@@ -128,11 +127,17 @@ namespace JsonApiDotNetCore.Builders
             return data;
         }
 
-        private bool ShouldIncludeAttribute(AttrAttribute attr)
+        private bool ShouldIncludeAttribute(AttrAttribute attr, object attributeValue)
         {
-            return (_jsonApiContext.QuerySet == null 
-                || _jsonApiContext.QuerySet.Fields.Count == 0 
-                || _jsonApiContext.QuerySet.Fields.Contains(attr.InternalAttributeName));
+            return !OmitNullValuedAttribute(attr, attributeValue)
+                   && ((_jsonApiContext.QuerySet == null
+                       || _jsonApiContext.QuerySet.Fields.Count == 0)
+                       || _jsonApiContext.QuerySet.Fields.Contains(attr.InternalAttributeName));
+        }
+
+        private bool OmitNullValuedAttribute(AttrAttribute attr, object attributeValue)
+        {
+            return attributeValue == null && _documentBuilderOptions.OmitNullValuedAttributes;
         }
 
         private void AddRelationships(DocumentData data, ContextEntity contextEntity, IIdentifiable entity)
--- a/src/JsonApiDotNetCore/Builders/DocumentBuilderOptions.cs
+++ b/src/JsonApiDotNetCore/Builders/DocumentBuilderOptions.cs
@@ -1,0 +1,16 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace JsonApiDotNetCore.Builders
+{
+    public struct DocumentBuilderOptions 
+    {
+        public DocumentBuilderOptions(bool omitNullValuedAttributes = false)
+        {
+            this.OmitNullValuedAttributes = omitNullValuedAttributes;
+        }
+
+        public bool OmitNullValuedAttributes { get; private set; }
+    }
+}
--- a/src/JsonApiDotNetCore/Configuration/NullAttributeResponseBehavior.cs
+++ b/src/JsonApiDotNetCore/Configuration/NullAttributeResponseBehavior.cs
@@ -1,0 +1,19 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace JsonApiDotNetCore.Configuration
+{
+    public struct NullAttributeResponseBehavior
+    {
+        public NullAttributeResponseBehavior(bool omitNullValuedAttributes = false, bool allowClientOverride = false)
+        {
+            OmitNullValuedAttributes = omitNullValuedAttributes;
+            AllowClientOverride = allowClientOverride;
+        }
+
+        public bool OmitNullValuedAttributes { get; }
+        public bool AllowClientOverride { get; }
+        // ...
+    }
+}
--- a/JsonApiDotnetCore.sln
+++ b/JsonApiDotnetCore.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26730.10
+VisualStudioVersion = 15.0.27004.2009
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "JsonApiDotNetCore", "src\JsonApiDotNetCore\JsonApiDotNetCore.csproj", "{C0EC9E70-EB2E-436F-9D94-FA16FA774123}"
 EndProject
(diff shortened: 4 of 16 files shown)


jaredcnance:
thanks for this!
> better to allow clients to opt-in for this functionality instead of having a global configuration

Not sure I understand what you mean here. The global configuration allows **server applications** to opt-in. How would you imagine a **client application** would opt-in?

src/JsonApiDotNetCore/Builders/DocumentBuilder.cs
```diff
@@ -128,6 +134,15 @@ private DocumentData GetData(ContextEntity contextEntity, IIdentifiable entity)
             return data;
         }
 
+        private bool ShouldIncludeNullValuedAttributes()
+        {
+            object isOmitNullValuedAttributes = false;
+            return 
+                (_requestMeta?.GetMeta()?.TryGetValue("jsonapidotnet:omit-null-valued-attributes", out isOmitNullValuedAttributes) ?? false ) ? 
+                (bool)isOmitNullValuedAttributes 
+                : _jsonApiContext.Options.OmitNullValuedAttributesFromResponses;
```
jaredcnance: This allows a client application to use this feature without explicit permission from the server. I would almost prefer a behavior enum in the options rather than a bool:
```
enum NullAttributeResponseBehavior {
  IncludeNullAttributes,
  ExcludeNullAttributes,
  OptionalForClients
}
```
or something along those lines. thoughts?

lheiskan:
Sounds reasonable, however, would a separate setting that enabled ability for the clients to override the global default be more flexible? It would better cater for schenarios like

global: include nulls, client:exclude
global: exclude nulls, client:include

The latter would be useful for api discovery, if by default server decided to omit nulls. Using the override at development time, clients could easily find out what is available.

jaredcnance:
That's fine as long as the server application has 100% control of possible behavior. So, the API you're proposing might look like:
```csharp
class NullAttributeResponseBehavior {
   public NullAttributeResponseBehavior(bool includeNullAttributes = true, bool allowClientOverride = false) {
    //...
  }
  // ...
}
class JsonApiOptions {
  public NullAttributeResponseBehavior  NullAttributeResponseBehavior { get; set; } 
    = new NullAttributeResponseBehavior()
}
```

I think I would still prefer to have it grouped into a class separate from the `JsonApiOptions` class.

lheiskan:
One question: i'd like to implement IRequestMeta service that would provide access to the meta data in the incoming jsonapi request, just wondering if there is a way to hook this in somehow? Something like this might make a reasonable default implementation as well? It could be then used as a way for the client to specify the null value handling. Does this make sense?

{
  "meta": {
    "jsonapidotnet:omit-null-valued-attributes": false,
  },
  "data": {
    // ...
  }
}

jaredcnance:
`IRequestMeta` is actually for generating **response** meta. See the [documentation](https://json-api-dotnet.github.io/JsonApiDotNetCore/meta.html) and the [usage](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/1226747e77c5551ec1b538e1a0b6d9b46a8e0771/src/JsonApiDotNetCore/Builders/DocumentBuilder.cs#L85-L86). The misnomer was unintentional and was only meant to distinguish itself from `IMeta` which was specific to resource models.

Currently, we don't do anything with the meta object during deserialization. It seems like adding something to the `JsonApiContext` might make sense since it's whole purpose is to carry ambient request information. The alternative would be to have an empty instance created in the scope whose properties are set in the deserializer. That might look like:

```class
public class RequestMetaDocument : IDictionary<string, object>
{
   public bool OmitNullValuedAttributes => this.TryGetValue("omit-null-valued-attributes", out var result) ? result : false;
  // ...
}
```

and then the `Document.Meta` type could be updated to `RequestMetaDocument`.

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/12eafc8f32d7350408a1ceccf4443ed2bcc00b44/src/JsonApiDotNetCore/Models/DocumentBase.cs#L15

lheiskan:
Ok, I kinda like the idea of placing the request meta into JsonApiContext, i could give it a shot.

public Dictionary<string, object> DocumentMeta { get; set; }

src/JsonApiDotNetCore/Builders/DocumentBuilder.cs
```diff
@@ -118,8 +127,14 @@ private DocumentData GetData(ContextEntity contextEntity, IIdentifiable entity)
 
             contextEntity.Attributes.ForEach(attr =>
             {
-                if(ShouldIncludeAttribute(attr))
-                    data.Attributes.Add(attr.PublicAttributeName, attr.GetValue(entity));
+                if (ShouldIncludeAttribute(attr))
+                {
+                    var value = attr.GetValue(entity);
+                    if (value != null || _documentBuilderOptions.IncludeNullAttributes)
```
jaredcnance: 2 things:
1. This probably belongs in the `ShouldIncludeAttribute` method, possibly as a call to another private method `AttrIsNotNullAndIsNotExcluded()` (or some better name).
2. This doesn't take into consider the `DocumentMeta` field, so I'm guessing this is still WIP?

src/JsonApiDotNetCore/Builders/DocumentBuilderOptions.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace JsonApiDotNetCore.Builders
+{
+    public struct DocumentBuilderOptions 
+    {
+        public DocumentBuilderOptions(bool isIncludeNullAttributes = false)
```
jaredcnance: nit: let's rename the parameter to `includeNullAttributes`

where were you planning on adding the setting to allow client overrides? maybe I'm missing something.

lheiskan:
I can solve my current use case with this implementation: I can inject null handling options of my choosing, and access to document meta that allows using client preferences as part of that process. Sweet!

Should I include an implementation of IDocumentBuilderBehaviour to be included with the library? Could developers using the library override the behavior with their own? I'm happy to include one if it seems useful. I guess we pretty much know the features that the implementation would have from the above discussion.

src/JsonApiDotNetCore/Builders/DocumentBuilder.cs
lheiskan: Cleaned this up a bit. My plan was to use the DocumentMeta in my own implementation if IDocumentBuilderBehavior - and that the interface could be similar to IRequestMeta in that library users can provide their own implementations. However, it might make sense to include a built in default implementation with the library - which I could do if you like.

src/JsonApiDotNetCore/Builders/DocumentBuilderOptions.cs
lheiskan: Hey - IDocumentBuilderBehaviour allows injecting an implementation that produces document builder options, i.e. an implementation could allow for determining null handling based no the global options and information provided in document meta. So if we want to provide a built in behaviour with the features discussed, i thought the client overrides would be handled in that implementation.

lheiskan:
Hey - added an implementation for a document builder behavior.

- If nothing is configured, nulls are not omitted and clients cannot affect this
- Options has a 'NullAttributeResponseBehavior'. This can be used to configure the global default and to allow clients to request a specific type of handling
- Clients can set omitNullValuedAttributes={true|false} query string parameter to request specific null handling. Server will ignore this if client overrides are not allowed.
- Query parameter only works if the 'AllowCustomQueryParameters' option is on (it is not on by default)

I actually ended up not using the document.meta, as this is not available for all methods - so perhaps there is no need to expose that, or the exposing might be more suited to another pull request.

If clients want to implement some other kind of mechanism for null behaviour, this can be done by service.Replace the default IDocumentBuilderBehaviour with a custom implementation - allowing implementations that use http headers, claims in access tokens or any other implementation specific mechanism.

The spec has a section that defines how to name implementation specific query parameters: http://jsonapi.org/format/#query-parameters. I think camelCase used above is compatible but other formats are also possible. Do you have a preference?

src/JsonApiDotNetCore/Builders/DocumentBuilderBehavior.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace JsonApiDotNetCore.Builders
+{
+    public class DocumentBuilderBehavior : IDocumentBuilderBehavior
+    {
+        private readonly IJsonApiContext _jsonApiContext;
+        private readonly IHttpContextAccessor _httpContextAccessor;
+
+        public DocumentBuilderBehavior(IJsonApiContext jsonApiContext, IHttpContextAccessor httpContextAccessor)
+        {
+            this._jsonApiContext = jsonApiContext;
```
jaredcnance: nit: `this.` is unnecessary and isn't used anywhere else in the library

src/JsonApiDotNetCore/Configuration/NullAttributeResponseBehavior.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public NullAttributeResponseBehavior(bool omitNullValuedAttributes = false, bool allowClientOverride = false)
+        {
+            OmitNullValuedAttributes = omitNullValuedAttributes;
+            AllowClientOverride = allowClientOverride;
+        }
+
+        public bool OmitNullValuedAttributes { get; }
+        public bool AllowClientOverride { get; }
+        // ...
```
jaredcnance: ☝️

src/JsonApiDotNetCore/Builders/DocumentBuilderBehavior.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using JsonApiDotNetCore.Services;
+using Microsoft.AspNetCore.Http;
+
+namespace JsonApiDotNetCore.Builders
+{
+    public class DocumentBuilderBehavior : IDocumentBuilderBehavior
```
jaredcnance: This class does not define the behavior of the `DocumentBuilder` but instead instantiates an options class based on config+request data. Th "behavior" suffix is confusing especially with the `NullAttributeResponseBehavior` struct whose purpose more closely matches the suffix. I would recommend renaming this class to `DocumentBuilderOptionsProvider`, `...Factory`, or  `...Service`

src/JsonApiDotNetCore/Builders/DocumentBuilder.cs
```diff
@@ -13,18 +13,14 @@ public class DocumentBuilder : IDocumentBuilder
...
-            _contextGraph = jsonApiContext.ContextGraph;
-        }
-
-        public DocumentBuilder(IJsonApiContext jsonApiContext, IRequestMeta requestMeta)
+        public DocumentBuilder(IJsonApiContext jsonApiContext, IRequestMeta requestMeta=null, IDocumentBuilderBehavior documentBuilderBehavior=null)
         {
             _jsonApiContext = jsonApiContext;
             _contextGraph = jsonApiContext.ContextGraph;
             _requestMeta = requestMeta;
+            _documentBuilderOptions = documentBuilderBehavior?.GetDocumentBuilderOptions() ?? _documentBuilderOptions;
```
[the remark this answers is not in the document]
lheiskan: You are right, it's much better that way.

src/JsonApiDotNetCore/Builders/DocumentBuilderBehavior.cs
lheiskan: Ok, makes sense. I'll go with DocumentBuilderOptionsProvider.

src/JsonApiDotNetCore/Builders/DocumentBuilderBehavior.cs
lheiskan: Cool, I'll remove that.

lheiskan:
Cleaned up the commits, I can have a look at the end to end test bit. We're currently using an internal branch, so from that perspective no need to back port things, but thanks for the offer :)

lheiskan:
Added few end to end test cases related to the null valued attribute handling.

lheiskan:
Nice - thank you for the library!

Issue #223, opened 2018-01-22
New Documentation


jaredcnance:
New documentation has been in the works for some time.

Some of the things I would like to do in conjunction with the new documentation:

- add setup document (adding services and middleware)
- add links in error logs to corresponding documentation (a great use for this is #170)
- document version migration steps and breaking changes
- API documentation, this would involve applying XML comments to all public APIs and then generating documentation 
- managing multiple DbContexts
- ...

Need to add documentation around:

- #227 
- #360 via #361

CoolGoose:
Hi @jaredcnance . Is there a plan on getting the new docummentation up and running ? Any way I can help consolidating the two websites (it is confusing at times)

bart-degreed:
Closing this, as the documentation has been updated. Consolidating multiple versions of online docs is tracked at #768.

PR #232, opened 2018-02-11
Fix broken url for IResourceService

Biarity:


jaredcnance:
Thanks! By the way I am in the process of creating new and updated documentation (#223) so most of this will move soon.

PR #243, opened 2018-03-23
Initializing PageOffset to be 1 (instead of default 0) so that the 'n…

### LINKED ISSUE #242: 'page[number]' value for 'next' pagination link incorrectly set to 1 when only 'page[size]' is passed
**Reproducible:** Always

**Steps to Reproduce:**

1. In the sample ToDoItem app create 5+ Todo Items
2. Access '/api/v1/todo-items?page[size]=2'
3. Notice the 'next' link's page[number] value is 1 when it should be 2

nwise:
…ext' pagination link is set properly when no 'page[number]' is passed in the query string.

Closes #242 

#### BUG FIX
- [x] fix issue
- [x] bump package version

#### FEATURE
- [ ] write tests that address the requirements outlined in the issue
- [ ] fulfill the feature requirements
- [ ] bump package version

### DIFF
--- a/src/JsonApiDotNetCore/Internal/Query/PageQuery.cs
+++ b/src/JsonApiDotNetCore/Internal/Query/PageQuery.cs
@@ -3,6 +3,6 @@ namespace JsonApiDotNetCore.Internal.Query
     public class PageQuery
     {
        public int PageSize { get; set; }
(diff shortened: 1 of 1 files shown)


jaredcnance:
Thanks! Will go out in the  deployment this weekend.

PR #264, opened 2018-04-20
Fix the return variable for the Resource Services doc

barryofguilder:
In the code sample on the Resource Services documentation, it was returning the wrong variable.

### DIFF
--- a/docs/ResourceServices.md
+++ b/docs/ResourceServices.md
@@ -37,7 +37,7 @@ public class TodoItemService : EntityResourceService<TodoItem> {
         _notificationService.Notify($"Entity created: { newEntity.Id }");
 
         // don't forget to return the new entity
-        return entity;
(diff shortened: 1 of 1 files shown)


jaredcnance:
Thanks! But, unfortunately the docs have been moved to: https://github.com/json-api-dotnet/json-api-dotnet.github.io.

The only reason I haven't removed this from the repo was in case anyone was still using the old links. But, it's probably time to shut it down.

But, thank you for pointing this out. It looks as though the problem exists in [the new docs](https://github.com/json-api-dotnet/json-api-dotnet.github.io/blob/dbbe123e5854da2b6c9c19e686749c55c9ab68cc/src/pages/extensibility/ResourceServices.js#L55) as well. So, I'll get that fixed. I have a few other changes I need to make as well.

PR #287, opened 2018-05-25
Add Negative filter

milosloub:
Only extension of current filter system by adding NotEqual case for filter[attribute]=ne:value

### DIFF
--- a/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
+++ b/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
@@ -195,6 +195,10 @@ namespace JsonApiDotNetCore.Extensions
                 case FilterOperations.like:
                     body = Expression.Call(left, "Contains", null, right);
                     break;
+                    // {model.Id != 1}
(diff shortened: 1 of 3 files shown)


jaredcnance:
Awesome! Thanks! Would you mind adding tests? You should be able to just add it onto these: https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/AttributeFilterTests.cs

milosloub:
Sure, I'll have some problems with testing. Locally tests passes, but after push test fail. I'll investigate this later

milosloub:
Test is fixed now. Problem was my clear postgres db in docker container. While running this test, I didn't count with data from previous tests. 

Btw I have another idea of extending filter for array use-cases. For example I'm
looking to find those people whose age's are present in the arrayValues. 
So "filter[age]=in:[15,18,30]" would be translated as:
People.Where(s=>arrayValues.Contains(s.age))
What do you think about this solution?

jaredcnance:
@milosloub that sounds good.

PR #299, opened 2018-06-12
Make `AddJsonApi` chainable

btecu:
Currently, it is not possible to chain `AddJsonApi`:
`services.AddJsonApi<DbContext>().AddMvcCore()`

It should be, to be inline with all other apis (including all Microsoft).

It's not a breaking change. I've modified the example application and a test to use it.

I also thought about adding it to `AddJsonApiInternals` but that should be private - seems like it will be in the future.


jaredcnance:
Looks good. Thanks!

Issue #290, opened 2018-06-05
Improve Contributing Guide


jaredcnance:
**References**
- [Servicestack](https://github.com/ServiceStack/ServiceStack/wiki/HowTo-write-unit-integration-tests)
- [Code of Conduct](https://gist.github.com/PurpleBooth/b24679402957c63ec426)
- [OpenStack](https://wiki.openstack.org/wiki/How_To_Contribute)
- [EmberJS](https://github.com/emberjs/ember.js/blob/master/CONTRIBUTING.md)
- [ReactJS](https://reactjs.org/docs/how-to-contribute.html)

**Examples of Things We Might Inclue**
- How to update documentation
- Searching through issues
- Proposing new features...The worst thing in the world is opening a PR that gets rejected.
- Cloning the repo
- Forking the repo
- Setting up PostgreSQL (currently in README)
- Running tests (currently in README)
- Changing example app data models (e.g. don't use migrations)
- Running the example apps
- The build/release system
- PR approval requirement
- [API Deprecration Strategy](https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/404#issuecomment-420483020):
  1. Introduce new APIs in minor release, deprecate old APIs with error: false
  2. Deprecate old APIs in major release with error: true
  3. Remove old APIs in major release
- New APIs must contain XML documentation (can fail the build for this?)
- Debugging test failures via appveyor

![image](https://user-images.githubusercontent.com/13851721/41691695-75259c12-74c1-11e8-8cb3-4e0e59fd76c8.png)

Issue #281, opened 2018-05-08
Run tests against multiple versions of ASP.Net Core


jaredcnance:
We should be able to run tests against multiple versions of ASP.Net Core by defining the default versions in Directory.Build.props and then override by passing the versions as  msbuild properties:

```
dotnet test ./test/UnitTests/UnitTests.csproj \
  /p:AspNetCoreVersion=2.0.0 \
  /p:EFCoreVersion=2.0.0 \
  /p:MicrosoftLoggingVersion=2.0.0
```

PR #300, opened 2018-06-12
Use `Mvc.Core` instead of `Mvc`

btecu:
I wanted to add the package to my project and it asked to install a lot of dependencies. First step to reduce that is to use `Mvc.Core` instead of `Mvc`. `Mvc` has a lot of extra unneeded packages, such as Razor (template support).
install

I also updated all other packages.

I noticed a lot of warnings related to `xunit`. Probably existing, but I fixed those as well.


btecu:
There appears to be one failing test. Until #290 is completed, I won't be able to figure it out which one is failing, since output is extremely verbose (and unhelpful). Otherwise, maybe someone can point out which test is failing?

Directory.Build.props
```diff
@@ -4,28 +4,27 @@
     <NetCoreAppVersion>netcoreapp2.0</NetCoreAppVersion>
     <NetStandardVersion>netstandard2.0</NetStandardVersion>
 
-    <AspNetCoreVersion>2.0.1</AspNetCoreVersion>
+    <AspNetCoreVersion>2.1.0</AspNetCoreVersion>
```
jaredcnance: I'm a bit nervous about pushing this in a non-major release until #281 is done

src/Examples/JsonApiDotNetCoreExample/Controllers/Restricted/ReadOnlyController.cs
```diff
@@ -5,7 +5,7 @@ namespace JsonApiDotNetCoreExample.Controllers.Restricted
 {
     [Route("[controller]")]
     [HttpReadOnly]
-    public class ReadOnlyController : Controller
+    public class ReadOnlyController : ControllerBase
```
jaredcnance: Thank you! I've been wanting to do this for a while. I don't _think_ anyone is generating views via their JADNC controllers, but if they are this would be a breaking change 🤔 ... I think this is the right direction. Just wondering about versioning.

test/JsonApiDotNetCoreExampleTests/Acceptance/Extensibility/CustomErrorTests.cs
```diff
@@ -11,7 +11,7 @@ public class CustomErrorTests
         public void Can_Return_Custom_Error_Types()
         {
             // arrange
-            var error = new CustomError("507", "title", "detail", "custom");
+            var error = new CustomError(507, "title", "detail", "custom");
```
jaredcnance: 👏

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/AttributeFilterTests.cs
```diff
@@ -131,7 +131,7 @@ public AttributeFilterTests(TestFixture<TestStartup> fixture)
 
             // assert
             Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-            Assert.False(deserializedTodoItems.Any(i => i.Ordinal == todoItem.Ordinal));
+            Assert.DoesNotContain(deserializedTodoItems, x => x.Ordinal == todoItem.Ordinal);
```
jaredcnance: :clap:

jaredcnance:
The test is failing because of this: https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/3fc1099e15257e55573178b3dd19edd0ae6f7f16/test/JsonApiDotNetCoreExampleTests/Helpers/Extensions/IQueryableExtensions.cs#L19

I'm using this to double check actual SQL output of EF queries (`ToSql`):
https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/3fc1099e15257e55573178b3dd19edd0ae6f7f16/test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/SparseFieldSetTests.cs#L56

But, it doesn't look like you upgraded EF, so I'm not sure why this is failing...

Directory.Build.props
```diff
@@ -4,28 +4,27 @@
...
-    <MicrosoftLoggingVersion>2.0.0</MicrosoftLoggingVersion>
-    <MicrosoftConfigurationVersion>2.0.0</MicrosoftConfigurationVersion>
-    <MicrosoftOptionsVersion>2.0.0</MicrosoftOptionsVersion>
+    <MicrosoftLoggingVersion>2.1.0</MicrosoftLoggingVersion>
+    <MicrosoftConfigurationVersion>2.1.0</MicrosoftConfigurationVersion>
+    <MicrosoftOptionsVersion>2.1.0</MicrosoftOptionsVersion>
 
-    <EFCoreVersion>2.0.1</EFCoreVersion>
-    <EFCoreToolsVersion>2.0.1</EFCoreToolsVersion>
+    <EFCoreVersion>2.1.0</EFCoreVersion>
```
jaredcnance: The test is failing because of this. We do some reflection on EF APIs to check the actual SQL output of queries (`ToSql`). Looks like these properties no longer exist or were renamed.

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/3fc1099e15257e55573178b3dd19edd0ae6f7f16/test/JsonApiDotNetCoreExampleTests/Helpers/Extensions/IQueryableExtensions.cs#L19

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/3fc1099e15257e55573178b3dd19edd0ae6f7f16/test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/SparseFieldSetTests.cs#L56

Directory.Build.props
btecu: We could get #281 and bump the version or just release it. I suspect there will be no significant issues since 2.0 to 2.1 is more of a performance upgrade so just releasing might be fine.

jaredcnance: fyi - I've started this work over in #302 , should have it done tonight or tomorrow....the fix for the failing tests here will be the same as in that PR.

Directory.Build.props
jaredcnance: #302 has a fix for this and includes the tests for the prior frameworks. I'll probably merge that in tonight and then you can rebase onto those changes.

jaredcnance: I decided to hold on #302 for now. I'm going to merge this in and get a release started.

btecu:
Thanks @jaredcnance!

PR #302, opened 2018-06-12
[WIP] Run Tests Against Multiple Versions of ASP.Net and EF Core

### LINKED ISSUE #281: Run tests against multiple versions of ASP.Net Core

jaredcnance:
Closes #281 

The idea is that for every combination of versions that we choose to explicitly guarantee support for, we will define a new .props file that specifies those versions. The example projects should then import those props files and we run tests against those versions. 

The primary package (JsonApiDotNetCore) will continue to import via Directory.Build.props which will define the default build dependencies. This allows us to simulate scenarios such as:

```
MyApp
├── JsonApiDotNetCore
│        └── AspNetCore v2.0.1
└── AspNetCore v2.1.0

MyApp
├── JsonApiDotNetCore
│        └── AspNetCore v2.1.0
└── AspNetCore v2.0.1
```

Build is currently failing for the same reason as #300 ... breaking EF Core changes

### TODO:

- [ ] Fix failing tests
- [ ] Document how this works and how we upgrade our default dependencies
- [ ] Can we drop `/p:NoWarn=NU1605` ?


jaredcnance:
Putting this on hold...can't justify the added complexity at this time.

Issue #301, opened 2018-06-12
Reduce build verbosity


jaredcnance:
- https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/3fc1099e15257e55573178b3dd19edd0ae6f7f16/Build.ps1#L26-L38

- https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/3fc1099e15257e55573178b3dd19edd0ae6f7f16/build.sh#L8-L11

PR #304, opened 2018-06-13
Further reduce tests verbosity

btecu:
Completes #301


btecu:
@jaredcnance this is ready for review.
Reduces log lines (Travis) from 30,870 to 3,201.

There are a lot of warnings and errors that can be cleaned up too.

PR #308, opened 2018-06-15
Fixing issue where total-records was 0 on POST and PATCH calls.

### LINKED ISSUE #307: total-records on POST and PATCH methods
On POST and PATCH methods, I noticed that total-records is returned as 0. I think this should either be 1 or that meta attribute should not be included.

nwise:
Closes #307

#### BUG FIX
- [x] fix issue
- [ ] bump package version

### DIFF
--- a/src/JsonApiDotNetCore/Services/JsonApiContext.cs
+++ b/src/JsonApiDotNetCore/Services/JsonApiContext.cs
@@ -86,20 +86,20 @@ namespace JsonApiDotNetCore.Services
             const char pathSegmentDelimiter = '/';
 
             var span = requestPath.AsSpan();
-            
+
             // we need to iterate over the string, from the end,
             // checking whether or not the 2nd to last path segment
             // is "relationships"
             // -2 is chosen in case the path ends with '/'
-            for(var i = requestPath.Length - 2; i >= 0; i--)
+            for (var i = requestPath.Length - 2; i >= 0; i--)
             {
                 // if there are not enough characters left in the path to 
                 // contain "relationships"
-                if(i < relationships.Length) 
+                if (i < relationships.Length)
                     return false;
 
                 // we have found the first instance of '/'
-                if(span[i] == pathSegmentDelimiter)
+                if (span[i] == pathSegmentDelimiter)
                 {
                     // in the case of a "relationships" route, the next
                     // path segment will be "relationships"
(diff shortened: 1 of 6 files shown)


src/JsonApiDotNetCore/Services/JsonApiContext.cs
```diff
@@ -119,11 +119,13 @@ private PageManager GetPageManager()
                 return new PageManager();
 
             var query = QuerySet?.PageQuery ?? new PageQuery();
+            var requestMethod = _httpContextAccessor.HttpContext.Request.Method;
 
             return new PageManager
             {
                 DefaultPageSize = Options.DefaultPageSize,
                 CurrentPage = query.PageOffset,
+                TotalRecords = (requestMethod == "POST" || requestMethod == "PATCH") ? 1 : 0,
```
jaredcnance: I feel like this may change the semantics of TotalRecords a bit. Since in GET requests, TotalRecords indicates the number of records in the database and not the number of records in the response, this would change the meaning of the field for POST and PATCH requests.

nwise:
Yeah, that makes sense. I'd prefer that, if the total-records value is included, it should be accurate. What do you think about simply not adding it on PATCH/POST responses? I think that would be simplest and also not require an extra call to the DB.

jaredcnance:
Yeah I 100% agree. I think we should exclude it in those cases.

milosloub:
I think that this is good idea. Property TotalRecords corresponds to amount of returned data in case of filtering. So in case of POST and PATCH is good idea to set TotalRecords  : 1. It trully returns only one record.

milosloub:
I'm sorry. I didn't notice 

> Yeah I 100% agree. I think we should exclude it in those cases.

I think this is better solution to exlude in all irrelevant cases. My suggestion was for "total-record" always presented

jaredcnance:
I _think_ we can simply make `TotalRecords` `int?` and change the following line from:
https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/8f7f0f7e2112993d656258b779afd784ac5b1b99/src/JsonApiDotNetCore/Builders/DocumentBuilder.cs#L78-L79

to:
 
```c#
if (_jsonApiContext.Options.IncludeTotalRecordCount && _jsonApiContext.PageManager.TotalRecords != null) 
  builder.Add( /* ... */ )
```

What this implies is that it will only be included if we have explicitly set it earlier in the pipeline.

jaredcnance:
Hey guys, take a look at the changes and let me know what you think. I've modified it so even if the payload returns empty, we will still include the `total-records` meta attribute. Example:

```http
GET /api/articles HTTP/1.1
```
```http
HTTP/1.1 200 OK

{ "data": [], "meta": { "total-records":0 } }
```

nwise:
I think this looks good. I like that it's always included on empty results too. It will make coding clients against it simpler. Thanks!

PR #316, opened 2018-06-28
Feature/#252 valid modelstate

### LINKED ISSUE #252: Check ModelState.IsValid by Default
Gitter transcript:

> Kevin Sigmund @ksigmund Mar 26 08:49
> If I have a custom ValidationAttribute that returns false, will the error naturally be returned in the result from the controller, or do I need to override PostAsync, check the ModelState, etc.?

> Kevin Sigmund @ksigmund Mar 26 09:22
> Basically I'm just looking to do simple model validation, but it doesn't seem that when my model is invalid that the failure is respected.

> Jared Nance @jaredcnance Mar 26 10:05
> We do not currently check ModelState.IsValid ... I don't want to make any assumptions about how model validation is implemented. However, I'm open to discussion and with a little more information I think it may be reasonable to do that in the library automatically (while providing some form of escape hatch). My main concern would be around not making any breaking changes without good reason.

> Kevin Sigmund @ksigmund Mar 26 10:09
> Right now it seems that the framework is fairly optimistic about the state of the model right? i.e. without any additional logic, failures are ignored essentially.

> Jared Nance @jaredcnance Mar 26 10:12
As long as it passes de-serialization, we perform no further model validations. That said, we also shouldn't be preventing any model validations in custom controllers. Is model validation something you believe should be handled by the framework by default?

> Kevin Sigmund @ksigmund Mar 26 10:16
> I think so - and I believe that's the way that the .NET Core framework is moving as well.
It seems like if I have a [Required] or [SomeOtherValidation] attribute if the validation fails, something is wrong and I probably don't want to ignore that. Maybe it's some sort of config at setup where we could instruct the framework to respect validations.

> Jared Nance @jaredcnance Mar 26 10:18
> I agree.

sigent:
Closes #252

#### FEATURE

Addressing Breaking Change - As there are no implicit Validation providers in MVC core, you need to explicitly add validation to the pipeline e.g. services.AddMvcCore().AddDataAnnotations()
Without this ModelState should always be valid.

### DIFF
--- a/src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs
+++ b/src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using JsonApiDotNetCore.Extensions;
 using JsonApiDotNetCore.Internal;
 using JsonApiDotNetCore.Models;
 using JsonApiDotNetCore.Services;
@@ -152,6 +153,7 @@ namespace JsonApiDotNetCore.Controllers
 
             if (!_jsonApiContext.Options.AllowClientGeneratedIds && !string.IsNullOrEmpty(entity.StringId))
                 return Forbidden();
+            if (!ModelState.IsValid) return BadRequest(ModelState.ConvertToErrorCollection());
 
             entity = await _create.CreateAsync(entity);
 
@@ -164,6 +166,7 @@ namespace JsonApiDotNetCore.Controllers
 
             if (entity == null)
                 return UnprocessableEntity();
+            if (!ModelState.IsValid) return BadRequest(ModelState.ConvertToErrorCollection());
 
             var updatedEntity = await _update.UpdateAsync(id, entity);
 
--- a/src/JsonApiDotNetCore/Extensions/ModelStateExtensions.cs
+++ b/src/JsonApiDotNetCore/Extensions/ModelStateExtensions.cs
@@ -1,0 +1,23 @@
+using JsonApiDotNetCore.Internal;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.EntityFrameworkCore.Internal;
+
+namespace JsonApiDotNetCore.Extensions
+{
+    public static class ModelStateExtensions
+    {
+        public static ErrorCollection ConvertToErrorCollection(this ModelStateDictionary modelState)
+        {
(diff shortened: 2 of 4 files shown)


src/JsonApiDotNetCore/JsonApiDotNetCore.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <VersionPrefix>2.3.2</VersionPrefix>
+    <VersionPrefix>2.3.3</VersionPrefix>
```
jaredcnance: Wondering if this should be a major version bump. It _could_ break existing apps if they have annotations but aren't currently checking them. We could release this as a patch version behind a feature flag (e.g. `options.ValidateModelState`) with a default disabled value. Then in the next major release enable it by default. What do you think?

src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs
```diff
@@ -152,6 +153,7 @@ public class BaseJsonApiController<T, TId>
 
             if (!_jsonApiContext.Options.AllowClientGeneratedIds && !string.IsNullOrEmpty(entity.StringId))
                 return Forbidden();
+            if (!ModelState.IsValid) return BadRequest(ModelState.ConvertToErrorCollection());
```
jaredcnance: nit: `return` on newline

src/JsonApiDotNetCore/Extensions/ModelStateExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace JsonApiDotNetCore.Extensions
+{
+    public static class ModelStateExtensions
+    {
+        public static ErrorCollection ConvertToErrorCollection(this ModelStateDictionary modelState)
+        {
+            ErrorCollection errors = new ErrorCollection();
+            foreach (var entry in modelState)
+            {
+                if (!entry.Value.Errors.Any()) continue;
```
jaredcnance: nit: `continue` on newline

src/JsonApiDotNetCore/JsonApiDotNetCore.csproj
sigent: You could argue that if they have annotations already on their entities, then its already invalid and it's now being flagged. However, the feature flag route makes sense for a minor release :-D 

Made the changes

PR #330, opened 2018-07-05
v2.3.4 Introduce ResourceDefinition

### LINKED ISSUE #297: Prevent attributes from being serialized
From @rtablada:

> For some models it would be helpful to be able to define which attrs are/aren't used for serializing vs deserializing. [...] Ideally we should be able to declare/override which attrs are used for `fetch`/`create`/`update` operations similar to JSON API resources http://jsonapi-resources.com/v0.9/guide/resources.html#Fetchable-Attributes (we probably can't reproduce this exact API since JADNC uses model decorators instead of a separate resource class)

## Proposal
```c#
public UserAttributeFilter : IAttributeFilter<User> 
{
    // exclude the User.Password attribute from serialization
    public HashSet<AttrAttribute> ShouldInclude(HashSet<AttrAttribute> attributes)
      => return attributes
            .Where(attr => attr.InternalAttributeName != nameof(User.Password))
            .ToHashSet();
}
```

### LINKED ISSUE #321: Enable SourceLink to improve debugging experience
https://github.com/dotnet/sourcelink

jaredcnance:
Closes #297
Closes #321

### DIFF
--- a/src/JsonApiDotNetCore/Models/AttrAttribute.cs
+++ b/src/JsonApiDotNetCore/Models/AttrAttribute.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Reflection;
 using JsonApiDotNetCore.Internal;
 
 namespace JsonApiDotNetCore.Models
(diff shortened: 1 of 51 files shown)


src/JsonApiDotNetCore/Models/AttrAttribute.cs
```diff
@@ -69,34 +70,64 @@ public AttrAttribute(string publicName, string internalName, bool isImmutable =
...
+            }            
+        }
 
-            if (propertyInfo != null)
+        private PropertyInfo GetResourceProperty(object resource)
+        {
+            // There are some scenarios, especially ones where users are using a different
+            // data model than view model, where they may use a repository implmentation
+            // that does not match the deserialized type. For now, we will continue to support
+            // this use case.
```
jaredcnance: Original release notes...saving for later:

> We are now caching `PropertyInfo` when we build the `ContextGraph` so that we can reduce the amount of reflection performed per-request. Previously when setting attribute values during serialization, we would look up properties by name on the incoming type. This allowed sharing repository implementations for multiple types that may have common attribute names. However, this is incorrect behavior and it has been indirectly corrected through this performance improvement.

Issue #317, opened 2018-06-28
RFC: Type rename proposals


jaredcnance:
### Feedback on this issue is desired

There are several types that could have better names. We can change the names and deprecate the old ones via inheritance.

| Current Type | Description | Proposal(s) | Final |
|---|---|---|---|
| ContextGraph | A graph of all resources and their metadata | ResourceGraph<br>JsonApiResourceGraph<br>JsonApiSchema | |
| ContextGraphBuilder | Constructs the resource graph | ResourceGraphBuilder<br>JsonApiSchemaBuilder  | |
| ContextEntity | A json:api resource and its metadata | Resource<br>JsonApiResource<br>~ResourceDefinition~ | |
| DocumentData | The json:api document resource object as defined in the spec | ResourceObject | ResourceObject |
| JsonApiException<br>.GetError() | Returns the exception as a json:api formatted error object | JsonApiException.GetErrorCollection() | |
| EntityResourceService | ... | DefaultResourceService | |
| IEntity...Repository | Used by the resource service to construct and materialize queries | IQueryBuilder | |
| [`IEntityReadRepository.Get()`](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/491a6cbed1b68a6fcc9ce69ca8b00d3b20b586c7/src/JsonApiDotNetCore/Data/IEntityReadRepository.cs#L21) | Used to get the base `IQueryable` for which most queries are built from. See https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/404#issuecomment-420468203 | . GetQueryable(...) | |
| `GenericProcessor` | When updating relationships, it is helpful to be able to use generic methods even though the generic parameter is different from the controller's generic parameters. The `GenericProcessor` gets created and then used through a non-generic interface. However, all of this makes what it does a bit confusing. Really, all it is used for is updating relationships. The fact that we chose to do it this particular way should not influence the name. | `UpdateRelationshipProcessor`<br />`PatchRelationshipProcessor`  | |

Any other suggestions?

maurei:
All but the proposal for `IQueryBuilder` have been processed in #589. We think that the repository layer is more broad than the responsibility as implied by the name `IQueryBuilder`

PR #344, opened 2018-07-18
Entity and Resource separation

### LINKED ISSUE #112: RFC: Resource to Entity Mapping
**Start Date**: N/A
**Status**: WIP
**RFC PR**: N/A

## Summary

Allow models to map to more than 1 resource.

## Terminology

- **Entity** refers to the database model. For EF users this is the POCO class that corresponds to a `DbSet<>` on the `DbContext`
- **Resource** refers to the model exposed by the API

## Motivation

Currently, there is a 1:1 relationship between the database entity and the exposed resource. There may be a need for different forms of an entity to be exposed. I see two different needs:

- `1` entity : `N` resources (one-to-many)
- `N` entities : `1` resource (many-to-one)

I believe the second need is already supported since this would require a custom implementation at the repository layer. The first however should be supported out-of-the-box and it currently is not.

## Detailed Design

I think this is a relatively simple solution and that is to change the current dependency graph from:

```
|- Controller<TEntity>
|--- IResourceService<TEntity>
|------ IRepository<TEntity>
```

to

```
|- Controller<TResource>
|--- IResourceService<TResource, TEntity>
|------ IRepository<TEntity>
```

This shows that mapping would take place in the service layer. An application could define a mapping that, if exists, would be used to map the final entity result (post-database call) into the resource. For this, I think it is safe to choose [AutoMapper](https://github.com/AutoMapper/AutoMapper) as the standard mapping tool. Users would then define their maps using the `ContextGraphBuilder`:

```csharp
options.BuildContextGraph((builder) => {
  builder.UseAutoMapper(AutoMapperConfig.GetMapper());
});
```

and the controller might look like:

```csharp
public class UsersController : JsonApiController<User>
{
    public UsersController(
        IJsonApiContext jsonApiContext, 
        IResourceService<User, Person> resourceService, 
        ILoggerFactory loggerFactory) 
        : base(jsonApiContext, resourceService, loggerFactory)
    { }
}
```

### ContextGraphBuilder

The [ContextGraphBuilder](https://github.com/Research-Institute/json-api-dotnet-core/blob/master/src/JsonApiDotNetCore/Builders/ContextGraphBuilder.cs) will have a dictionary of mappings:

```csharp
Dictionary<Type, List<IResourceMap>> _resourceMaps;
```

Calling `builder.UseAutoMapper(AutoMapperConfig.GetMapper()` will create the resource maps for every entity. All ContextEntities will be assigned an `EmptyMapping` which will handle calls for mapping `TResource` to `TEntity` when `TResource == TEntity`:

```csharp
_resourceMaps[typeof(UnMappedEntity)] = new List<IResourceMap> { new EmptyMapping() };
```

As the ContextEntities are created, they will be assigned their maps:

```csharp
_entities.Add(new ContextEntity
{
    Mappings = _resourceMaps[entityType],
    // ...
});
```

### Resource Services

In order to maintain backwards compatibility, the current resource service definitions with generic parameters should remain available. We can define an additional generic parameter overload:

```csharp
public interface IResourceService<TEntity> : IResourceService<TEntity, int>
{}

public interface IResourceService<TEntity, TId> : IResourceService<TEntity, TEntity, TId>
{}

public interface IResourceService<TResource, TEntity, TId>
{
 // ...
}
```

A `IResourceService` implementation should then perform the mapping prior to returning the resource:

```csharp
public EntityResourceService(
    IJsonApiContext jsonApiContext,
    // ...
    )
{
    _jsonApiContext = jsonApiContext;
    _mapper = jsonApiContext.ContextGraph.GetMapper<TResource, TEntity>();
    // ...
}

public async Task<T> GetAsync(TId id)
{
    // ...
    var entity = //...
    var resource = _mapper.Map(entity); 
    return entity;
}
```

### Automapper IResourceMap

```csharp
public class AutoMapperService<TResource, TEntity>  
  : IResourceMap<TResource, TEntity>
{
  private readonly IMapper _mapper;
  public AutoMapperService(IMapper mapper) {
    _mapper = mapper;
  }

  public TResource Map(TEntity entity) 
     => _mapper.Map<TEntity, TResource>(entity);

  public IEnumerable<TResource> Map(IEnumerable<TEntity> entities) 
     => _mapper.Map<IEnumerable<TEntity>, IEnumerable<TResource>>(entities);

  public TEntity Map(TResource resource) 
     => _mapper.Map<TResource, TEntity>(resource);

  public string GetEntityRelationshipName(string resourceRelationshipName) {
     // ...
  }
}
```

### Unspecified Mappings

If a user defines a controller but no mapping has been defined, a 500 error will be returned.

### Relationships

Mappings will be used to translate relationship names. In the following example, a request for `include=owner` would map to `Entity.User`: 

```csharp
Mapper.Initialize(cfg => 
      cfg.CreateMap<Entity, Resource>()
    	.ForMember(dest => dest.Owner, opt => opt.MapFrom(src => src.User)));

// ...
public class Entity : Identifiable {
  public virtual Person User { get; set; }
}

public class Resource : Identifiable {
  [HasOne("owner")]
  public Person Owner { get; set; }
}
```

Validation of these definitions _should_ occur on app startup.

### Constraints

The types `TResource` and `TEntity` **MUST** implement `IIdentifiable<TId>` where `TId` is of the same type for both classes:

```csharp
public interface IResourceService<TResource, TEntity, TId> 
  where TResource : class, IIdentifiable<TId>
```

roblankey:
Closes #112

- [X] fulfill the feature requirements
- [X] bump package version

A couple of specific notes RE: the implementation.
 - I didn't think it was necessary to add mapping support to the contextgraph, instead opting for a much simpler DI of automapper profiles.
 - To simplify relationship mapping and limit the amount of reflection necessary, there is an assumed convention of naming the mapped properties the same. Mapping between them is accomplished by use of automapper profiles 

Finally, an additional note. Initially I did not want to include resource representations of the many-to-many entities that entity framework requires, but the additional complexity for mapping these out to relationships the context graph would know about does not seem to provide real value, since you can accomplish the same via a resource declaration.


src/JsonApiDotNetCore/Services/EntityResourceService.cs
```diff
@@ -133,7 +116,7 @@ private bool ShouldIncludeRelationships()
             var relationship = _jsonApiContext.ContextGraph
                 .GetContextEntity(typeof(T))
                 .Relationships
-                .FirstOrDefault(r => r.InternalRelationshipName == relationshipName);
+                .FirstOrDefault(r => r.PublicRelationshipName == relationshipName);
```
jaredcnance: Awesome catch! Should actually use: `r.Is(relationshipName)`

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/404231bc2a96ee5cfbacf8b89c67b6af1d19a6eb/src/JsonApiDotNetCore/Models/RelationshipAttribute.cs#L74-L75

src/JsonApiDotNetCore/Services/MappingResourceService.cs
```diff
@@ -0,0 +1,204 @@
...
+using AutoMapper;
+using JsonApiDotNetCore.Data;
+using JsonApiDotNetCore.Extensions;
+using JsonApiDotNetCore.Internal;
+using JsonApiDotNetCore.Models;
+using Microsoft.Extensions.Logging;
+
+namespace JsonApiDotNetCore.Services
+{
+    public class MappingResourceService<TResource, TEntity>
```
jaredcnance: I would prefer to not duplicate the actual implementation details. My thought was that we would have a single, generalized implementation that would look like:

```c#
public ResourceService<TResource> 
  : ResourceService<TResource, int>, IResourceService<TResource> 
{ /* ... */ }

public ResourceService<TResource, TId> 
  : ResourceService<TResource, TResource, TId>, IResourceService<TResource, TId> 
{ /* ... */ }

public ResourceService<TResource, TEntity, TId> 
  : IResourceService<TResource, TEntity, TId>  
{
    public ResourceService(
                IJsonApiContext jsonApiContext,
                IEntityRepository<TEntity, TId> entityRepository,
                ILoggerFactory loggerFactory) 
    {
         // no mapper provided, TResource & TEntity must be the same type
         if(typeof(TResource) != typeof(TEntity)) throw new InvalidOperationException("...");
    }

    public ResourceService(
                IJsonApiContext jsonApiContext,
                IEntityRepository<TEntity, TId> entityRepository,
                ILoggerFactory loggerFactory,
                IMapper mapper) { /* ... */ }
}
```

Now obviously this would allow for runtime errors, so #326 could include checking to verify at startup.

src/JsonApiDotNetCore/Services/MappingResourceService.cs
```diff
@@ -0,0 +1,204 @@
...
+        : MappingResourceService<TResource, TEntity, int>,
+        IResourceService<TResource>
+        where TResource : class, IIdentifiable<int>
+        where TEntity : class, IIdentifiable<int>
+    {
+        public MappingResourceService(
+          IJsonApiContext jsonApiContext,
+          IEntityRepository<TEntity> entityRepository,
+          ILoggerFactory loggerFactory,
+          IMapper mapper)
```
jaredcnance: As already stated, we don't want a hard dependency on AutoMapper. As outlined in #112, it should take an internally defined interface such as a `IResourceMap<TResource, TEntity>` and not make any assumptions about a single general map instance. This also enables mapper agnostic startup validation as stated in the previous comment. 

Example: 

```c#
services.AddScoped<IResourceService<FooResource, FooEntity, int>>();
services.AddSingleton<IResourceMap<FooResource, FooEntity>>(
  new AutoMapperAdapter<FooResource, FooEntity>()
);

// OR this could be bundled into a JsonApiDotNetCore.AutoMapper extension method
// on IServiceCollection like:

services.AddAutoMappedResource<FooResource, FooEntity>();
```

By doing this we can do a few things at startup:
1. For every resource, check to see if an `IResourceService<TResource, TEntity, int>` implementation exists and has been registered on the container
2. If the implementation exists, check to see if an `IResourceMap<FooResource, FooEntity>` exists and has been registered on the container
3. Also, the `AutoMapperAdapter` (or whatever it is called) can verify the mapping is actually defined.

If these checks fail, we can provide quick feedback that a configuration error has occurred:
 - A mapper has not been registered `(TResource != TEntity && MapperIsRegistered() == false)`
 - The AutoMapper.Mapper does not define a required mapping
 - maybe others ?

test/ResourceEntitySeparationExampleTests/.gitignore
```diff
@@ -0,0 +1,234 @@
+## Ignore Visual Studio temporary files, build results, and
```
jaredcnance: shouldn't be necessary

src/Examples/JsonApiDotNetCoreExample/Models/Resources/CourseStudentResource.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace JsonApiDotNetCoreExample.Models.Resources
+{
+    /// <summary>
+    /// Note: EF Core *requires* the creation of an additional entity
+    /// for many to many relationships and no longer implicitly creates
+    /// it. While it may not make sense to create a corresponding "resource"
+    /// for that relationship, due to the need to make the underlying 
+    /// framework and mapping understand the explicit navigation entity, 
+    /// a mirroring DTO resource is also required.
+    /// </summary>
```
jaredcnance: 🤔 I _think_ we can find a way to get around this... 
Maybe not in v1, but it would be a pretty cool feature if we could allow applications to hide this implementation detail.

src/Examples/JsonApiDotNetCoreExample/Migrations/20180717134637_ResourceSeparation.cs
```diff
@@ -0,0 +1,117 @@
+using Microsoft.EntityFrameworkCore.Migrations;
+using Npgsql.EntityFrameworkCore.PostgreSQL.Metadata;
+
+namespace JsonApiDotNetCoreExample.Migrations
+{
+    public partial class ResourceSeparation : Migration
```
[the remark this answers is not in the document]
roblankey: thanks!

src/Examples/ResourceEntitySeparationExample/Profiles/StudentProfile.cs
```diff
@@ -0,0 +1,39 @@
...
+            foreach (CourseStudentEntity reg in registrations)
+            {
+                CourseEntity e = reg.Course;
+                courses.Add(new CourseResource
+                {
+                    Id = e.Id,
+                    Number = e.Number,
+                    Title = e.Title,
+                    Description = e.Description
+                });
```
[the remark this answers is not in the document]
roblankey: This bit here is for mapping the underlying many to many 3 entities to a more logical 2 resource setup. This really just provides the direct link from Student to Courses without having to go through the CourseStudent artificial resource.

src/JsonApiDotNetCore/Builders/ContextGraphBuilder.cs
```diff
@@ -125,7 +125,7 @@ protected virtual List<RelationshipAttribute> GetRelationships(Type entityType)
             {
                 var attribute = (RelationshipAttribute)prop.GetCustomAttribute(typeof(RelationshipAttribute));
                 if (attribute == null) continue;
-                attribute.InternalRelationshipName = prop.Name;
+                if (attribute.InternalRelationshipName == null) attribute.InternalRelationshipName = prop.Name;
```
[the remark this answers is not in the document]
roblankey: Thanks for catching! I had originally done this so that I could override the name in the attribute but have since moved away from that time the convention of naming things the same

src/JsonApiDotNetCore/JsonApiDotNetCore.csproj
```diff
@@ -17,9 +17,11 @@
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <Version>2.4.1</Version>
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="4.0.1" />
```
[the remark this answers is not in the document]
roblankey: Good call. Thanks

src/JsonApiDotNetCore/Services/MappingResourceService.cs
roblankey: Ah. Okay, yeah.i misunderstood that bit. Makes sense.

src/Examples/JsonApiDotNetCoreExample/Models/Resources/CourseStudentResource.cs
roblankey: Definitely. This bit was really only needed for patching the many to many relationship due to the framework not immediately knowing anything about the entity and the genericprocessor not knowing what to do with it.

I really wanted to avoid creating this but thought the rest might be valuable enough to postpone this bit.

roblankey:
Thanks for the feedback and catches. Next step is generalizing out the mapping resource.

roblankey:
the last question i have here is around the 112 conversation with configuration checking. for the moment, i've left IResourceMapper untyped really only because that's how IMapper works and i wanted to be able to inject an AutoMapper configuration pretty directly on to it like in

`
public class AutoMapperAdapter : IResourceMapper
    {
        private readonly IMapper _mapper;

        public AutoMapperAdapter(IMapper mapper)
        {
            _mapper = mapper;
        }

        public TDestination Map<TDestination>(object source)
        {
            return _mapper.Map<TDestination>(source);
        }

        public TDestination Map<TSource, TDestination>(TSource source)
        {
            return _mapper.Map<TSource, TDestination>(source);
        }
    }
`

and Startup

`
public ConfigureServices(...)
{
    ....
    services.AddAutoMapper();
    services.AddScoped<IResourceMapper, AutoMapperAdapter>();
    ....
}
`

What are your thoughts on that? Ideas on how to bridge that gap from AutoMapper's overall approach to type specific IResourceMapper implementations?

jaredcnance:
🤔 so, from a discovery/validation perspective `IResourceMapper` would just need to expose a `ContainsMapping(Type in, Type out)` method so that we can check if the mapping exists (this doesn't have to be done right now). This is arguably a better experience than having to lookup on the container to see if a particular type (e.g. `IMapper<TIn, TOut>`) has been registered. 

Also, if you need type specific overrides, you can just create a different implementation or interface and depend on that instead. So, let's say I use AutoMapper for 99% of my mapping needs, but there is this one resource/entity pair where I need something else. I would achieve that by doing something like:

```c#
// CustomMapper.cs
public class FooMapper : IResourceMapper { /* ... */ }

// Startup.cs
services.AddSingleton(new CustomMapper());

// FooService.cs
public class FooService : DefaultResourceService {
  public FooService(
    IJsonApiContext jsonApiContext,
    IEntityRepository<Bar> entityRepository,
    ILoggerFactory loggerFactory,
    FooMapper mapper // <----- need to inject concrete type (or custom abstraction)
  ) : base(jsonApiContext, entityRepository,  loggerFactory, mapper) { /* ... */ }
}
```

So, I think that _should_ be fine. I'll need to think about other scenarios as well.

src/JsonApiDotNetCore/Services/DefaultResourceService.cs
```diff
@@ -0,0 +1,246 @@
...
+using JsonApiDotNetCore.Models;
+using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace JsonApiDotNetCore.Services
+{
+    public class DefaultResourceService<TResource> : DefaultResourceService<TResource, int>,
```
jaredcnance: I'd prefer not to rename this class just yet. This would result in a wide-spread breaking change. I'd prefer to hold off until #317. I can add this as a suggestion to the open issue.

roblankey: Good deal. Undoing the rename.

jaredcnance:
I've retargeted to the develop branch so I can publish a release to our MyGet feed (https://www.myget.org/F/research-institute/api/v3/index.json). I'll be integrating this into a larger project tomorrow and if everything goes smoothly, I'll merge into master early this week. Thanks again for all your hard work! 🍻

roblankey:
Awesome! Thanks for all the feedback along the way!

PR #371, opened 2018-08-07
Validations Send source.pointer

rtablada:


src/JsonApiDotNetCore/Extensions/ModelStateExtensions.cs
```diff
@@ -6,7 +6,7 @@ namespace JsonApiDotNetCore.Extensions
 {
     public static class ModelStateExtensions
     {
-        public static ErrorCollection ConvertToErrorCollection(this ModelStateDictionary modelState)
+        public static ErrorCollection ConvertToErrorCollection<T>(this ModelStateDictionary modelState, IContextGraph contextGraph)
```
jaredcnance: So, that we don't have to release this in a major version, let's leave the original implementation and deprecate it using the `[Obsolete("Use the generic overload instead")]`. This prevents us from breaking anyone that is calling this API directly.

Also, I've recently been leaning towards the idea of depending on the static `ContextGraph.Instance` rather than injecting everywhere. I haven't found many reasons the `ContextGraph` should change after app start. Thoughts?

src/JsonApiDotNetCore/Extensions/ModelStateExtensions.cs
```diff
@@ -16,10 +16,19 @@ public static ErrorCollection ConvertToErrorCollection(this ModelStateDictionary
 
                 foreach (var modelError in entry.Value.Errors)
                 {
+                    var attrName =contextGraph.GetPublicAttributeName<T>(entry.Key);
```
jaredcnance: let's do this lookup outside this loop since `entry.Key` doesn't change...results in unnecessary enumerations of the graph

rtablada: 👍

src/JsonApiDotNetCore/Extensions/ModelStateExtensions.cs
rtablada: 🤔 for `ContextGraph.Instance` I wonder if that's safe given possible different versions of the API running at once with different context graphs?

If `ContextGraph.Instance` is used other places that could still be an issue there, but IDK if we'd want to tie this down to?

PR #373, opened 2018-08-10
fix(#343): reload relationships from database if included during POST

### LINKED ISSUE #343: [2.4.0-beta1]: POST with self referencing inclusion returns null values (or default) in included entity
I have entity Client like this:

```
    {
        [Attr("firstName")]
        public string FirstName  { get; set; }

        [Attr("lastName")]
        public string LastName  { get; set; }

        //[HasOne("advisor", Link.None)]
        //public Advisor Advisor { get; set; }
        //public int AdvisorId { get; set; }

        [HasOne("recommender", Link.None)]
        public Client Recommender { get; set; }
        public int? RecommenderId { get; set; }
    }

```
Trying to post like:
POST: http://localhost:5001/api/clients?include=recommender

```
{  
   "data":{  
      "type":"clients",
      "attributes":{  
         "firstName":"Mike",
         "lastName":"Razor"
      },
      "relationships":{  
         "recommender":{  
            "data":{  
               "type":"clients",
               "id":"1"
            }
         }
      }
   }
}
```

It returns:

```
{  
   "data":{  
      "attributes":{  
         "firstName":"Mike",
         "lastName":"Razor"
      },
      "relationships":{  
         "recommender":{  
            "data":{  
               "type":"clients",
               "id":"1"
            }
         }
      },
      "type":"clients",
      "id":"2"
   },
   "included":[  
      {  
         "attributes":{  
            "firstName":null,  // <--- This should be not null
            "lastName":null   // <--- This should be not null
         },
         "relationships":{  
            "recommender":{}
         },
         "type":"clients",
         "id":"1"
      }
   ]
}
```
This is caused only when there is self referencing inclusion (client include another client).
When I change POST with entity of another type (commented Advisor for example), POST with inclusion returns all data normally.

In version 2.3.1 it works fine. 2.3.2 and above has this problem

jaredcnance:
Closes #343

#### BUG FIX
- [x] fix issue
- [x] bump package version

Verified solution results in only two queries regardless of number of included relationships:

image

### DIFF
--- a/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
+++ b/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
@@ -13,6 +13,7 @@ using Microsoft.Extensions.Logging;
 
 namespace JsonApiDotNetCore.Data
 {
+    /// <inheritdoc />
     public class DefaultEntityRepository<TEntity>
         : DefaultEntityRepository<TEntity, int>,
         IEntityRepository<TEntity>
@@ -26,6 +27,10 @@ namespace JsonApiDotNetCore.Data
         { }
     }
 
+    /// <summary>
+    /// Provides a default repository implementation and is responsible for
+    /// abstracting any EF Core APIs away from the service layer.
+    /// </summary>
     public class DefaultEntityRepository<TEntity, TId>
         : IEntityRepository<TEntity, TId>
         where TEntity : class, IIdentifiable<TId>
@@ -48,7 +53,7 @@ namespace JsonApiDotNetCore.Data
             _genericProcessorFactory = _jsonApiContext.GenericProcessorFactory;
         }
 
-        /// </ inheritdoc>
+        /// <inheritdoc />
         public virtual IQueryable<TEntity> Get()
         {
             if (_jsonApiContext.QuerySet?.Fields != null && _jsonApiContext.QuerySet.Fields.Count > 0)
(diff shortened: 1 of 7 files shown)


src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
```diff
@@ -185,10 +193,57 @@ public virtual IQueryable<TEntity> Include(IQueryable<TEntity> entities, string
...
+            if (relationshipAttr.IsHasOne && _jsonApiContext.HasOneRelationshipPointers.Get().TryGetValue(relationshipAttr, out var pointer))
+            {
+                await _context.Entry(pointer).ReloadAsync();
+            }
+
+            if (relationshipAttr.IsHasMany && _jsonApiContext.HasManyRelationshipPointers.Get().TryGetValue(relationshipAttr, out var pointers))
+            {
+                foreach (var hasManyPointer in pointers)
+                {
+                    await _context.Entry(hasManyPointer).ReloadAsync();
```
jaredcnance: this is bad ☹️  ... need to reload the entire navigation property

milosloub: First of all. Thank you, for this solution.
For HasMany relationships, what about something like:
_context.Entry(entity).Collection(relation).Load();

instead of:
foreach (var pointer in hasManyRelationship.Value)
{
       _context.Entry(pointer).State = EntityState.Detached;
}

jaredcnance: Thanks for the suggestion, I wasn't aware of the `.Collection` api. I _think_ I like the approach I ended up taking better. The reason is that we should end up with a single SQL request that uses `JOIN` to load all the relationships rather than reloading each included relationship individually. I'll review it again later tonight to verify the difference.

PR #378, opened 2018-08-14
[WIP] Feat/#39: Deeply Nested Inclusions

### LINKED ISSUE #39: Deeply Nested Relationship Inclusion
```http
GET /articles/1?include=comments.author HTTP/1.1
```
> Note: Because compound documents require full linkage (except when relationship linkage is excluded by sparse fieldsets), intermediate resources in a multi-part path must be returned along with the leaf nodes. For example, a response to a request for comments.author should include comments as well as the author of each of those comments.

> Note: A server may choose to expose a deeply nested relationship such as comments.author as a direct relationship with an alias such as comment-authors. This would allow a client to request /articles/1?include=comment-authors instead of /articles/1?include=comments.author. By abstracting the nested relationship with an alias, the server can still provide full linkage in compound documents without including potentially unwanted intermediate resources.

Will also need to include support for Sparse Fieldsets of included relationships

jaredcnance:
Closes #39

#### FEATURE
- add unit tests for
  - [x] Serialization
  - [x] De-Serialization
- add acceptance tests for
  - [x] deeply nested inclusions spanning a HasMany relationship (`?include=many.one`)
  - [x] deeply nested inclusions targeting a HasMany relationship (`?include=one.many`)

#### KNOWN ISSUES

- sparse field-sets on included entities not supported yet
- sort and filter on nested relationship attrs not yet supported

### DIFF
--- a/src/JsonApiDotNetCore/Builders/DocumentBuilder.cs
+++ b/src/JsonApiDotNetCore/Builders/DocumentBuilder.cs
@@ -192,20 +192,35 @@ namespace JsonApiDotNetCore.Builders
             return relationshipData;
         }
 
-        private List<DocumentData> GetIncludedEntities(List<DocumentData> included, ContextEntity contextEntity, IIdentifiable entity)
+        private List<DocumentData> GetIncludedEntities(List<DocumentData> included, ContextEntity rootContextEntity, IIdentifiable rootResource)
         {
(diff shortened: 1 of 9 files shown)


jaredcnance:
We still need to do some work on this, but I want to go ahead and get it out in v3.0.0-beta4 to drive community feedback.

PR #386, opened 2018-08-28
WIP: Support Filtering using ne:null on nullable attribute

NullVoxPopuli:
#### BUG FIX
- [x] fix issue
- [ ] bump package version

The root of this was that I wanted to be able to `?filter[nullable-field]=ne:`, meaning that I wanted all resources where `nullable-field` had a value. 

I had originally tried using `?filter[nullable-field]=ne:null` but, this was interpreting `null` as `"null"`, which resulted in the error: `Could not cast null to Nullable'1`. 

Since checking if something is not null can be handy, I added a check for the `Nullable` generic type, and if the value is an empty string, swap it out for null. :-\

This _may_ be a breaking change. idk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,3 +75,41 @@ public class Startup
     }
 }
 ```
+
+### Development
+
+Restore all nuget packages with:
+
+```bash
+dotnet restore
+```
+
+#### Testing
+
+Running tests locally requires access to a postgresql database.  
+If you have docker installed, this can be propped up via: 
+
+```bash
+docker run --rm --name jsonapi-dotnet-core-testing \
+  -e POSTGRES_DB=JsonApiDotNetCoreExample \
+  -e POSTGRES_USER=postgres \
+  -e POSTGRES_PASSWORD=postgres \
+  -p 5432:5432 \
+  postgres
+```
+
+And then to run the tests:
+
+```bash
+dotnet test
+```
+
+#### Cleaning
+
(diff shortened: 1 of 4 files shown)


NullVoxPopuli:
what if I add two FilterOperations for `notnull` and `@null`?

NullVoxPopuli:
Maybe something like this?
image
(just now value on the right side of the colon)

milosloub:
Yeah, thats the problem of string stored NULL values.
What about something like filter[attr]=is:null and filter[attr]=isnot:null?
This is similar to SQL queries

NullVoxPopuli:
what's to stop the usage of `is:somethingelse`? :-\ would we hard-code checks on `null` value usage?

milosloub:
Here is null value like every other value after colon. So there must be conversion to null in this case. Maybe someone comes with better idea.
In case of is:somethingElse...its should be the same as eq:somethingElse. I didnt try the difference between Items.Where(i => i.Attr == value) and Items.Where(i => i.Attr.Equals(value)) from EF Core point of view.
I dont know if EF core translate Items.Where(i => i.Attr.Equals(null)) to SQL:  "WHERE Attr IS NULL" by default.
If yes, operator "is" or "isnot" can be used to build Expression based od Equals Method.

NullVoxPopuli:
cool. I'll submit a different PR with `is` and `isnot`

Issue #294, opened 2018-06-10
RFC: Fluent Testing API


jaredcnance:
# Update: I am considering moving this as an add-on to [DotnetTestAbstractions](https://github.com/jaredcnance/DotnetTestAbstractions).

This is part of the initiative defined by #251.

**Status:** WIP (feedback desired)

# Summary 

Provide a fluent testing API that abstracts away the boilerplate required by most json:api requests.

# Motivation

Currently, there is a lot of boilerplate required to test a single json:api endpoint. For example, this is one of the tests used against the example app in this repo:

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/a1832f764a9861c06ed919d75b8a9a80fa1f525e/test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/CreatingDataTests.cs#L43-L83

We should be able to drastically reduce this to something a bit more readable:

```c#
// arrange
var expectedOwner = new Person();
var expectedTodoCollection = new TodoCollection();
var request = new CreateRequest<TodoCollection>(expectedTodoCollection)
   // need to add `expectedOwner` to db prior to sending the request
  .WithNewRelationship(todoCollection => todoCollection.Owner, expectedOwner);

// act
var resultantTodoCollection = await request.SendAndValidateAsync();
/* your assertions go here */
```

# Detailed Design

Create a new package `JsonApiDotNetCore.Testing`

## Possible APIs

| API | Description |
|---|---|
| `JsonApiRequest<T>.SendAndValidateAsync()` | Sends the request and performs json:api related validations |
| `JsonApiRequest<T>.SendAsync()` | Sends the request without performing validations |
| `JsonApiRequest<T>.Dump()` | Writes the HTTP request to the console |
| `CreateRequest<T>` |  RequestBuilder for POST requests |
| `GetRequest<T>` |  RequestBuilder for GET requests |
| `UpdateRequest<T>` |  RequestBuilder for PATCH requests |
| `DeleteRequest<T>` |  RequestBuilder for DELETE requests |
| `BulkRequest` |  RequestBuilder for operations request |
| `IWriteRequest<T>.WithNewRelationship` | This would create the entity prior to sending the request and then include the rio in the `relationships.{}.data` element |
| `IRequest<T>.WithHeader` | Add custom headers |
| `IGetRequest<T>.Filter` | Add filter |
| `IGetRequest<T>.Sort` | Add filter |
[...]

PR #387, opened 2018-08-28
Additional filter operations: isnull and isnotnull

NullVoxPopuli:
This superscedes: https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/386/files
Closes #{ISSUE_NUMBER}

#### FEATURE
- [ ] bump package version

/cc @milosloub

### DIFF
--- a/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
+++ b/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
@@ -127,11 +127,12 @@ namespace JsonApiDotNetCore.Extensions
 
                     return source.Where(lambdaIn);
                 }
-                else if (op == FilterOperations.@is || op == FilterOperations.isnot) {
+                else if (op == FilterOperations.isnotnull || op == FilterOperations.isnull) {
+                    // {model}
                     var parameter = Expression.Parameter(concreteType, "model");
                     // {model.Id}
                     var left = Expression.PropertyOrField(parameter, property.Name);
-                    var right = Expression.Constant(filterQuery.PropertyValue, typeof(string));
+                    var right = Expression.Constant(null);
 
                     var body = GetFilterExpressionLambda(left, right, op);
                     var lambda = Expression.Lambda<Func<TSource, bool>>(body, parameter);
@@ -139,22 +140,9 @@ namespace JsonApiDotNetCore.Extensions
                     return source.Where(lambda);
                 }
                 else
-                {   
-                    var isNullabe = IsNullable(property.PropertyType);
-                    var propertyValue = filterQuery.PropertyValue;
-                    var value = propertyValue;
-
-                    if (op == FilterOperations.@isnot || op == FilterOperations.isnot)
-                    {
-                        if (isNullabe && propertyValue == "null") 
-                        {
-                            value = null;
-                        }
-                    }
-
-                    // convert the incoming value to the target value type
+                {   // convert the incoming value to the target value type
                     // "1" -> 1
-                    var convertedValue = TypeHelper.ConvertType(value, property.PropertyType);
+                    var convertedValue = TypeHelper.ConvertType(filterQuery.PropertyValue, property.PropertyType);
                     // {model}
                     var parameter = Expression.Parameter(concreteType, "model");
                     // {model.Id}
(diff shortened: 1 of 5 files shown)


test/JsonApiDotNetCoreExampleTests/Acceptance/TodoItemsControllerTests.cs
```diff
@@ -90,6 +91,52 @@ public async Task Can_Filter_TodoItems()
...
+            var route = $"/api/v1/todo-items?filter[updated-date]=isnot:null";
+            var request = new HttpRequestMessage(httpMethod, route);
+
+            // Act
+            var response = await _fixture.Client.SendAsync(request);
+            var body = await response.Content.ReadAsStringAsync();
+            var deserializedBody = _fixture.GetService<IJsonApiDeSerializer>().DeserializeList<TodoItem>(body);
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
```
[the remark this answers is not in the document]
NullVoxPopuli: I getchya. I was just copying from other tests ;)

src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
```diff
@@ -113,31 +113,56 @@ public static IQueryable<TSource> Filter<TSource>(this IQueryable<TSource> sourc
...
 
                     return source.Where(lambdaIn);
                 }
+                else if (op == FilterOperations.@is || op == FilterOperations.isnot) {
+                    var parameter = Expression.Parameter(concreteType, "model");
+                    // {model.Id}
+                    var left = Expression.PropertyOrField(parameter, property.Name);
+                    var right = Expression.Constant(filterQuery.PropertyValue, typeof(string));
+
+                    var body = GetFilterExpressionLambda(left, right, op);
```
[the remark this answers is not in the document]
NullVoxPopuli: > Maybe something like this. And any value after the operator is ignored.

This is actually what I started with, should be easy to go back. :)

> [Attr(unsetFilterValueBehavior: UnsetFilterValueBehavior.AsNull)]

I guess the case I keep thinking of is what if someone has a nullable string field?


In previously projects (rails), I used https://github.com/activerecord-hackery/ransack which has a bunch of matchers: https://github.com/activerecord-hackery/ransack#search-matchers including null and not_null

test/JsonApiDotNetCoreExampleTests/Acceptance/TodoItemsControllerTests.cs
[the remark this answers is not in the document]
NullVoxPopuli: @jaredcnance how much refactoring would you want on acceptance tests? I typically extract all the repetitive stuff like request building and such into helper methods (I could do this in a subsequent PR)

test/JsonApiDotNetCoreExampleTests/Acceptance/TodoItemsControllerTests.cs
```diff
@@ -90,6 +91,56 @@ public async Task Can_Filter_TodoItems()
...
+            // Act
+            var response = await _fixture.Client.SendAsync(request);
+
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            var body = await response.Content.ReadAsStringAsync();
+            var deserializedBody = _fixture.GetService<IJsonApiDeSerializer>().DeserializeList<TodoItem>(body);
+
+            // Assert
+            Assert.NotEmpty(deserializedBody);
```
jaredcnance: Let's verify the returned `TodoItem`s have actually been filtered (e.g. all results have `UpdatedDate = null`. 
Maybe even consider explicitly adding a `TodoItem` to the db with `UpdatedDate = notNullDateValue;`
It shouldn't be an issue after #294, but for now I think we should assume we're not working with an empty db.

test/JsonApiDotNetCoreExampleTests/Acceptance/TodoItemsControllerTests.cs
```diff
@@ -90,6 +91,56 @@ public async Task Can_Filter_TodoItems()
...
```
jaredcnance: See comment below.

NullVoxPopuli: added assertion

test/JsonApiDotNetCoreExampleTests/Acceptance/TodoItemsControllerTests.cs
NullVoxPopuli: added assertion

PR #388, opened 2018-08-28
Exact Error Codes for Error Collections

rtablada:
#### FEATURE
 
This change makes it so that error collections will send a status based on the status code of all items in the collection.

If the items in the collection have the same status code, then it return the status code of the entire collection.

If not, it will return 500 or 400 as before.

### DIFF
--- a/test/UnitTests/Internal/JsonApiException_Test.cs
+++ b/test/UnitTests/Internal/JsonApiException_Test.cs
@@ -1,0 +1,31 @@
+using JsonApiDotNetCore.Internal;
+using Xunit;
+
+namespace UnitTests.Internal
+{
+    public class JsonApiException_Test
+    {
+        [Fact]
+        public void Can_GetStatusCode()
+        {
+            var errors = new ErrorCollection();
+            var exception = new JsonApiException(errors);
+
+            // Add First 422 error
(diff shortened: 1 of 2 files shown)


test/UnitTests/Internal/JsonApiException_Test.cs
```diff
@@ -0,0 +1,31 @@
...
+            errors.Add(new Error(422, "Something wrong"));
+            Assert.Equal(422, exception.GetStatusCode());
+
+            // Add a second 422 error
+            errors.Add(new Error(422, "Something else wrong"));
+            Assert.Equal(422, exception.GetStatusCode());
+
+            // Add 4xx error not 422
+            errors.Add(new Error(401, "Unauthorized"));
+            Assert.Equal(400, exception.GetStatusCode());
```
NullVoxPopuli: yeeeesssss

rtablada:
@jarednance what would be the timeline for this to be released?

rtablada:
Is master the 3.0 beta or stil 2.5.1+?

If we could back port this and cut a 2.5.2 that would be ❤️

jaredcnance:
Change has been backported to 2.5.2. Let me know if you run into any issues with that release.

PR #409, opened 2018-09-23
Feat/#293: Naming Conventions, Improvements to AutoDiscovery, and ResourceDefinition

### LINKED ISSUE #293: Attribute Naming Conventions
It would be great if we could define an app-wide naming convention for attributes and relationships with default options. So, instead of:
```c#
[Attr("fooBar")]
public object FooBar { get; set; }
```

You would specify a convention:

```c# 
// Startup.cs
options.AttributeNameConvention = NamingConvention.CamelCase;

// Model.cs

[Attr]
public object FooBar { get; set; }
```

A possible interface for `AttributeNameConvention` might be:

```c#
public interface INamingConvention {
  string Convert(string propertyName);
}
```

### LINKED ISSUE #384: Provide API for easy service/repo registration
## Description

```c#
services.AddResourceService<FooService>();
```

should register any applicable forms:

```c#
services.Add<IResourceService<Foo>, FooService>();
services.Add<IResourceService<Foo, int>, FooService>();
services.Add<IGetAllService<Foo>, FooService>();
services.Add<IGetAllService<Foo, int>, FooService>();
// ... 
```

It should also handle services with mappings:

```c#
services.Add<IResourceService<Foo, Bar, int>, FooService>();
```

This API may be usable by the service discovery pipeline

### LINKED ISSUE #402: RFC: Enhancements to ResourceDefintion
## Description

The `ResourceDefinition` class exposes developer friendly hooks into how their resources are exposed. It is intended to improve the experience and reduce boilerplate for commonly required features.

As with the prior implementation, the `ResourceDefinition` will be resolved from the container and can accept any dependencies that have been registered.

### Custom Filters

- Introduce `QueryFilters` type which is just an "alias" to Dictionary in an attempt to improve DX 
```c#
public class QueryFilters
  : Dictionary<string, Func<IQueryable<TEntity>, IQueryable<TEntity>>
{ }
```

- The repository becomes responsible for loading the `QueryFilters` and applying them
- Any keys present in the `QueryFilters` will not be applied using the default behavior

#### Example 

```c#
public class FooResource : ResourceDefinition<Foo>
{
  protected override QueryFilters GetQueryFilters() => new QueryFilters { 
            { "key1", (query, value) => query.Select(x => x) },
            { "key2", (query, value) => query.Select(x => x) },
        };
}
 
```

Should be used to handle requests like:

```
/foos?filter[key1]=value1
```

---

### Default Sort

Allow a resource to have a default sort if no `sort` key is provided.

- Introduce `PropertySortOrder` type which is just an "alias" to ReadOnlyList in an attempt to improve DX 

```c#
public class PropertySortOrder<T>
  : ReadOnlyList<(Expression<Func<T, dynamic>>, SortDirection)>
{ }
```

- The repository becomes responsible for applying the default sort if no other sort has been requested

#### Example 

```c#
public class FooResource : ResourceDefinition<Foo>
{
  public override PropertySortOrder<Foo> DefaultSortOrder = new { 
     (foo => foo.Bar, SortDirection.Ascending),
     (foo => foo.Baz, SortDirection.Descending),
  }
}
```

jaredcnance:
Closes #293
Closes #384
Closes #402

Note: this is the first step to creating a path that allows removal of `[Attr]` altogether, but we're not quite there yet. For now you will be able to change models from:

```c#
[Attr("compound-name")]
public string CompoundName { get; set; }
```

To:

```c#
[Attr]
public string CompoundName { get; set; }
```

#### OtherChanges

- ContextGraphBuilder now allows resources to be added without specifying the resource name:

```c#
var builder = new ContextGraphBuilder();

// before
builder.AddResource<TestResource>("test-resources");

// after
builder.AddResource<TestResource>();
```

- ResourceDefinition contains custom filters and default sort

#### FEATURE

### TODO

- Assess and identify any breaking changes
- Tests on `ResourceDefinition`


wisepotato:
@jaredcnance  maybe a good idea to add auto-discovery in the docs, or am i missing a page?

jaredcnance:
@wisepotato see this: https://json-api-dotnet.github.io/JsonApiDotNetCore/usage/resource-graph.html#auto-discovery

Let me know if something is missing or if the organization could be improved.

PR #410, opened 2018-09-24
Nested sort - Init

### LINKED ISSUE #359: [Question] Sort by relationship attributes?
I've been working with JsonApiDotNetCore and EmberJS for a bit and everything has been working well so far, but I've run into something that I can't seem to find the answer for, so I figured I'd post here. On the JsonApi website it says

> Note: It is recommended that dot-separated (U+002E FULL-STOP, “.”) sort fields be used to request sorting based upon relationship attributes. For example, a sort field of author.name could be used to request that the primary data be sorted based upon the name attribute of the author relationship.

Does JsonApiDotNetCore currently support this, or does it plan to in the future? I didn't see anything on the documentation related to this and I have some models that I'd like to be able to sort by using the relationship values. If it isn't currently supported, do you have any advice on how it might be possible to get it working in the current build? Any help would be greatly appreciated.

milosloub:
Closes #359


milosloub:
@jaredcnance 
This handles only nested sort only, not sparse fields. I had problem with fields due to Binding nested classes and make Expression dynamic for building Select queries like:
`Select(i => new Item( Nested = new Nested()));`

There is no caching attributes (as we talk about it). This can be done later.
Please check

src/JsonApiDotNetCore/Internal/Query/FilterOperations.cs
```diff
@@ -1,7 +1,9 @@
 // ReSharper disable InconsistentNaming
+using System;
+
 namespace JsonApiDotNetCore.Internal.Query
 {
-    public enum FilterOperations
+    public enum FilterOperationsEnum
```
jaredcnance: No need for the rename here. I'm not too keen on appending the type classification (enum, class, struct) to the name (also, it's an unnecessary breaking change).

src/JsonApiDotNetCore/Internal/Query/FilterOperations.cs
```diff
@@ -15,4 +17,33 @@ public enum FilterOperations
...
+        {
+            if (prefix.Length == 0) return FilterOperationsEnum.eq;
+
+            if (Enum.TryParse(prefix, out FilterOperationsEnum opertion) == false)
+                throw new JsonApiException(400, $"Invalid filter prefix '{prefix}'");
+
+            return opertion;
+        }
+
+        public static string GetFilterOperationFromQuery(string query)
```
jaredcnance: Please add xml docs with `<example>` tags showing what the expected input/output should look like. I'm in the process of migrating documentation to docfx and want to make sure we have fully documented public APIs. This comment applies to all new `public` and `protected` (unless the owning class is `internal`)  methods.

e.g.:

```c#
/// <summary>
/// ...
/// </summary>
///
/// <example>
/// <code>
/// FilterOperations.GetFilterOperationFromQuery("eq:value").Dump();
/// // > "eq"
/// </code>
/// </example>
```

src/JsonApiDotNetCore/Internal/Query/FilterOperations.cs
```diff
@@ -15,4 +17,33 @@ public enum FilterOperations
...
+        {
+            var values = query.Split(QueryConstants.COLON);
+
+            if (values.Length == 1)
+                return string.Empty;
+
+            var operation = values[0];
+            // remove prefix from value
+            if (Enum.TryParse(operation, out FilterOperationsEnum op) == false)
```
jaredcnance: I'm thinking we should just return the parsed `FilterOperationsEnum` here. Once we've parsed the input, I think the only thing that should be passed around internally is the `FilterOperationsEnum` and not the `string`. Looking at one of the consumers I see things like:

```c#
string.Equals(op, FilterOperationsEnum.@in.ToString(), /* ... */);
```

which means we're parsing it, throwing it away and then forcing a subsequent string allocation when we could just check value equality.

Now this may have significant side-effects to the rest of the package—I don't know off the top of my head. But, I think this is the direction we should start taking it. What do you think?

src/JsonApiDotNetCore/Services/QueryParser.cs
```diff
@@ -199,53 +192,68 @@ protected virtual List<string> ParseFieldsQuery(string key, string value)
...
+            if (string.IsNullOrEmpty(s))
+            {
+                return string.Empty;
+            }
+            // Return char and concat substring.
+            return char.ToUpper(s[0]) + s.Substring(1);
+        }
 
-        protected virtual AttrAttribute GetAttribute(string propertyName)
+        protected virtual RelationshipAttribute GetRelationshipAttribute(string relationship)
```
jaredcnance: ☝️ breaking change. not saying it's a problem, but we need to call it out in the release notes. Ideally, we should try to leave the previous implementation (removing all internal references) and drop an `[Obsolete("Use ... instead")]` so users have some idea of how to fix their code.

src/JsonApiDotNetCore/Internal/Query/QuerySet.cs
```diff
@@ -8,6 +8,6 @@ public class QuerySet
         public PageQuery PageQuery { get; set; } = new PageQuery();
         public List<SortQuery> SortParameters { get; set; } = new List<SortQuery>();
         public List<string> IncludedRelationships { get; set; } = new List<string>();
-        public List<string> Fields { get; set; } = new List<string>();
```
jaredcnance: Same comment as before, I think this should remain and be deprecated, adding the new property as maybe `QueryFields` or something.

src/JsonApiDotNetCore/Internal/Query/FilterOperations.cs
milosloub: Ok

src/JsonApiDotNetCore/Internal/Query/FilterOperations.cs
milosloub: I was thinking about same think. This change will change FilterQuery and it's string operation {get;set;} to FilterOperations operation{get;set;}. Then few tests must be fixed (usage of new FilterQuery("attr","value","operation")). If it is ok, I'll fix this everywhere.

src/JsonApiDotNetCore/Services/QueryParser.cs
milosloub: If you look closer, there are both protected methods - GetAttribute and new GetRelationshipAttribute. I change methods order and diff is now confusing.

src/JsonApiDotNetCore/Internal/Query/QuerySet.cs
milosloub: Ok, no problem. I think it's better revert all changes connected with Fields ald make another issue.
There is one problem, that in QueryParser is no chance to get attribute of relationship attribute. Is it ok to inject IJsonApiContext here? (just for info for my next steps in case of fixing fields)

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/SparseFieldSetTests.cs
```diff
@@ -96,5 +97,40 @@ public async Task Fields_Query_Selects_Sparse_Field_Sets()
             Assert.Equal(todoItem.Description, deserializeBody.Data.Attributes["description"]);
             Assert.Equal(todoItem.CreatedDate.ToString("G"), ((DateTime)deserializeBody.Data.Attributes["created-date"]).ToString("G"));
         }
+
+        //[Fact]
```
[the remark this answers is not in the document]
milosloub: Commented code removed

test/JsonApiDotNetCoreExampleTests/Acceptance/TodoItemsControllerTests.cs
```diff
@@ -217,6 +223,82 @@ public async Task Can_Sort_TodoItems_By_Ordinal_Ascending()
...
+            var route = $"/api/v1/todo-items?include=owner&sort=owner.age";
+            var request = new HttpRequestMessage(httpMethod, route);
+
+            // Act
+            var response = await _fixture.Client.SendAsync(request);
+            var body = await response.Content.ReadAsStringAsync();
+            var deserializedBody = _fixture.GetService<IJsonApiDeSerializer>().DeserializeList<TodoItem>(body);
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
```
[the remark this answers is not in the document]
milosloub: Now assert on status code is right behind getting response

src/JsonApiDotNetCore/Internal/Query/FilterOperations.cs
milosloub: Renamed back

src/JsonApiDotNetCore/Internal/Query/FilterOperations.cs
milosloub: Deleted, finally this method is not necessary

src/JsonApiDotNetCore/Services/QueryParser.cs
milosloub: RelationshipAttribute() deleted for now, it was only due to Fields. GetAttribute method is still there

src/JsonApiDotNetCore/Internal/Query/QuerySet.cs
milosloub: Reverted back

milosloub:
Finally I removed implementations of AttrQuery and RelatedAttrQuery (AttrFilterQuery,AttrSortQuery, etc). Now these objects contains constructor overloading and it will build object base on passed query type (SortQuery, FilterQuery...folowing FieldsQuery). 
This leads to better chance of caching - now it's easy to store Attribute and RelationshipAttribute and for example if we create AttrQuery in Filter() method, then it can be fetched in Sort() method and provided only Direction property (not Attribute and RelationshipAttribute again via JsonApiContext). 
I'll make this caching later, I've got TODO :)

src/JsonApiDotNetCore/Internal/Query/FilterOperations.cs
milosloub: Logic was replaced by GetFilterOperationAndValue() method. Only problem here is that this breaks 
virtual ParseFilterOperation(). Because ParseFilterOperation() is called directly from ParseFilterQuery(), there is no chance to handle possible custom implementation of ParseFilterOperation() and be compatible with new GetFilterOperationAndValue().

Update:
I wrote comment below on QueryParser. So this and the orher comment is about same problem

milosloub:
@jaredcnance introducing new abstract BaseAttrQuery for reducing Sort, Filter and OrderBy operations to simpler calls. Now there is always only one private method for each call and also with respect to attibute/relationship.attribute logic.
Please check if you can

milosloub:
I have to turn most of the code back because of compatibility without previous implementation. Those dependencies makes my head blow. So now this should be without any breaking changes.

milosloub:
Closing this because of bad approach from the beginning. I had to start from the beginning with know how of each dependency of each query, attrquery, IJsonApiContext etc. I'll create new PR later with more complex solution.

Issue #404, opened 2018-09-11
Optimize Queries


btecu:
There are a few improvements that can be done to the auto generated endpoints:
 - GET endpoints could have `.AsNoTracking()` added to the query
 - DELETE endpoints could do `.Attach()` -> `.Remove()` instead of `.Get()` -> `.Remove()` to avoid making two calls
 - UPDATE could possibly be improved similarly to DELETE

jaredcnance:
Very good points!

#221 was opened to address your first point and was implemented in #225. However, I closed the PR because I didn't have cycles at that time to complete it. From the PR:

> This is on hold until I can evaluate the scope of the breaking changes. Not only are there API implications but also impacts on possible extensibility scenarios. By setting the query as NoTracking, acceptance tests using the same context may be impacted.
> Need to put together documentation on these breaking changes and how users should handle acceptance testing that uses the same DbContext instance.
>
> **Update**
> I think the best way to move forward is to allow a global options flag to be set in Startup that will enable AsNoTracking. It will need to default to false to avoid breaking changes and will allow tests to enable tracking for all queries.
> There may also be impact to #238

In retrospect, based on experience with other projects (including [DotnetTestAbstractions](https://github.com/jaredcnance/DotnetTestAbstractions)), I think tests should not share the same `DbContext` (in most cases -- but there is definitely a notable performance benefit) and should instead share the underlying connection and transaction. So, it would be less of a concern. However, it should still be configurable at first and then default in a later major version.

btecu:
How about having `GetRawAsync(bool isReadOnly)` and be called from the other endpoints?

Somewhat related, better naming could be nicer. Looking at `Get()` and `GetAsync()` is not obvious the difference. Both are `async` but the first returns a collection while the second one single record.

jaredcnance:
I _think_ your reference to GetRawAsync is actually the intent of Get. 

Get is not async. It doesn’t actually run a query since it returns IQueryable. It serves as the foundation for (most) other requests.

btecu:
Right, it could be called `GetRaw(bool isReadOnly)`.

jaredcnance:
I'm open to discussion on the name change. For me, "GetRaw" doesn't actually clear the situation up any and begs additional questions. Perhaps just `GetQueryable`?

The rollout might look like:

```c#
// 1st release
IQueryable<TEntity> Get() => GetRaw(isReadOnly: false);
[...]

PR #416, opened 2018-09-30
Nested sort

### LINKED ISSUE #359: [Question] Sort by relationship attributes?

milosloub:
Closes #359 

There is new approach of handling nested sort with refactor through all necessary layers.
I tried my best to keep compatibility with current implementations.


src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
```diff
@@ -28,68 +29,137 @@ private static MethodInfo ContainsMethod
...
+                throw new JsonApiException(400, $"It's not possible to provide {nameof(SortQuery)} without {nameof(SortQuery.SortedAttribute)} parameter." +
+                    $" Use Sort method with IJsonApiContext parameter instead.");
+
             return sortQuery.Direction == SortDirection.Descending
                 ? source.ThenByDescending(sortQuery.SortedAttribute.InternalAttributeName)
                 : source.ThenBy(sortQuery.SortedAttribute.InternalAttributeName);
         }
 
-        public static IOrderedQueryable<TSource> OrderBy<TSource>(this IQueryable<TSource> source, string propertyName)
+        public static IQueryable<TSource> Sort<TSource>(this IQueryable<TSource> source, IJsonApiContext jsonApiContext, List<SortQuery> sortQueries)
```
milosloub: Because of backward compatibility, I have to copy all three Sort methods with IJsonApiContext parameter.
It's important to distinguish whether the client use SortQuery with "old" SortedAttribute or new string based Attribute parameter. SortedAttribute depends on IJsonApiContext and string based Attribute not, so that's why this injection.

src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
```diff
@@ -28,68 +29,137 @@ private static MethodInfo ContainsMethod
...
         {
             // {x}
             var parameter = Expression.Parameter(typeof(TSource), "x");
-            // {x.propertyName}
-            var property = Expression.Property(parameter, propertyName);
-            // {x=>x.propertyName}
-            var lambda = Expression.Lambda(property, parameter);
+            MemberExpression member;
+
+            var values = propertyName.Split('.');
```
milosloub: To keep OrderBy, OrderByDescending, ThenBy and ThenByDescending method parameters without change, it's necessary to provide propertyName as path( "prop" or "rel.prop").

src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
```diff
@@ -100,126 +170,24 @@ public static IQueryable<TSource> Filter<TSource>(this IQueryable<TSource> sourc
...
-                }
-            }
-            catch (FormatException)
-            {
-                throw new JsonApiException(400, $"Could not cast {filterQuery.PropertyValue} to {property.PropertyType.Name}");
-            }
-        }
-
-        public static IQueryable<TSource> Filter<TSource>(this IQueryable<TSource> source, RelatedAttrFilterQuery filterQuery)
+        public static IQueryable<TSource> Filter<TSource>(this IQueryable<TSource> source, BaseFilterQuery filterQuery)
```
milosloub: BaseFilterQuery hier as ancestor of AttrFilterQuery and RelatedAttrFilterQuery, so all filtering can be handled in one method due to DRY

src/JsonApiDotNetCore/Services/QueryComposer.cs
```diff
@@ -30,8 +30,7 @@ public string Compose(IJsonApiContext jsonApiContext)
         private string ComposeSingleFilter(FilterQuery query)
         {
             var result = "&filter";
-            var operation = string.IsNullOrWhiteSpace(query.Operation) ? query.Operation : query.Operation + ":";
-            result += QueryConstants.OPEN_BRACKET + query.Attribute + QueryConstants.CLOSE_BRACKET + "=" + operation + query.Value;
+            result += QueryConstants.OPEN_BRACKET + query.Attribute + QueryConstants.CLOSE_BRACKET + "=" + query.OperationType + QueryConstants.COLON + query.Value;
```
milosloub: If we want to use only enum typed operations, there can't be empty string

src/JsonApiDotNetCore/Services/QueryParser.cs
```diff
@@ -85,32 +85,30 @@ protected virtual List<FilterQuery> ParseFilterQuery(string key, string value)
...
+                    var opVal = ParseFilterOperationAndValue(value);
+                    queries.Add(new FilterQuery(propertyName, opVal.value, opVal.operation));
                 }
             }
 
             return queries;
         }
 
+        [Obsolete("This method is not used anymore! Use " + nameof(ParseFilterOperationAndValue) + " method with FilterOperations operation return value." +
+            "Operation as string is not used at all.")]
```
milosloub: If we want to use FilterOperations base on enum, this method must be replaced by new ParseFilterOperationAndValue() which eliminates string operation value.
This is old cascade:
Parse() calls ParseFilterQuery() that calls ParseFilterOperation() which returns string op and string value
This is new cascade:
Parse() calls ParseFilterQuery() that calls ParseFilterOperationAndValue() which returns FilterOperations op and string value. This completely eliminates the ability to override the ParseFilterOperation(), and will be ignored.
Can be this the problem?

src/JsonApiDotNetCore/Services/QueryParser.cs
```diff
@@ -121,6 +119,63 @@ protected virtual (string operation, string value) ParseFilterOperation(string v
...
+        /// Parse filter operation enum and value by string value.
+        /// Input string can contain:
+        /// a) property value only, then FilterOperations.eq and value is returned
+        /// b) filter prefix and value e.g. "prefix:value", then FilterOperations.prefix and value is returned
+        /// In case of prefix is provided and is not in FilterOperations enum,
+        /// invalid filter prefix exception is thrown.
+        /// </summary>
+        /// <param name="input"></param>
+        /// <returns></returns>
+        public static (FilterOperations operation, string value) ParseFilterOperationAndValue(string input)
```
milosloub: Optimized parsing values and operations in one method. It's combination of ParseFilterOperation and GetFilterOperationOld methods.

test/UnitTests/Services/QueryComposerTests.cs
```diff
@@ -56,7 +56,7 @@ public void Can_ComposeLessThan_FilterStringForUrl()
             // act
             var filterString = queryComposer.Compose(_jsonApiContext.Object);
             // assert
-            Assert.Equal("&filter[attribute]=le:value&filter[attribute2]=value2", filterString);
+            Assert.Equal("&filter[attribute]=le:value&filter[attribute2]=eq:value2", filterString);
```
milosloub: here again - empty string operation are not allowed

test/UnitTests/Services/QueryParser_Tests.cs
```diff
@@ -99,7 +100,7 @@ public void Filters_Properly_Parses_DateTime_Without_Operation()
 
             // assert
             Assert.Equal(dt, querySet.Filters.Single(f => f.Attribute == "key").Value);
-            Assert.Equal(string.Empty, querySet.Filters.Single(f => f.Attribute == "key").Operation);
+            Assert.Equal(FilterOperations.eq, querySet.Filters.Single(f => f.Attribute == "key").OperationType);
```
milosloub: Same use case as in the comment above

milosloub:
@jaredcnance Hi, I fixed branche conflicts and found code like this:

```            if (_resourceDefinition != null)
            {
                var defaultSortOrder = _resourceDefinition.DefaultSort();
                if (defaultSortOrder != null && defaultSortOrder.Count > 0)
                {
                    foreach (var sortProp in defaultSortOrder)
                    {
                        // this is dumb...add an overload, don't allocate for no reason
                        entities.Sort(new SortQuery(sortProp.Item2, sortProp.Item1));
                    }
                }
            }
```
I think this pull request can solve this DefaultSort() by using SortQuery with new constructor overload. Please check before release stable 3.0 version.

src/JsonApiDotNetCore/Services/QueryParser.cs
```diff
@@ -121,6 +119,63 @@ protected virtual (string operation, string value) ParseFilterOperation(string v
...
+                return (operation, value);
+            }
+        }
+
+        /// <summary>
+        /// Returns typed operation result and info about parsing success
+        /// </summary>
+        /// <param name="operation">String represented operation</param>
+        /// <returns></returns>
+        public static (FilterOperations operation, bool succeeded) GetFilterOperation(string operation)
```
jaredcnance: Should be `private`. But, is this method really necessary? Why not just `Enum.TryParse(operation, out FilterOperations opertion)`?

jaredcnance:
> I think this pull request can solve DefaultSort() by using SortQuery with new constructor overload. Please check before release stable 3.0 version.

I'm not sure what you mean here. Can you elaborate? If you'd like a quicker mode of communication, feel free to ping me on Gitter.

test/UnitTests/Services/QueryComposerTests.cs
jaredcnance: We have to allow queries with default equality operations. This will be way too much work for some code bases.

milosloub:
@jaredcnance Hi, I've got one suggestion - what about reverting all filter stuff and make focus only on nested sort feature? Most of comments refer to filtering, so maybe it would be better to split this PR to two separate PR and deliver nested sort as soon as posible. After while I'll contact you on gitter - good idea

src/JsonApiDotNetCore/Internal/Query/AttrFilterQuery.cs
```diff
@@ -6,32 +7,25 @@ namespace JsonApiDotNetCore.Internal.Query
...
     public class AttrFilterQuery : BaseFilterQuery
     {
-        private readonly IJsonApiContext _jsonApiContext;
-
         public AttrFilterQuery(
             IJsonApiContext jsonApiContext,
             FilterQuery filterQuery)
+            : base(jsonApiContext, 
+                  null,
+                  filterQuery.Attribute,
```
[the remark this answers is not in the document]
milosloub: Yes, first solution was different ( no class BaseAttrQuery). Now I can pass filterQuery and BaseAttrQuery will have BaseQuery in constructor parameters. Good point!

src/JsonApiDotNetCore/Services/QueryComposer.cs
milosloub: Exactly, that was my thought

src/JsonApiDotNetCore/Services/QueryParser.cs
milosloub: Ok, if it is problem, I can revert all fiter stuff. For now I changed access attribute to protected virtual and cleaned up language - sorry for my English representation, I'm from Czech republic (Europe) 😄

test/UnitTests/Services/QueryComposerTests.cs
milosloub: Ok. What about QueryComposer without change and don't make Operation property (FilterQuery) obsolete?

src/JsonApiDotNetCore/Internal/Query/AttrFilterQuery.cs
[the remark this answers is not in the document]
milosloub: Changed

src/JsonApiDotNetCore/Internal/Query/AttrFilterQuery.cs
```diff
@@ -6,32 +7,25 @@ namespace JsonApiDotNetCore.Internal.Query
...
+            FilteredAttribute = Attribute;
         }
 
-        public AttrAttribute FilteredAttribute { get; }
-        public string PropertyValue { get; }
-        public FilterOperations FilterOperation { get; }
-
-        private AttrAttribute GetAttribute(string attribute) =>
-            _jsonApiContext.RequestEntity.Attributes.FirstOrDefault(attr => attr.Is(attribute));
+        [Obsolete("Use " + nameof(Attribute) + " property of " + nameof(BaseAttrQuery) + "class. This property is shared for all AttrQuery and RelatedAttrQuery (filter,sort..) implementations.")]
```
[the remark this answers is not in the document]
milosloub: Fixed

src/JsonApiDotNetCore/Internal/Query/BaseAttrQuery.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// It elimines boilerplate of providing specified type(AttrQuery or RelatedAttrQuery) 
+    /// while filter and sort operations and eliminates plenty of methods to keep DRY principles 
+    /// </summary>
+    public abstract class BaseAttrQuery
+    {
+        private readonly IJsonApiContext _jsonApiContext;
+
+        public BaseAttrQuery(IJsonApiContext jsonApiContext, string relationship, string attribute)
+        {
+            _jsonApiContext = jsonApiContext;
```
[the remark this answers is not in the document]
milosloub: Maybe I don't understand, but all the time AttrFilterQuery and RelatedAttrQuery depends on IJsonApiContext, so why IContextGraph?

src/JsonApiDotNetCore/Internal/Query/BaseQuery.cs
```diff
@@ -0,0 +1,26 @@
+using JsonApiDotNetCore.Models;
+using JsonApiDotNetCore.Services;
+using System;
+using System.Linq;
+
+namespace JsonApiDotNetCore.Internal.Query
+{
+    public abstract class BaseQuery
```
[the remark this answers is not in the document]
milosloub: Because concept is same as BaseAttrQuery, that only holds info about attributes, relationship etc. I don't know about use case, when client should create new instance of this BaseQuery, only descendant class (FilterQuery, SortQuery). Or do you have some idea where to use it without abstract?

src/JsonApiDotNetCore/Internal/Query/FilterQuery.cs
```diff
@@ -1,23 +1,40 @@
...
+            Operation = operationType.ToString();
+            OperationType = operationType;
         }
-        
+
         [Obsolete("Key has been replaced by '" + nameof(Attribute) + "'. Members should be located by their public name, not by coercing the provided value to the internal name.")]
         public string Key { get; set; }
-        public string Attribute { get; }
         public string Value { get; set; }
+        [Obsolete("Operation has been replaced by '" + nameof(OperationType) + "'. OperationType is typed enum value for Operation property. This should be default property for providing operation type, because of unsustainable string (not typed) value.")]
```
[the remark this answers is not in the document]
milosloub: Fixed

src/JsonApiDotNetCore/Internal/Query/RelatedAttrFilterQuery.cs
```diff
@@ -6,45 +7,28 @@ namespace JsonApiDotNetCore.Internal.Query
...
 
-            FilteredRelationship = relationship;
-            FilteredAttribute = attribute;
-            PropertyValue = filterQuery.Value;
-            FilterOperation = GetFilterOperation(filterQuery.Operation);
+            FilteredRelationship = Relationship;
+            FilteredAttribute = Attribute;
         }
 
+        [Obsolete("Use " + nameof(Attribute) + " property. It's shared for all implementations of BaseAttrQuery(better sort, filter) handling")]
```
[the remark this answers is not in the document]
milosloub: Fixed

src/JsonApiDotNetCore/Services/QueryParser.cs
```diff
@@ -199,35 +248,16 @@ protected virtual List<string> ParseFieldsQuery(string key, string value)
...
-            }
-            catch (InvalidOperationException e)
-            {
-                throw new JsonApiException(400, $"Attribute '{propertyName}' does not exist on resource '{_controllerContext.RequestEntity.EntityName}'", e);
-            }
-        }
-
-        private string GetFilterOperation(string value)
+        [Obsolete("Delete also when " + nameof(ParseFilterOperation) + " deleted.")]
+        private string GetFilterOperationOld(string value)
```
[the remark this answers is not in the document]
milosloub: Fixed

src/JsonApiDotNetCore/Services/QueryParser.cs
```diff
@@ -121,6 +119,63 @@ protected virtual (string operation, string value) ParseFilterOperation(string v
...
+            if (input.Length == 0)
+                return (FilterOperations.eq, input);
+
+            // split value
+            var values = input.Split(QueryConstants.COLON);
+            // value only
+            if (values.Length == 1)
+                return (FilterOperations.eq, input);
+            // prefix:value
+            else if (values.Length == 2)
```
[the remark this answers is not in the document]
milosloub: ?filter[prop]=some:value, is it really posible? What about GetFilterExpressionLambda in IQueryableExtensions...if I understand it correctly, if client provides custom filter operation, it will be stopped here.

src/JsonApiDotNetCore/Services/QueryParser.cs
milosloub: Fixed

src/JsonApiDotNetCore/Internal/Query/BaseFilterQuery.cs
```diff
@@ -1,9 +1,24 @@
...
+            string attribute,
+            string value,
+            FilterOperations op)
+        : base(jsonApiContext, relationship, attribute)
+        {
+            PropertyValue = value;
+            FilterOperation = op;
+        }
+
+        [Obsolete("To resolve operation use enum typed " + nameof(FilterQuery.OperationType) + " property of "+ nameof(FilterQuery) +" class")]
```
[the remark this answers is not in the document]
milosloub: Fixed

milosloub:
Reverted filter changes back. There is no missing property, in some cases property is moved to new ancestor only

Spacing changes only in:
- QueryParser_Tests.cs
- QueryComposerTests.cs
- SparseFieldSetTests.cs
- QuerySet.cs
You can ignore these files

milosloub:
Reverted filter changes back.
Changes description:
1) Person.cs -> only new Age property for testing purposes
2) DefaultEntityRepository.cs -> Sort() call with IJsonApiContext param - same approach as Filter()
3) IQueryableExtensions.cs -> most changes. 
   -New Sort methods with IJsonApiContext param, old ones as obsolete, 
   -One Filter() method instead of two - now require BaseFilterQuery param instead of AttrFilterQuery 
     and RelatedAttrFilterQuery.
   -Add methods CallGenericWhereContainsMethod() and CallGenericWhereMethod() to keep 
     convention just likeCallGenericOrderMethod()
4) AttrFilterQuery - GetAttribute() now in ancestor BaseAttrQuery, PropertyValue and FilterOperation  moved to BaseFilterQuery (neccessary due to IQueryableExtension.Filter() method)
5) new AttrSortQuery - same approach as AttrFilterQuery
6) new BaseAttrQuery - shared ancestor, that handles getting Attribute, Relationship from IJsonApiContext
7) BaseFilterQuery - new ancestor BaseAttrQuery. PropertyValue and FilterOperation are here as shared properties for AttrFilterQuery and RelatedAttrFilterQuery
8) new BaseQuery.cs - ancestor of all Queries( FilterQuery, SortQuery...). Holds info about atribute, relationship and if attribute belongs to relationship
9) FilterQuery - new ancestor BaseQuery. Attribute and IsAttributeOfRelationship now in BaseQuery
10) QueryConstants - new DOT constant
11) RelatedAttrFilterQuery - same as point 4. + GetRelationship() now in ancestor BaseAttrQuery
12) new RelatedAttrSortQuery - same approach as RelatedAttrFilterQuery
13) SortQuery - new constructor overload - just like FilterQuery only require string attribute - this is the key for nested sort feature. Previous constructor as obsolete
14) QueryParser - building SortQuery by new constructor overload. BuildFilterQuery deleted
15) TodoItemsControllerTests - Added two tests for nested sort feature

Spacing changes only in:
- QueryParser_Tests.cs
- QueryComposerTests.cs
- SparseFieldSetTests.cs
- QuerySet.cs

You can ignore these files

src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
```diff
@@ -28,68 +28,137 @@ private static MethodInfo ContainsMethod
...
 
             return orderedEntities;
         }
 
+        [Obsolete("Use Sort method with IJsonApiContext parameter instead. New Sort method provides nested sorting.")]
         public static IOrderedQueryable<TSource> Sort<TSource>(this IQueryable<TSource> source, SortQuery sortQuery)
         {
+            // For clients using SortQuery constructor with string based parameter
+            if (sortQuery.SortedAttribute == null)
+                throw new JsonApiException(400, $"It's not possible to provide {nameof(SortQuery)} without {nameof(SortQuery.SortedAttribute)} parameter." +
```
jaredcnance: 3 things:
- Does this need to be added? It looks like `SortedAttribute` is deprecated, but it looks like we're still setting the value. Has something else changed, or does this handle an error case that wasn't handled properly in the current version and would have resulted in a null ref exception below?
- The language refers to internal method usage (i.e. it's directed at the application developer), but a 400 is returned. If this is a request error, the language should be targeted towards the client / client application: e.g. "Cannot sort on an empty attribute". OR if this is a developer concern it should return a 500. 
- Is this something that should be handled earlier in the request lifecycle?

Still reviewing so it's possible I'm missing something.

src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
```diff
@@ -28,68 +28,137 @@ private static MethodInfo ContainsMethod
...
         }
 
-        public static IOrderedQueryable<TSource> OrderBy<TSource>(this IQueryable<TSource> source, string propertyName)
+        public static IQueryable<TSource> Sort<TSource>(this IQueryable<TSource> source, IJsonApiContext jsonApiContext, List<SortQuery> sortQueries)
         {
-            return CallGenericOrderMethod(source, propertyName, "OrderBy");
+            if (sortQueries == null || sortQueries.Count == 0)
+                return source;
+
+            var orderedEntities = source.Sort(jsonApiContext, sortQueries[0]);
```
jaredcnance: why do we need this here rather than run it all in the `for` loop?

src/JsonApiDotNetCore/Internal/Query/BaseAttrQuery.cs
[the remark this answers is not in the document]
jaredcnance: yes, you're correct 😞 , the idea was to only depend on the interface that you actually need. #253 elaborates on this a bit more.

src/JsonApiDotNetCore/Internal/Query/BaseAttrQuery.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Linq;
+
+namespace JsonApiDotNetCore.Internal.Query
+{
+    /// <summary>
+    /// Abstract class to make available shared properties of all query implementations
+    /// It elimines boilerplate of providing specified type(AttrQuery or RelatedAttrQuery) 
+    /// while filter and sort operations and eliminates plenty of methods to keep DRY principles 
+    /// </summary>
+    public abstract class BaseAttrQuery
```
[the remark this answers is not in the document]
jaredcnance: ah I see, generally I only use `abstract` when I have `abstract` members and would use a `protected` constructor instead, but i think this is perfectly fine

src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
milosloub: 1) It's not necessary. I did it because of compatible solution for clients, that use SortQuery instance with SortedAttribute. I have question - do you use deprecations like "It's not use at all, fix it to new object"? If yes, it's better to me - then I can have 3 Sort methods instead of 6 and use it only with IJsonApiContext parameter. But this leads to one problem - developers, that calls IQueryable.Sort() method receive error during build time (because of missing obligatory IJsonApiContext param). What do you think about this? 
2) If "deprecation question" in point 1. is valid, this makes no sense. Otherwise I change 400 to 500 - it's definitely developer concern.
3) This also depends on point 1.

src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
milosloub: It's current implementation, from me there is no change, please check. Also I think this approach is fine

src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
jaredcnance: Yes, we use deprecations based on the following strategy (where possible):

- Introduce new APIs and deprecate with functionality intact using an `[Obsolete(..., error: false)]`
- Fully deprecate in a major release with `[Obsolete(..., error: true)]`
- Remove API in next major release

This was originally documented [here](https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/404#issuecomment-420483020) and is being finalized as part of #290.

So, ideally we would deprecate and leave the existing path fully intact so developers can upgrade with minimal impact to their builds. Then in the next major release we force a build error. However, I would rather force a new build error than a new runtime error.

In this case, the deprecation is on the `SortQuery` but the old functionality should still be fine, correct? If so, then the deprecation on the method itself should be sufficient. As far as I can tell this is not a 400 error. So, ultimately this comes down to a single question:

Was a new failure mode introduced for this method? i.e. was something added that would cause `SortedAttribute` to be null? If the answer is no, then the exception is _fine_ just make it a 500. If there is a new failure mode, then I need to understand what it is.

milosloub: Ok, that deprecation strategy makes sence. So now it's made as the first point [Obsolete(..., error: false)]
I would rather also jump to step two  [Obsolete(..., error: true)], because of eliminating bad constructor usage of SortQuery combined with bad Sort() method.

Explanation:
- Before PR - QueryParser creates SortQuery list based on SortedAttribute, then 
 DefaultEntityRepository calls Sort(`List<SortQuery>`)
-  After PR - QueryParser creates SortQuery list based on **Attribute( as string)**, then DefaultEntityRepository calls Sort(`IJsonApiContext, List<SortQuery>`)

So everything is fine and old functionality is kept. But JsonApiException will be thrown only in this specific case: Update to new version -> developer creates new SortQuery instance somewhere in his code -> use inappropriate Sort() method. This means that for standard implementations everything is OK, but for custom implementations this "criss cross" usage may occur.

I see two solutions here:
1) Make [Obsolete(..., error: true)], then JsonApiException is not neccessary at all
2) Keep it as it is and only change JsonApiException from 400 to 500.

What do you think about this?

src/JsonApiDotNetCore/Internal/Query/BaseAttrQuery.cs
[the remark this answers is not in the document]
milosloub: Yes, it depends on use case. I was inspired by EF approach and it's table inheritance - abstract base classes holds shared properties for all descendants and prevents creating new instances directly

https://www.learnentityframeworkcore.com/inheritance/table-per-hierarchy

src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
jaredcnance: Okay thanks for the explanation. Let's go with option #1. I'd prefer a build time error with a clear explanation over a new runtime exception.

wisepotato:
Oh man this looks good!

Issue #425, opened 2018-10-11
`[FromBody]` not working on PatchAsync


wisepotato:
## Description

When defining the Patch behaviour you have used:
```c#
       [HttpPatch("{id}")]
        public override async Task<IActionResult> PatchAsync(TId id, [FromBody] T entity)
            => await base.PatchAsync(id, entity);
```
Which is fine, however invalid. The tests are not catching this because (I believe) there are no end-to-end tests?

The correct input according to the json:api spec is to change the name of a team:
```json
{  
   data:{  
      id:"1",
      type:"team",
      attributes:{  
         Name:"new-name"
      }
   }
}
```
However, because of the previous code, the following works:
```json
{  
   Name:"Joe",
}
```
Maybe a container around `T` is the answer? If you have any insights on how to solve this problem I will fix this as I need this fixed ASAP.

## Environment

- JsonApiDotNetCore Version:  V3.0.0-beta5
- Other Relevant Package Versions:

wisepotato:
I keep doing this, forgot the correct content-type and my resource was not properly identified as a resource, the errors in the console were more than capable of explaining it to me. Can be closed.

wisepotato:
Alright, more information. I debugged the hell out of `JsonApiDotNetCoreExampleTests.Acceptance.Can_Fetch_Many_To_Many_Through` and it seems that the  body that is being generated does **not** include an `includes` key. This is also signified by the fact that the tag is not properly fetched.

This seems to support my theory, and gave me a headache :)

Atleast I now know your pipeline a bit better. Cool beans.

maurei:
#426 contains two tests that are different from before:
`Can_Fetch_Many_To_Many_Through` =>  /api/v1/articles?include=tags
and 
`Can_Fetch_Many_To_Many_Through_Nested`     => /api/v1/articles/{article.Id}?include=tags

In both cases, the actual response body does not include a "include" section in the DocumentData.

The deserializer does convert the response body to a `Article` object with a `Tag` relation, but this `Tag` instance is only partially filled (as deduced from the ArticleTag jointable, i.e. just the `id` is filled, and `name` is `null`)

PR #426, opened 2018-10-12
Many to many improved tests that show problem with includes

maurei:
An improvement of the tests that fails with respect to issue #425


maurei:
Closing this one because #425 replaces this PR

PR #427, opened 2018-10-14
fix(#425): fetching HasManyThrough returns no includes

maurei:
#### BUG FIX
Closes issue #425 
- [X] fix issue

Replaces pull request #426, which can be ignored and deleted.


maurei:
I don't know why the two new tests are failing. Locally they run fine.

Issue #437, opened 2018-10-24
Bulk operations: create with relations


maurei:
## Bulk add and setting relation simultaneously
 Are bulk operations possible when simultaneously creating relations ? 

[In this test](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/2221d577581a3a314e55a99e5cb4ae6c61c5f6fa/test/OperationsExampleTests/Add/AddTests.cs#L98) articles are added while also adding authors using local ids.

However, my current case is a bit different. Consider the request below

```json
PATCH api/bulk HTTP/1.1
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json
{
  "operations": {
    "op": "add",
    "data": {
      "type": "article",
      "attributes": {
        "title": "Ember Hamster",
        "content": "some content"
      },
      "relationships": {
        "author": {
          "data": {
            "type": "people",
            "id": "9"
          }
        }
      }
    }
  }
}
```
I.e. I am not creating a new author, I'm only trying to set a relation to an already existing author. ([This is possible in normal requests, see these specs ](https://jsonapi.org/format/#crud-creating)).

Is this currently supported?

Is it correct that [this line](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/2221d577581a3a314e55a99e5cb4ae6c61c5f6fa/src/JsonApiDotNetCore/Services/Operations/Processors/CreateOpProcessor.cs#L53) suggests that it is not supported, because the `included` argument is not being passed?

If this isn't supported, any pointers on what to override?

## Environment
- JsonApiDotNetCore Version: latest beta

maurei:
I have been trying to reproduce my issue by covering the missing testcases (#438). I have not succeeded.

In my local project the equivalent of the `Task Can_Create_Articles_With_Existing_Author` test that I wrote fails. 
* In my local app, the deserialization result is equivalent to that of the result for the passing test, nothing unexpected here; all data is still contained (the return value of [this function call here](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/2221d577581a3a314e55a99e5cb4ae6c61c5f6fa/src/JsonApiDotNetCore/Services/Operations/Processors/CreateOpProcessor.cs#L53). 
* Also the result of the `createAsync` method below that line is fine: relations are included
* Therefore, in fact, the response of the api/bulk request even includes all related objects, in my case: mail_to_record_folder.client != null for all operationResults in the response.
[...]

PR #438, opened 2018-10-26
test: bulk operations add while relating to existing entities

maurei:
Some tests in attempt to reproduce the issue of #437. 

So far I have been unsuccessful.

To be continued.

#### BUG FIX
- [x] reproduce issue in tests (attempt to, anyway missing test cases)


maurei:
AH crap didnt notice that. Yes will do!

PR #446, opened 2018-11-13
fix/#445

### LINKED ISSUE #445: Can't create resource with relationship in Entity-Resource separation mode
Consider following:

Entities:
```
    {
        public Parent Parent { get; set; }
        public int ParentId { get; set }
    }
```
```
    public class Parent: Identifiable
    {
    }
```

Resources:
```
    public class ClientResource : Identifiable
    {
        [Attr]
        [Attr]
        [HasOne]
        public ParentResource Parent { get; set; }
    }
```
```
    public class ParentResource: Identifiable
    {
        [Attr]
        [Attr]
    }
```

Three Layer flow (with Automapper Profile)
`1) Controller - JsonApiController<ClientResource>`
`2) Service - EntityResourceService<ClientResource, Client, int>`
`3) Repository - DefaultEntityRepository<Client>`

If I POST request like this:
```
{
    "data": {
        "attributes": {
            "first-name": "Homer",
            "last-name": "Simpson"
        },
        "relationships": {
            "parent": {
                "data": {
                    "type": "parents",
                    "id": "1"
                }
            }
        },
        "type": "clients"
    }
}
```
the InvalidOperationException is thrown, because:
"The entity type 'ParentResource' was not found. Ensure that the entity type has been added to the model."
This exception points to line 234 of DefaultEntityRepository and it's AttachHasOnePointers method.

Do you have any idea, how to handle this? Does this mean, that Repository should be responsible for proper resolve of navigation property type?

roblankey:
Closes #445 

#### BUG FIX
- [X] reproduce issue in tests
- [X] fix issue
- [X] bump package version


src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
```diff
@@ -227,12 +237,27 @@ private void AttachHasManyThrough(TEntity entity, HasManyThroughAttribute hasMan
...
         /// </summary>
-        private void AttachHasOnePointers()
+        private void AttachHasOnePointers(TEntity entity)
         {
             var relationships = _jsonApiContext.HasOneRelationshipPointers.Get();
             foreach (var relationship in relationships)
-                if (_context.Entry(relationship.Value).State == EntityState.Detached && _context.EntityIsTracked(relationship.Value) == false)
-                    _context.Entry(relationship.Value).State = EntityState.Unchanged;
+            {
+                if (relationship.Key.GetType() != typeof(HasOneAttribute))
```
jaredcnance: Under what circumstances does this happen?

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
```diff
@@ -227,12 +237,27 @@ private void AttachHasManyThrough(TEntity entity, HasManyThroughAttribute hasMan
...
+                if (hasOne.EntityPropertyName != null)
+                {
+                    var relatedEntity = entity.GetType().GetProperty(hasOne.EntityPropertyName)?.GetValue(entity);
+                    if (relatedEntity != null && _context.Entry(relatedEntity).State == EntityState.Detached && _context.EntityIsTracked((IIdentifiable)relatedEntity) == false)
+                        _context.Entry(relatedEntity).State = EntityState.Unchanged;
+                }
+                else
+                {
+                    if (_context.Entry(relationship.Value).State == EntityState.Detached && _context.EntityIsTracked(relationship.Value) == false)
+                        _context.Entry(relationship.Value).State = EntityState.Unchanged;
```
jaredcnance: Let's refactor this a bit so it's clearer and we don't have separate definitions for how the EF ChangeTracker entry state is modified. The only thing that needs to be in the conditional is how we determine the  `relatedEntity`. so, maybe something like: 

```c#
IIdentifiable relatedEntity = (hasOne.EntityPropertyName != null)
    ? entity.GetType().GetProperty(hasOne.EntityPropertyName)?.GetValue(entity) as IIdentifiable
    : relationship.Value;

if(relatedEntity != null 
    && _context.Entry(relatedEntity).State == EntityState.Detached 
    && _context.EntityIsTracked(relatedEntity) == false) {
        _context.Entry(relatedEntity).State = EntityState.Unchanged;
}
```

roblankey:
Hah. Yeah, I was thinking the same thing last night @jaredcnance 
The more I mess with the entity/resource separation, the more I want to just update ContextG..., I mean ResourceGraph =), with the ability to track both there.

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
roblankey: This is totally just a guard clause since the dictionary key is RelationshipAttribute.

jaredcnance:
So, after thinking about this a little bit more, I realized that this is probably the serializer's responsibility. In order to set these relationships using a pointer, we have to create an instance of the relationship. So, what's happening right now (before this PR) is that the [`ResourceObject`](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/cd02b20dfee084fbaf9c4283af6102afa52b0e0d/src/JsonApiDotNetCore/Models/ResourceObject.cs#L6) is the object that the json payload gets deserialized into directly and we then instantiate an instance of the the related type and assign it to the navigation property. Then the repository detaches it so EF doesn't try to create it again. So, the path in the serializer looks like:

- `SetHasOneNavigationPropertyValue`
https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/8570b216c7ede3a7b5f9fc7a4912404a14926df5/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs#L264-L266
- which calls into `GetIncludedRelationship` to create an instance of the related type:
https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/8570b216c7ede3a7b5f9fc7a4912404a14926df5/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs#L305-L308

So, this is where I think we need to instantiate the correct type.

milosloub:
@jaredcnance 
Looks good. Just for sure - do you know, if it's possible with this approach to have multi relations pointing to same entity?
For example this:
```
    public class Resource: Identifiable
    {
        [HasOne("parent", targetEntity: typeof(Resource))]
        public Resource Parent { get; set; }

        [HasOne("child", targetEntity: typeof(Resource))]
        public Resource Child { get; set; }
    }
```

roblankey:
> So, after thinking about this a little bit more, I realized that this is probably the de-serializer's responsibility. In order to set these relationships using a pointer, we have to create an instance of the related type. So, what's happening right now (before this PR) is that the [`ResourceObject`](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/cd02b20dfee084fbaf9c4283af6102afa52b0e0d/src/JsonApiDotNetCore/Models/ResourceObject.cs#L6) is the object that the json payload gets deserialized into directly and we then instantiate an instance of the the related type and assign it to the navigation property. Then the repository detaches it so EF doesn't try to create it again. So, the path in the de-serializer looks like:
> 
> * `SetHasOneNavigationPropertyValue`
>   
>     
>       
>         [JsonApiDotNetCore/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/8570b216c7ede3a7b5f9fc7a4912404a14926df5/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs#L264-L266)
>       
>       
>           Lines 264 to 266
>         in
>         [8570b21](/json-api-dotnet/JsonApiDotNetCore/commit/8570b216c7ede3a7b5f9fc7a4912404a14926df5)
>       
>       
>       
>       
>   
>           
>             
>              var includedRelationshipObject = GetIncludedRelationship(rio, included, hasOneAttr); 
>           
>   
>           
>             
>              if (includedRelationshipObject != null) 
>           
>   
>           
>             
>                  hasOneAttr.SetValue(entity, includedRelationshipObject);
> * which calls into `GetIncludedRelationship` to create an instance of the related type:
>   
>     
>       
>         [JsonApiDotNetCore/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/8570b216c7ede3a7b5f9fc7a4912404a14926df5/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs#L305-L308)
>       
>       
>           Lines 305 to 308
>         in
>         [8570b21](/json-api-dotnet/JsonApiDotNetCore/commit/8570b216c7ede3a7b5f9fc7a4912404a14926df5)
>       
>       
>       
>       
>   
>           
>             
>              private IIdentifiable GetIncludedRelationship(ResourceIdentifierObject relatedResourceIdentifier, List<ResourceObject> includedResources, RelationshipAttribute relationshipAttr) 
>           
>   
>           
>             
>              { 
>           
>   
>           
>             
>                  // at this point we can be sure the relationshipAttr.Type is IIdentifiable because we were able to successfully build the ResourceGraph 
>           
>   
>           
>             
>                  var relatedInstance = relationshipAttr.Type.New<IIdentifiable>();
> 
> So, this is where I think we need to instantiate the correct type. So, maybe something like:
> 
> ```cs
> var relatedEntityType = relationshipAttr?.TargetEntityType ?? relationshipAttr.Type;
> var relatedInstance = relatedEntityType.New<IIdentifiable>();
> ```
> 
> and then the `HasOne` API becomes something like:
> 
> ```cs
> [HasOne("department", targetEntity: typeof(Department))]
> public DepartmentResource Department { get; set; }
> ```
> 
> What do you think?

So I was thinking about this more last night. The problem is with Resource/Entity separation, which means that the Resource isn't necessarily the same thing as the Entity when it gets to the Repository. We would have to expose the IResourceMapper further up.

But then we're already doing the resource mapping within the ResourceService so I started thinking about what we're trying to do with this whole pointers thing and I realized that, since the spec doesn't support creating the relationship entity on a POST, that all we really need to do it grab the ID from the Entity at the resource level, look up the affected relationship properties (that we can identify via the EntityPropertyName of the attribute) and detach them. 

The question that I really have is whether or not EF will even care. I didn't think it would update the navigation property attributes by default on a save. I'm going to do some testing outside of the project to confirm that, because if that's the case, we don't need this at all I don't think.

roblankey:
Hm. So my testing confirmed that, by default, EF won't try to update the members of the navigation property, but that within our setup, it does. @jaredcnance, I'm not a huge EF guy. Do you happen to know what we've got that's causing that? I wonder if it's not a better strategy to just not cross apply the updates like that.

jaredcnance:
@roblankey can you push a repo with your test and I can see what the difference is? But, generally EF will try to create both resources if I do:
```c#
var article = new Article { Author = new Author() };
dbContext.Articles.Add(article);
await dbContext.SaveChangesAsync();
```

roblankey:
@jaredcnance 

Just had an "A-Ha" moment. My original tests were only populating the navigation property identifier. When I populate the the navigation property entity as well, with only that Id, EF does throw that exception.

To that end, what do you think about the DeSerializer not serializing HasOne / HasMany inputs?, instead just utilizing the identifers?

roblankey:
The relevant pieces of the spec:

> If a relationship is provided in the relationships member of the resource object, its value MUST be a relationship object with a data member. The value of this key represents the linkage the new resource is to have.

`
Resource Linkage
Resource linkage in a compound document allows a client to link together all of the included resource objects without having to GET any URLs via links.

Resource linkage MUST be represented as one of the following:

null for empty to-one relationships.
an empty array ([]) for empty to-many relationships.
a single resource identifier object for non-empty to-one relationships.
an array of resource identifier objects for non-empty to-many relationships.
`

jaredcnance:
There's a reason we had to do it that way (which I can't remember right now -- will comment back later when I have a better answer and more time). But, how would that work with EF and HasMany relationships since the target resource is the independent side of the relationship (no FKs).

roblankey:
Cool, would be good to know if we ran into other issues. Good question on the HasMany, hadn't thought of that yet. That may make us detach anyway. I'll run some more tests as well.

roblankey:
Got the HasMany piece complete! Sorry it took so long!

PR #452, opened 2018-11-22
Feat/#451

### LINKED ISSUE #451: Ability to turn off self and related links globally
## Description

If you provide [HasOne] or [HasMany] attribute, self and related links are presented by default.
Example:

```
{
  "type": "articles",
  "id": "4309",
  "relationships": {
     "author": {
       "links": {
         "self": "/api/v1/articles/4309/relationships/author",
         "related": "/api/v1/articles/4309/author"
       }
     }
  }
}
```
To turn them off, now it's necessary to provide extra parameter - [HasOne(documentLinks: Link.None)].

To turn them off globaly, there should be something like:
`options.DisableSelfAndRelatedLinks = true;`
With output:
```
{
  "type": "articles",
  "id": "4309",
  "relationships": {
     "author": { }
     }
  }
}
```

## Environment
- JsonApiDotNetCore Version: 3.0.0

milosloub:
Closes #451


src/JsonApiDotNetCore/Configuration/JsonApiOptions.cs
```diff
@@ -102,6 +102,16 @@ public class JsonApiOptions
...
 
+        /// <summary>
+        /// Whether or not to include self and related links 
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// options.DisableSelfAndRelatedLinks = true;
+        /// </code>
+        /// </example>
+        public bool DisableSelfAndRelatedLinks { get; set; }
```
jaredcnance: I think this should just be the global default [`Link`](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/cd02b20dfee084fbaf9c4283af6102afa52b0e0d/src/JsonApiDotNetCore/Models/Link.cs#L6). Maybe something like:

```c#
public Link DefaultRelationshipLinks { get; set;} = Link.All
```

milosloub: That makes sense with Link.All. I thought about this more and got following: 
a) we can expose new property on JsonApiOptions "DefaultRelationshipLinks" (instead of DisableSelfAndRelatedLinks )
b) we can use present property ResourceGraphBuilder on JsonApiOptions, set it's DocumentLinks and than check this property while building resource in DocumentBuilder.cs.

Second solution might be breaking change, if you are developer who set it like this before. But I can´t imagine why anyone would do this.
What do you think?

milosloub:
For now I did it like options.DefaultRelationshipLinks. I didn't move DocumentLinks. If you want to do it at once - this will lead to change unit test Page_Links_Can_Be_Disabled_Globally.

PR #456, opened 2018-12-06
Fixed HasManyThrough relationships returning empty data array without 'include'

### LINKED ISSUE #453: HasManyThrough relationships always return empty data:[] array on GET
## Description

Seems when trying to GET a record that contains a HasManyThrough relationship, it returns an empty `data: []` array even though I'm not specifying `?include`. Meanwhile, HasMany relationships seem to work just fine.

Here's a somewhat-anonymized sample response (comments added) which hopefully illustrates the issue:
```js
{
    "data": [
        {
            "attributes": {
                "title": "Sample Article",
                "text": "...",
                "create-date": "2018-08-14T16:19:13.653"
            },
            "relationships": {
                "comments": { // HasMany relationship (works correctly)
                    "links": {
                        "self": "/api/articles/1/relationships/comments",
                        "related": "/api/articles/1/comments"
                    }
                },
                "authors": { // HasManyThrough relationship
                    "links": {
                        "self": "/api/articles/1/relationships/authors",
                        "related": "/api/articles/1/authors"
                    },
                    "data": [] // <-- this shouldn't be here
                }
            },
            "type": "articles",
            "id": "1"
        }
    ]
}
```

To clarify, the article has authors, and they're returned successfully when I `?include` them, so this empty array is a bit of a red herring that's monkeying up the client application (`ember-data`).

I did stumble across [this issue](https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/318) and double-checked to make sure I'm not initializing any lists anywhere, and it doesn't seem so. Not sure what else, if anything, might be configured differently in my project versus the repo example, so I'm a bit stumped.

## Environment

- JsonApiDotNetCore Version: 3.0.0
- Other Relevant Package Versions: [?]

XaserAcheron:
Closes #453

#### BUG FIX
- [x] fix issue
- [ ] bump package version

The basic gist here is that there was an iterator function that was returning an empty IEnumerable even if the property value for the relationship was `null`. All fixed now.

Let me know if I should bump the package version or anything. I'm not totally sure what the process is there.

### DIFF
--- a/src/JsonApiDotNetCore/Internal/ResourceGraph.cs
+++ b/src/JsonApiDotNetCore/Internal/ResourceGraph.cs
@@ -144,11 +144,18 @@ namespace JsonApiDotNetCore.Internal
             var throughProperty = GetRelationship(parent, hasManyThrough.InternalThroughName);
             if (throughProperty is IEnumerable hasManyNavigationEntity)
             {
-                foreach (var includedEntity in hasManyNavigationEntity)
-                {
-                    var targetValue = hasManyThrough.RightProperty.GetValue(includedEntity) as IIdentifiable;
-                    yield return targetValue;
-                }
+                // wrap "yield return" in a sub-function so we can correctly return null if the property is null.
+                return GetHasManyThroughIter(hasManyThrough, hasManyNavigationEntity);
+            }
+            return null;
+        }
+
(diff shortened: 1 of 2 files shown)


XaserAcheron:
Glad to help -- thanks for the merge & tag!

PR #457, opened 2018-12-06
Update TypeMatchFilter.cs

jonnolen:
should only be typechecking on json api requests that are patch or post... because of left to right evaluation and lack of parens  this was evaluating as false || is post if the request was not a jsonapi request.


jaredcnance:
Thanks!

Issue #454, opened 2018-12-06
Null valued attributes are not omitted from included data


ddaspit:
## Description
If `DocumentBuilderOptions.OmitNullValuedAttributes` is set to `true` and an included resource object has a null valued attribute, it will not be omitted. Any null valued attributes in primary resource objects are correctly omitted. It looks like `DocumentBuilder.GetIncludedEntity()` does not check `OmitNullValuedAttribute()` method when adding attributes.

## Environment

- JsonApiDotNetCore Version: 3.0
- Other Relevant Package Versions:

PR #458, opened 2018-12-09
Feat(#406): Allow filter[id] and filter[related.id]

### LINKED ISSUE #406: Filtering on Id?
Do you know, how should be handled filtering like this?
api/clients?filter[id]=in:1,2,3

this leads to make all resources derive from "BaseEntity" with Identifiable override:

>         [Attr("id")]
>         public override int Id { get => base.Id; set => base.Id = value; }

it's ok, but then id is presented in top-level and in attributes also. Do you know 
how to handle this as the best practice?

milosloub:
Closes #406

### DIFF
--- a/src/JsonApiDotNetCore/Builders/ContextGraphBuilder.cs
+++ b/src/JsonApiDotNetCore/Builders/ContextGraphBuilder.cs
@@ -148,6 +148,7 @@ namespace JsonApiDotNetCore.Builders
                         InternalAttributeName = prop.Name
                     };
                     attributes.Add(idAttr);
+                    continue;
                 }
 
                 var attribute = (AttrAttribute)prop.GetCustomAttribute(typeof(AttrAttribute));
--- a/src/JsonApiDotNetCore/Builders/ContextGraphBuilder.cs
+++ b/src/JsonApiDotNetCore/Builders/ContextGraphBuilder.cs
@@ -139,6 +139,17 @@ namespace JsonApiDotNetCore.Builders
 
             foreach (var prop in properties)
             {
+                if (prop.Name == nameof(Identifiable.Id))
+                {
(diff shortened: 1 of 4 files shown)


milosloub:
@jaredcnance Concept of this is to register "id" as common attribute, work with this as user need (filter, sort....), but for output is ignored (to not double this property in top level and in attribute level).

My first try was to just expose id, but after that 5 tests were failing -> there are some quite hard dependencies. So I had to hide this and just change two Unit test that asks for attributes from
`graph.GetContextEntity(typeof(TestResource)); `
Id is ignored in DocumentBuilder and those two test don't work with DocumentBuilder, so this is the reason to change Assert.Single to Assert.Contains

Btw - During this implementation I discovered, that DocumentBuilder.GetIncludedEntity() doesn't care about hiding null values as it's described in #454. But I let it be for now.

src/JsonApiDotNetCore/Builders/ContextGraphBuilder.cs
```diff
@@ -139,6 +139,17 @@ protected virtual List<AttrAttribute> GetAttributes(Type entityType)
...
             {
+                if (prop.Name == nameof(Identifiable.Id))
+                {
+                    var idAttr = new AttrAttribute()
+                    {
+                        PublicAttributeName = JsonApiOptions.ResourceNameFormatter.FormatPropertyName(prop),
+                        PropertyInfo = prop,
+                        InternalAttributeName = prop.Name
+                    };
+                    attributes.Add(idAttr);
```
jaredcnance: should there be a `continue` here, after the id property is added?

milosloub: Yes, you're right. Now it's already there.

PR #464, opened 2019-01-23
feat: added extra information to FilterQuery

wisepotato:
I just really like intellisense.  And needed to go through the code to know what it is used for. Hope it helps just a tiny bit. Also maybe my interpretation is wrong but atleast its a start.


jaredcnance:
Thanks!

PR #476, opened 2019-02-20
Fix sparse fields with DefaultEntityRepository.Get() override

### LINKED ISSUE #475: Sparse fields stops working when DefaultEntityRepository.Get() overriden
## Description
Steps to reproduce:
1) I have model like this (reduced for simplicity)
```
public class Client: Identifiable
{
      [Attr]

      [HasOne]
      public Advisor Advisor { get; set; }
      public int AdvisorId { get; set; }
}
```

2) I have registered custom repository with Get() override:
```
        public override IQueryable<Client> Get()
            => base.Get()
                .Where(e => e.AdvisorId == 1);
```
3) I made request: GET /api/clients?fields[clients]=firstName
4) Empty array is returned and console shows:

> warn: Microsoft.EntityFrameworkCore.Query[20500]
>       The LINQ expression 'where (new Client() {Id = [x].Id, FirstName = [x].FirstName}.AdvisorId == 1)' could not be translated and will be evaluated locally.

I can see, that there is strange translation with Where condition (There is statament that belongs to Select query)
I think, that using this line cause the problem
https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/b25ea1bfec80db6dc7f7121c74ef7cb8b39fe1fb/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs#L401

Do you have any idea how to fix this? I can make PR, but I have no idea how to combine those 
Provider.CreateQuery, standard LINQ query builder, selects and where queries...

## Environment

- JsonApiDotNetCore Version: 3.1.0

milosloub:
Closes #475 

Current implementation:
- sparse fields calls Select() query first, than follows Filter() (and other methods in IQueryableExtensions). In case Select() (with internal Provider.CreateQuery<TEntity>() call) is applied like that, LINQ Where() method is reduced in runtime and has no info about other properties outside Select() => this is exactly the use-case, when overriding repository Get() method. And this leads to broken sql translation.

Requested implementation:
- Select() call is moved from DefaultEntityRepository.Get() to EntityResourceService.GetAsync() before Skip() and Take(). There is no "reduce" danger.

What do you think about this change?

### DIFF
--- a/JsonApiDotnetCore.sln
+++ b/JsonApiDotnetCore.sln
@@ -190,7 +190,7 @@ Global
 		{6DFA30D7-1679-4333-9779-6FB678E48EF5}.Release|x64.ActiveCfg = Release|Any CPU
 		{6DFA30D7-1679-4333-9779-6FB678E48EF5}.Release|x64.Build.0 = Release|Any CPU
 		{6DFA30D7-1679-4333-9779-6FB678E48EF5}.Release|x86.ActiveCfg = Release|Any CPU
-		{6DFA30D7-1679-4333-9779-6FB678E48EF5}.Release|x86.Build.0 = Release|Any CPU\
+		{6DFA30D7-1679-4333-9779-6FB678E48EF5}.Release|x86.Build.0 = Release|Any CPU
 		{09C0C8D8-B721-4955-8889-55CB149C3B5C}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 		{09C0C8D8-B721-4955-8889-55CB149C3B5C}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{09C0C8D8-B721-4955-8889-55CB149C3B5C}.Debug|x64.ActiveCfg = Debug|Any CPU
@@ -203,6 +203,7 @@ Global
 		{09C0C8D8-B721-4955-8889-55CB149C3B5C}.Release|x64.Build.0 = Release|Any CPU
 		{09C0C8D8-B721-4955-8889-55CB149C3B5C}.Release|x86.ActiveCfg = Release|Any CPU
 		{09C0C8D8-B721-4955-8889-55CB149C3B5C}.Release|x86.Build.0 = Release|Any CPU
+		{DF9BFD82-D937-4907-B0B4-64670417115F}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
--- a/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
+++ b/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
@@ -79,12 +79,15 @@ namespace JsonApiDotNetCore.Data
         }
 
         /// <inheritdoc />
-        public virtual IQueryable<TEntity> Get()
-        {
-            if (_jsonApiContext.QuerySet?.Fields != null && _jsonApiContext.QuerySet.Fields.Count > 0)
-                return _dbSet.Select(_jsonApiContext.QuerySet?.Fields);
+        public virtual IQueryable<TEntity> Get() 
+            => _dbSet;
 
-            return _dbSet;
+        public virtual IQueryable<TEntity> Select(IQueryable<TEntity> entities, List<string> fields)
+        {
+            if (fields?.Count > 0)
+                return entities.Select(fields);
+
+            return entities;
         }
 
         /// <inheritdoc />
--- a/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
+++ b/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
@@ -78,8 +78,18 @@ namespace JsonApiDotNetCore.Data
             _resourceDefinition = resourceDefinition;
         }
 
+
+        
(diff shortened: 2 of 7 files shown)


jaredcnance:
Thanks for the PR as always. This is a pretty big breaking change. I need to think about it a little more. We should be able to deprecate the existing Get API without causing a breaking change right away. But, I’d like to spend some more time understanding the problem. I know that we override this method in many places and haven’t run into this issue. I’m wondering if there is an issue here that needs to be pushed back onto EF? But anyways, I need to work on a minimum reproduction of the issue so I can understand it better.

milosloub:
@jaredcnance To understand the problem, you can simply try this two use-cases in TodoItemRepository with basic query [GET] api/todo-items
a)
```
        public override IQueryable<TodoItem> Get()
            => base.Get()
                .Select(i => new TodoItem() { Description = i.Description }) // this line comes from IQueryableExtensions.Select() in runtime
                .Where(e => e.OwnerId == 1);
```
b)
```
        public override IQueryable<TodoItem> Get()
            => base.Get()
                .Where(e => e.OwnerId == 1)
                .Select(i => new TodoItem() { Description = i.Description });
```
Only order of Select() and Where() is changed.
In a) example you will notice this line in console:

> The LINQ expression 'where (new TodoItem() {Description = [i].Description }.OwnerId == 1)' could not be translated and will be evaluated locally.

In b) everything is OK

I think this is not EF Core issue, because every LINQ query (method or query syntax) should end with select clause. 

And if you modify example a) and add OwnerId to Select, than everything works as expected.
```
        public override IQueryable<TodoItem> Get()
            => base.Get()
                // OwnerId provided extra, than it is working
                .Select(i => new TodoItem() { Description = i.Description, OwnerId = i.OwnerId })
                .Where(e => e.OwnerId == 1);
```
That was the reason I was talking about "reducing" properties -> Where() doesn't see property, which is not provided in previous Select()

Is it more clear, now?

milosloub:
@jaredcnance hey man, or you can try simply this (without any Repository Layer and overrides)

[GET] api/todo-items?filter[ordinal]=1&fields[todo-items]=description
-> you will get empty array
[GET] api/todo-items?filter[ordinal]=1&fields[todo-items]=ordinal
-> you will get some values

The problem is same - applying filters and it's Where() method is called after Select()

jaredcnance:
Yes, I see. You're correct. And I don't think moving the order alone should be a breaking change. I think we can probably use C# 8 default interface methods to keep this from being a breaking change entirely. Later this week I may put together a draft document that describes our approach and philosophy around these kinds of breaking changes so that users have clear expectations.

But, other than that, this approach looks good to me. Thanks!

milosloub:
@jaredcnance What about new GetQueryable() method like you mentioned here https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/404#issuecomment-420483020 ?
Then deprecate Get() (in DefaultEntityRepository and IEntityReadRepository), then EntityResourceService can call this new method and add that Select() line. 

This is breaking change only for those, who implements custom ResourceService by override or implement IEntityReadRepository. They will see this deprecation with clear description. If developer let deprecation be, function is the same. If not he has to change to GetQueryable() and provide Select() row to his custom service. 
What do you think?

jaredcnance:
No, I like the idea. However, for the first rollout I think I would prefer to try using default interface methods and deprecation warnings. This lets us release it as a minor version first so users don’t have to fix all overrides in one go (some projects make heavy use of overrides and the work will take a while). But, overall, I like the proposal. 👍

milosloub:
Ok, no problem. But can you give me some explained guidance, where to put changes? Because I'm not familiar with default interface methods, yet. Right now I can't imagine the right approach.

jaredcnance:
Sure. Tomorrow evening I’ll extend this PR with what I’m thinking if that’s alright

milosloub:
Ok, thanks.

jaredcnance:
Ah never mind, I was think that default interface methods were already available in the 3.0.0-preview SDK. But, apparently it's still at the [proof-of-concept phase](https://github.com/dotnet/csharplang/blob/master/proposals/default-interface-methods.md). So, let's bump the major version and we can release this as alpha.

milosloub:
Yeah I see, never mind. Thanks for approval, but do you want to do it like this "first pass" way? What about deprecating Get() and introduce new GetQueryable() to minimize posible breaking changes? Just asking for sure

milosloub:
@jaredcnance Can you look at the question above? This is quite blocking feature for my company at this moment and I don't know if I can realease as alpha...

maurei:
> What about deprecating Get() and introduce new GetQueryable() to minimize possible breaking changes? 

I think we should do this. We'll completely drop `Get()` in future releases.

maurei:
Indeed, @milosloub, there are still some issues; I was too hasty with replacing `Get` with the new `GetQueryable`.

Interestingly enough tests are not failing. Some investigation about this:

* `Fields_Query_Selects_Sparse_Field_Sets` is not failing because, even though the sparse field set is not processed in the SQL query, the `DocumentBuilder` [still removes the fields](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/11fe0dac79468af58f55c7c56bab013e1fdeb8f0/src/JsonApiDotNetCore/Builders/DocumentBuilder.cs#L134) from the output JSON. This behaviour annoys me: (de)serialization should be decoupled from `JsonApiContext`, see #497.

* I suspect the same goes for the other sparse field set tests.

PR #487, opened 2019-03-24
explain ModelState validation option ValidateModelState

CoolGoose:
This should make it clearer for people on how to have nice validation when creating / updating requests.

### DIFF
--- a/docs/usage/options.md
+++ b/docs/usage/options.md
@@ -61,7 +61,7 @@ services.AddJsonApi<AppDbContext>(options => {
 
 ## Custom Query Parameters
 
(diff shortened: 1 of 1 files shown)


docs/usage/options.md
```diff
@@ -78,3 +78,13 @@ If you want to change the default serializer settings, you can:
 options.SerializerSettings.NullValueHandling = NullValueHandling.Ignore;
 options.SerializerSettings.ContractResolver = new DasherizedResolver();
 ```
+
+## Enable ModelState Validation
+
+If you would like to use ModelState validation into your controllers when creating / updating resources you set `ValidateModelState = true`. The default behavior is to try creating the database entries.
```
jaredcnance: Thanks for the PR! This statement is a bit unclear to me 

> The default behavior is to try creating the database entries.

I would probably change this to something along the lines of:

> By default, no model validation is performed.

CoolGoose: Will update, thank you for the comment.

Issue #497, opened 2019-04-24
Deserializer only links attributes and not included relationships


amit-ch:
For a simple model like below

```
    public class Customer : Identifiable
    {
        [HasMany] public List<Order> Orders { get; set; }
    }

    public class Order : Identifiable
    {
        [HasMany] public List<Invoice> Invoices { get; set; }
    }

    public class Invoice : Identifiable { }
```
Serializer works fine and get customer with included relationship returns the desired response
/api/v1/customers?included=orders.invoices

```
{
    "data": {
        "attributes": {},
        "relationships": {
            "orders": {
                "links": {
                    "self": "https://localhost:44351/customers/1234/relationships/orders",
                    "related": "https://localhost:44351/customers/1234/orders"
                },
                "data": [
                    {
                        "type": "orders",
                        "id": "4567"
                    }
                ]
            }
        },
        "type": "customers",
        "id": "1234"
    },
    "included": [
        {
            "attributes": {},
            "relationships": {
                "invoices": {
                    "links": {
                        "self": "https://localhost:44351/orders/4567/relationships/invoices",
                        "related": "https://localhost:44351/orders/4567/invoices"
                    },
                    "data": [
                        {
                            "type": "invoices",
                            "id": "7890"
                        }
                    ]
                }
            },
            "type": "orders",
            "id": "4567"
        },
        {
            "attributes": {},
            "type": "invoices",
            "id": "7890"
        }
    ]
}
```
However on using the same json as input for post/put operation and trying to deserialize into customer instance with all fields/relationships populated,  it does not load included relationship.
In this instance, customer has orders attributes populated but invoices is appearing blank.

![image](https://user-images.githubusercontent.com/28331812/56700812-07c64b00-66f4-11e9-8dee-71e040ebb33d.png)

Is it because we are only setting attributes and not relationship. Can we please modify it to add second level or more relationship be included as well.
https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/9059560ffdaa9cc263fd352bb7dc1106255d6148/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs#L323

wisepotato:
Can you link to json:api spec where it says this should be supported? Seems like it should.

maurei:
I don't think it's a part of the JSON:API spec to support create/update with nested relationships. 

[...]

PR #498, opened 2019-04-25
fix: sparse fields set not applied to query

maurei:
continuation of #476 

Indeed, @milosloub, there are still some issues; I was too hasty with replacing `Get` with the new `GetQueryable`.

Interestingly enough tests are not failing. Some investigation about this:

* `Fields_Query_Selects_Sparse_Field_Sets` is not failing because, even though the sparse field set is not processed in the SQL query, the `DocumentBuilder` [still removes the fields](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/11fe0dac79468af58f55c7c56bab013e1fdeb8f0/src/JsonApiDotNetCore/Builders/DocumentBuilder.cs#L134) from the output JSON. This behaviour annoys me: (de)serialization should be decoupled from `JsonApiContext`, see #497. I don't think we can properly integration test the usage of the new `GetQueryable` without getting rid of this behaviour in the serialization layer, but that is beyond the scope of this PR.

* I suspect the same goes for the other sparse field set tests. 

I have changed a few things that should apply it to the queries again. All tests are passing, let me know if you have any more comments or an idea how we could more thoroughly test the new `GetQueryable`. (Note that the previous `Get` wasn't properly tested for sparse field selection neither)

### DIFF
--- a/JsonApiDotnetCore.sln
+++ b/JsonApiDotnetCore.sln
@@ -204,6 +204,7 @@ Global
 		{09C0C8D8-B721-4955-8889-55CB149C3B5C}.Release|x86.ActiveCfg = Release|Any CPU
 		{09C0C8D8-B721-4955-8889-55CB149C3B5C}.Release|x86.Build.0 = Release|Any CPU
 		{DF9BFD82-D937-4907-B0B4-64670417115F}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{DF9BFD82-D937-4907-B0B4-64670417115F}.Debug|Any CPU.Build.0 = Debug|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
--- a/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
+++ b/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
@@ -140,7 +140,7 @@ namespace JsonApiDotNetCore.Data
         /// <inheritdoc />
         public virtual async Task<TEntity> GetAsync(TId id)
         {
-            return await GetQueryable().SingleOrDefaultAsync(e => e.Id.Equals(id));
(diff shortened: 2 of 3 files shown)


milosloub:
Yeah, you are right. I totally forgot about that logic in DocumentBuilder. I remember, that there is a problem with this - for example you have integer property, so default value is 0 and you are not able to distinguish if it is standard value stored in DB, or if it is default value while new instance is created and you are not requesting this property. Right now I don't have idea how to solve this other than this approach :/

maurei:
> ... so default value is 0 and you are not able to distinguish if it is standard value stored in DB, ...

You're right, I don't think it's possible like this. Even if we only supported sparse fields selection for nullable fields, eg nullable int, to distinguish between `0` and `default(int)`, it still wouldn't cover every scenario because there might be cases in which users actually want to use `null` as meaningful data. The same goes for strings and most/all other datatypes.

So yep, The (de)serialization, when sparse field selection is involved, can not _completely_ (although still more than it is now) be decouple from `JsonApiContext` .

With respect to this I have the following thoughts:

* We can still better organise the deserializers so that they become more reusable. Eg: if the deserializer would be used in a [JsonApiClient](https://github.com/maurei/JsonApiDotNetCoreClient), it will never have to worry about sparse fields. So I think that `DocumentBuilder` still should naively convert entities into documents without relying on `JsonApiContext`, and the actual `IJsonApiDeserializer` can perform additional logic, like 
      * taking into account sparse fields when the serializer is used in a server (`public class IJsonApiServerDeserializer` : IJsonApiDeserializer`)
      * or removing metadata / links when used in a client (`public class IJsonApiClientDeserializer` : IJsonApiDeserializer`)

* Sparse field selection on the level of the repositories should be tested using integration tests that involve only repositories, using an [in-memory database](https://spin.atomicobject.com/2018/11/27/dot-net-core-testing/). The current `Fields_Query_Selects_Sparse_Field_Sets` test is actually an e2e test.



Anyway, this is both ofc beyond the scope of this PR

PR #499, opened 2019-04-25
Feat/#482

### LINKED ISSUE #482: Increase ResourceDefinition GetFilterQuery Flexibility
I really like the new ResourceDefinition interface, and it has allowed me to really separate and clean out some of the manual filtering I was doing traditionally in my EntityRepositories. That said, I do have a hitch with the following line:
https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/e58a48dc25248a042bf3f3142186736faa2e908a/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs#L98

Is there are particular reason why the delegate takes only the ```filterQuery.Value``` and not the whole FilterQuery object (or at least value and operation)? 

While the new ResourceDefintion has cleared up the basic filtering logic I have, I still have to keep messy logic for specific ```FilterQuery.Operation```'s. 

```
public class FooRepository : DefaultEntityRepository<Foo>
...
public override IQueryable<Foo> Filter(IQueryable<Foo> foos, FilterQuery filterQuery)
{
  if(filterQuery.Operation == "in" && filterQuery.Attribute == "bar")
    return someInFilter(filterQuery.Value);
}
...
```

The new QueryFilter alias would then be something like:
```
public class QueryFilters : Dictionary<string, Func<IQueryable<T>, string, string, IQueryable<T>>> { }
```
or
```
public class QueryFilters : Dictionary<string, Func<IQueryable<T>, FilterQuery, IQueryable<T>>> { }
```

Such that in the ```ResourceDefinition``` ```GetQueryFilters``` you can also know the filter operation being used on the values provided.

I would be glad to get thoughts on this and eventually open a PR for this.

joshhubers:
Closes #482 

#### FEATURE
- [ ] bump package version

### DIFF
--- a/docs/usage/resources/resource-definitions.md
+++ b/docs/usage/resources/resource-definitions.md
@@ -113,9 +113,9 @@ public class ItemResource : ResourceDefinition<Item>
     // handles queries like: ?filter[was-active-on]=2018-10-15T01:25:52Z
     public override QueryFilters GetQueryFilters()
         => new QueryFilters {            
-            { "was-active-on", (items, value) => DateTime.TryParse(value, out dateValue)
+            { "was-active-on", (items, filter) => DateTime.TryParse(filter.Value, out dateValue)
                 ? items.Where(i => i.Expired == null || dateValue < i.Expired)
-                : throw new JsonApiException(400, $"'{value}' is not a valid date.")
+                : throw new JsonApiException(400, $"'{filter.Value}' is not a valid date.")
             }
         };
 }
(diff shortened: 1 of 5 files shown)


maurei:
Looks good! I think if we want we could avoid this to be a breaking change, but we're bumping the major version anyway so yeah

joshhubers:
That's a good point. 

I haven't done much external versioning management before 🤷‍♂️, but I'll definitely keep that in mind with contributions down the road.

maurei:
An overload with an obsolete message would do the trick yep. If you feel like adding that in another PR would be greatly appreciated. Also feel free to PM us on gitter if you have any questions or feel like getting involved more!

Issue #502, opened 2019-04-30
Refactor AttachRelationships: prevent reattachment, support implicit remove


maurei:
### Problem
Consider the following scenario 
* `Person` one-to-one `Passport`
* initial condition of the database:  
     * two `Person`s items: `person1, person2`.
     * only one `Passport` item: `passport`
     * foreign key `PassportId` on `person1` equals id of `passport`

We now do a `PATCH /people/2` and attempt to update the relationship of  `person2` by relating it to `passport`. What I would expect it to do is to perform an **implicit remove** of the old relationship: implicit in the sense that the relationship with `person1` is updated even though it was never explicitly mentioned in the original request.

Right now, it results in a 500 because a foreign key constraint is violated. This is because `person1` is already related to `passport` and the implicit remove is not taken care of, resulting in a FK constraint error.

[Update bug reproduced in this test](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/926a94594559e0963b313be18d39e6c7bcf0c859/test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/UpdatingRelationshipsTests.cs#L626)

Issue: a 500 is not acceptable.

### Solution

Two options
1) Don't actually do the implicit remove, but return a `400 Bad Request` with a body stating that the relation between `person1` and `passport` should be removed first, thus more or less reflecting the foreign key constraint error message that is generated by the database~
2) handle these situations intelligently in the repository layer. The repository layer would take care of breaking the relationship between `person1` and `passport`  before relating `person2` to `passport`. EF Core can handle this trivially if we just attach `person1` to the `dbContext` before saving the changes (by just loading the Person relationship of passport). Right now that doesn't happen.

I prefer the second option.

**[EDIT]**

This goes wrong for `CREATE`, `UPDATE` for both one-to-one and one-to-many. All goes in a similar way: the entity on the inverse navigation property needs to be attached. This means that the fix is not specific to one particular "pipeline", but the `AttachRelationship` method which is called in all involved pipelines needs an upgrade:
* for `many-to-many` and `one-to-many`: needs to attach relationships of the to-be-manipulated entity so that complete replacement actually works (see #492  and #494)
* for `one-to-one` and `one-to-many`: It needs to do attachment of inverse relationships to prevent the error as mentioned above from occurring. See #503: it has a proof of concept for the specific example above
[...]

PR #503, opened 2019-04-30
Support implicit remove when creating/updating one-to-one relationship

maurei:
The scope of the PR has just significantly changed, see #502. Closing this as long as it is outdated

### DIFF
--- a/src/JsonApiDotNetCore/Extensions/DbContextExtensions.cs
+++ b/src/JsonApiDotNetCore/Extensions/DbContextExtensions.cs
@@ -14,6 +14,30 @@ namespace JsonApiDotNetCore.Extensions
         public static DbSet<T> GetDbSet<T>(this DbContext context) where T : class 
             => context.Set<T>();
 
+        public static void LoadInverseNavigation<TParent>(
+            this DbContext context,
+            RelationshipAttribute parentToChildAttribute,
+            object childEntity) where TParent : class, IIdentifiable
+        {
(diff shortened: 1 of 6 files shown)


wisepotato:
in the future please add the solution you are proposing in the PR request, instead of just referencing the issue, this gives the reviewer a bette runderstanding of what you did to achieve, and what solution you used (out of the 2 proposed)

src/JsonApiDotNetCore/Extensions/DbContextExtensions.cs
```diff
@@ -14,6 +14,30 @@ public static class DbContextExtensions
         public static DbSet<T> GetDbSet<T>(this DbContext context) where T : class 
             => context.Set<T>();
 
+        public static void LoadInverseNavigation<TParent>(
```
wisepotato: PPlease comment

src/JsonApiDotNetCore/Extensions/DbContextExtensions.cs
```diff
@@ -14,6 +14,30 @@ public static class DbContextExtensions
...
+            this DbContext context,
+            RelationshipAttribute parentToChildAttribute,
+            object childEntity) where TParent : class, IIdentifiable
```
wisepotato: Could this be `IIdentifiable` or an underlying interface?  Why not `IIdentifiable`?

PR #505, opened 2019-05-07
Create test for accessing many to many data through id.  ie articles/…

sarahentzel:
…1/tags

#Issue {491}

- [ ] reproduce issue in tests


wisepotato:
We will be working on this after release of v4. We will check if it is still reproducible.

maurei:
Closing as outdated and unmaintained. This does not mean this is not a valid contribution, but it is incomplete. An updated version of this PR is welcome

PR #508, opened 2019-05-16
fix: fixed isimmutable not working

### LINKED ISSUE #507: IsImmutable weird behaviour
Lets say I have the following model
```c#
using Tbdc.Apis.Mgmt.Authorization;
using Tbdc.Apis.Mgmt.Models.Authorization;
using JsonApiDotNetCore.Models;
using System;
using System.Collections.Generic;
using System.ComponentModel.DataAnnotations.Schema;
using Bdc.Api.Management.Models;
using Bdc.Api.Management.Models.Contracts;
using Bdc.Api.Management.Models.Abstract;

namespace Tbdc.Apis.Mgmt.Models
{
    /// <summary>
    /// The user model
    /// </summary>
    public class User :  Identifiable, ITraversable
    {
        [HasMany("cachedReports", canInclude: false), NotMapped]
        public List<Report> CachedReports { get; set; }
        [Attr("firstName", isImmutable: true)]

        [Attr("tussenvoegsel", isImmutable: true)]
        public string Tussenvoegsel { get; set; }

        [Attr("fullName", isImmutable: true)]
        public string FullName
        {
            get => this.FirstName + (this.Tussenvoegsel != null ? " " + this.Tussenvoegsel : "") + " " + this.LastName;
        }
        [Attr("lastName", isImmutable: true)]
    }
}

```
If I deserialize something coming in that has `fullName` I dont want it to be set.

## Environment

- JsonApiDotNetCore Version: 

wisepotato:
Closes #507

### DIFF
--- a/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs
+++ b/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs
@@ -146,11 +146,11 @@ namespace JsonApiDotNetCore.Serialization
             {
                 if (attributeValues.TryGetValue(attr.PublicAttributeName, out object newValue))
                 {
+                    if (attr.IsImmutable)
+                        continue;
                     var convertedValue = ConvertAttrValue(newValue, attr.PropertyInfo.PropertyType);
                     attr.SetValue(entity, convertedValue);
-
-                    if (attr.IsImmutable == false)
-                        _jsonApiContext.AttributesToUpdate[attr] = convertedValue;
+                    _jsonApiContext.AttributesToUpdate[attr] = convertedValue;
                 }
             }
 
--- a/JsonApiDotnetCore.sln
+++ b/JsonApiDotnetCore.sln
@@ -1,6 +1,7 @@
-﻿Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2010
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28606.126
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "JsonApiDotNetCore", "src\JsonApiDotNetCore\JsonApiDotNetCore.csproj", "{C0EC9E70-EB2E-436F-9D94-FA16FA774123}"
 EndProject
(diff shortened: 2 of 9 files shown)


src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs
```diff
@@ -146,11 +146,11 @@ public object DocumentToObject(ResourceObject data, List<ResourceObject> include
             {
                 if (attributeValues.TryGetValue(attr.PublicAttributeName, out object newValue))
                 {
+                    if (attr.IsImmutable)
```
milosloub: With this fix, I think you cannot set value also in POST. Can you check this? 
Also test for this would be great. There is only Immutable_Attrs_Are_Not_Included_In_AttributesToUpdate
but only as Unit, Acceptance would be much better for this

wisepotato: I'll check POST, added check for PATCH which gives a 422, which is to be expected, but would like a more thorough response,

wisepotato: didnt add the thorough response because it would require a slight refactor. Will take that into account when refactoring `JsonApiContext`

milosloub: So we should ignore this for now?

wisepotato: I'll put in another test for POST, I would assume it to not work when posting, as you are requesting something that shouldn't work.

milosloub: But DocumentToObject is called also while POST, so user must be able to create this value on first pass. This is standard use-case and this eliminate the posibility to create value for the first time at all. 
For example - you create user with male/female property and you don't want to allow change this in later operations.

wisepotato: fair enough, so should we consider a new constructor parameter for `Attr`? I think `isImmutable` can be considered a good approach for disapproving patch operations on a certain attribute, but what about when we don't want it to be editable at all? `isSettable` ?

wisepotato: where `isSettable` also sets `isImmutable` by proxy of course.

milosloub: Property is readonly when has only get accessor. I think there is no need to introduce new parameter. If you have fullname attribut with only get accessor and client send POST/PATCH it will throw exception -> it controls bad approach while someone implementing you api. Or do I miss something?

wisepotato: we could catch the specific Exception and pass it on as information to the enduser... seems good to me

milosloub: yes, that exception meesage in 422 would be ok, I think

PR #510, opened 2019-05-17
Fix/#509

### LINKED ISSUE #509: DateTimeOffset is modified during deserialization
## Description

The problem is on this line
https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/0416df760719a41612fda20bae21e3f7d6707855/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs#L39

When using DateTimeOffset type and using JToken.Parse, it is converted to DateTime (so it is modified) as it is hardcoded as a standard behavior, see:
https://github.com/JamesNK/Newtonsoft.Json/issues/862
(it is closed, but still unresolved for JToken.Parse)

Solution is change JToken.Parse to JToken.Load with StringReader and set option DateParseHandling = DateParseHandling.None; (default is DateTime). It is described here:
https://github.com/JamesNK/Newtonsoft.Json/issues/862#issuecomment-222406929

This issue solved global datetimeoffset destroying data, beacuse there is no option to pass another settings to JToken.Parser to override this default.

## Environment

- JsonApiDotNetCore Version 3.1

milosloub:
Closes #509

### DIFF
--- a/src/Examples/JsonApiDotNetCoreExample/Models/TodoItem.cs
+++ b/src/Examples/JsonApiDotNetCoreExample/Models/TodoItem.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using JsonApiDotNetCore.Models;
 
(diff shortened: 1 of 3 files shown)


wisepotato:
you should add the email you are using to commit so it links to your profile :)

PR #511, opened 2019-05-22
V4.0

milosloub:
new package version


maurei:
I agree on what you said about alpha versions. Should we do v4.0.0-alpha.1 for this release?  Then, later this week / next week we could do an alpha.2 with hooks included.  Also @wisepotato is working on decoupling `JsonApiContext` and is expecting to have first releasable results next week, which we could put in an alpha.3. Then maybe by the end of next week we could make a RC and release it after it has survived some time

PR #512, opened 2019-05-23
[draft] Feat/serializer context decoupling

wisepotato:
The main goal of this PR is to counteract the coupling observed in JADNC. Let's make it more testable!

- [x] No more `JsonApiContext` dependency in Controllers
- [x] The entity is set via `ResourceGraph` in the middleware now
- [x] no more `ApplyContext<T>` in the controller, making for a coupling that is not needed
- [x] get all tests working

#### BUG FIX
- [x] reproduce issue in tests
- [ ] fix issue
- [ ] bump package version

#### FEATURE
- [ ] write tests that address the requirements outlined in the issue
- [ ] fulfill the feature requirements
- [ ] bump package version

Issue #494, opened 2019-04-11
Update many-to-many through PATCH on resource not working


maurei:
Exactly same part of the spec as #492  but for `many-to-many` instead of `one-to-many`. 

Currently creating relations in the simplest scenario works, as clarified by [this test](https://github.com/wisepotato/JsonApiDotNetCore/blob/0158da3e80cf0c82b846f13b52d5133f0ce3c50b/test/JsonApiDotNetCoreExampleTests/Acceptance/ManyToManyTests.cs#L205)

More complex scenarios will fail. Eg, consider and article with tag_1
* sending a patch with a new relationship to tag_2 (not including tag_1) results in having both. No complete replacement was performed
* Similarly, when including both tag_1 and tag_2, it will try to recreate a relationship to tag_1, which will break

Issue #492, opened 2019-04-11
Update one-to-many through PATCH on resource not working


maurei:
A [core feature of the spec](https://jsonapi.org/format/#crud-updating-resource-relationships) is currently not supported:
```json
PATCH /articles/1 HTTP/1.1
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "data": {
    "type": "authors",
    "id": "1",
    "relationships": {
      "books": {
        "data": [
          { "type": "books", "id": "2" },
          { "type": "books", "id": "3" }
        ]
      }
    }
  }
}
```

**Cause of the problem**
Because of incomplete information on `_jsonApiContext` in the case of `one-to-many`, which is set in the `JsonApiDeserializer`, the `DefaultEntityRepository` does not attempt to update such relationships. Details:

*  In the case of `one-to-one`, the `JsonApiDeserializer` adds meta info about this to `_jsonApiContext.RelationshipsToUpdate` [right here in the SetHasOneRelationship method](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/8f685a6a23515acf8f440c70d20444a0cec1c502/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs#L214).
* This property however is not updated in the case of `one-to-many`: [in the SetHasManyRelationship method](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/8f685a6a23515acf8f440c70d20444a0cec1c502/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs#L299), only `_jsonApiContext.HasManyRelationshipPointers` are updated, which is not enough.
*  [this is where the repository](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/8f685a6a23515acf8f440c70d20444a0cec1c502/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs#L299) would be updating `one-to-many` relationships if they were included in `_jsonApiContext.RelationshipsToUpdate`.
* Also in the repo: `AttachRelationships(oldEntity)` is fired **before** `relationship.Key.SetValue(oldEntity, relationship.Value);` is called, so any updates to the relations are currently not tracked.

A complication: the request above should perform a **complete replace**. This means that before updating the `Author.BookId` to its new value, the foreign keys of all previous `Book`s for that author should be set to null (and an error should be thrown it is non-nullable).

**Solution**
[...]

PR #478, opened 2019-02-20
Hooks 

wisepotato:
[see comment below on usage guide](https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/478#issuecomment-486258385)

Also, removed some outdated comments/discussion to clear the clutter.

PR #518, opened 2019-06-07
DefaultEntityRepository cleanup

### LINKED ISSUE #502: Refactor AttachRelationships: prevent reattachment, support implicit remove

### LINKED ISSUE #520: Bad PATCH request (without ID) does not return HTTP 400
There are currently several PATCH related tests (for example [Can_Patch_CamelCasedModels](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/8b630c471d7973fa6621ed6f978bc866a6d95a7c/test/JsonApiDotNetCoreExampleTests/Acceptance/CamelCasedModelsControllerTests.cs#L133)) that are passing but shouldn't. When [according to the specs](https://jsonapi.org/format/#crud-updating), the id attribute must be included in the request body

> The PATCH request MUST include a single resource object as primary data. The resource object MUST contain type and id members.

Therefore any PATCH request without ID should return a http 400.

maurei:
This PR fixes of several issues and a bunch of refactoring.

* fixes an issue with EF Core exceptions thrown when `DefaultEntityRepository` tries to attach entities to the `DbContext` that are already attached. This may happened when entities involved to a update or create request were already loaded in the dbcontext elsewhere (eg Resource Hooks). Closes #502. This fix is needed for the tests to pass in #478  
     * added consistent handling for tracked entities
* support for implicit removal of relationships. For this the `IInverseRelationship` service is introduced (actually it was cherry picked from the Resource Hooks PR).  See #502.
* Refactored the `AttachRelationships` method and those associated to it such that
     * it doesn't also *assign* relationshipvalues in (only) the case of many-to-many
     * it no longer depends on `IJsonApiContext` for the *values* of updated relationships and attributes. Still uses it to identify *which* attributes/relationships to update. See #519
* Marked `UpdateAsync(TId id, TEnity entity)` as obsolete, added a new one: `UpdateAsync(TEnity updatedEntity)` which actually uses the updated entity in the repository layer, as opposed to it just being passed along and being left untouched in the previous implementation.  See #519 
* Refactored the fix that was included earlier to support complete replace for PATCH requests (see #492 and #494).
* Refactored the usage of the following properties of `IJsonApiContex` `RelationshipsToUpdate`, `HasOnePointers`, `HasManyPointers`. The first one has become an alias for the latter two, and the latter two are no longer populated with the actual relationship  values as these are not used anymore in the repository layer. This is related to #512.
* No longer allowing PATCH requests that do not contain ID values in the request body. Fixes #520

### DIFF
--- a/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
+++ b/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
@@ -149,6 +149,15 @@ namespace JsonApiDotNetCore.Data
         public virtual async Task<TEntity> CreateAsync(TEntity entity)
         {
             AttachRelationships(entity);
+            foreach (var relationshipEntry in _jsonApiContext.RelationshipsToUpdate)
+            {
+                var relationshipValue = relationshipEntry.Value;
+                if (relationshipEntry.Key is HasManyThroughAttribute throughAttribute)
+                {
+                    AssignHasManyThrough(entity, throughAttribute, (IList)relationshipValue);
+                }
+            }
+
             _dbSet.Add(entity);
 
             await _context.SaveChangesAsync();
--- a/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
+++ b/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
@@ -144,23 +144,29 @@ namespace JsonApiDotNetCore.Data
         /// <inheritdoc />
         public virtual async Task<TEntity> CreateAsync(TEntity entity)
         {
-            AttachRelationships(entity);
-            AssignRelationshipValues(entity);
+            foreach (var relationshipAttr in _jsonApiContext.RelationshipsToUpdate?.Keys)
+            {
+                var trackedRelationshipValue = GetTrackedRelationshipValue(relationshipAttr, entity, out bool wasAlreadyTracked);
+                // LoadInverseRelationships(trackedRelationshipValue, relationshipAttribute)
+                if (wasAlreadyTracked)
+                {
+                    /// We only need to reassign the relationship value to the to-be-added
+                    /// entity when we're using a different instance (because this different one
+                    /// was already tracked) than the one assigned to the to-be-created entity.
+                    AssignRelationshipValue(entity, trackedRelationshipValue, relationshipAttr);
+                } else if (relationshipAttr is HasManyThroughAttribute throughAttr)
+                {
+                    /// even if we don't have to reassign anything because of already tracked 
+                    /// entities, we still need to assign the "through" entities in the case of many-to-many.
+                    AssignHasManyThrough(entity, throughAttr, (IList)trackedRelationshipValue);
+                }
+            }
             _dbSet.Add(entity);
-
             await _context.SaveChangesAsync();
 
             return entity;
         }
 
-        /// <summary>
-
-        /// </summary>
-        protected virtual void AttachRelationships(TEntity entity = null)
-        {
-            AttachHasManyAndHasManyThroughPointers(entity);
-            AttachHasOnePointers(entity);
-        }
 
         /// <inheritdoc />
         public void DetachRelationshipPointers(TEntity entity)
@@ -207,7 +213,7 @@ namespace JsonApiDotNetCore.Data
         }
 
         /// <inheritdoc />
-        public virtual async Task<TEntity> UpdateAsync(TId id, TEntity entity)
+        public virtual async Task<TEntity> UpdateAsync(TId id, TEntity updatedEntity)
         {
             /// WHY is parameter "entity" even passed along to this method??
             /// It does nothing!
@@ -217,25 +223,73 @@ namespace JsonApiDotNetCore.Data
             if (oldEntity == null)
                 return null;
 
-            foreach (var attr in _jsonApiContext.AttributesToUpdate)
-                attr.Key.SetValue(oldEntity, attr.Value);
+            foreach (var attr in _jsonApiContext.AttributesToUpdate.Keys)
+                attr.SetValue(oldEntity, attr.GetValue(updatedEntity));
 
-            if (_jsonApiContext.RelationshipsToUpdate.Any())
+            foreach (var relationshipAttr in _jsonApiContext.RelationshipsToUpdate?.Keys)
             {
-                /// First attach all targeted relationships to the dbcontext.
-                /// This takes into account that some of these entities are 
-                /// already attached in the dbcontext
-                AttachRelationships(oldEntity);
-                /// load the current state of the relationship to support complete-replacement
-                LoadCurrentRelationships(oldEntity);
-                /// assign the actual relationship values.
-                AssignRelationshipValues(oldEntity);
+                LoadCurrentRelationships(oldEntity, relationshipAttr);
+                var trackedRelationshipValue = GetTrackedRelationshipValue(relationshipAttr, updatedEntity, out bool wasAlreadyTracked);
+                // LoadInverseRelationships(trackedRelationshipValue, relationshipAttribute)
+                AssignRelationshipValue(oldEntity, trackedRelationshipValue, relationshipAttr);
             }
             await _context.SaveChangesAsync();
             return oldEntity;
         }
-        /// <inheritdoc />
 
+
+        /// <summary>
+        /// Responsible for getting the relationship value for a given relationship 
+        /// attribute of a given entity. It ensures that the relationship value 
+        /// that it returns is attached to the database without reattaching duplicates instances 
+        /// to the change tracker.
+        /// </summary>
+        private object GetTrackedRelationshipValue(RelationshipAttribute relationshipAttr, TEntity entity, out bool wasAlreadyAttached)
+        {
+            wasAlreadyAttached = false;
+            if (relationshipAttr is HasOneAttribute hasOneAttribute)
+            {
+                /// This adds support for resource-entity separation in the case of one-to-one. 
+                var relationshipValue = GetEntityResourceSeparationValue(entity, hasOneAttribute) ?? (IIdentifiable)hasOneAttribute.GetValue(entity);
+                if (relationshipValue == null) 
+                        return null;
+                return GetTrackedHasOneRelationshipValue(relationshipValue, hasOneAttribute, ref wasAlreadyAttached);
+            }
+            else
+            {
+                IEnumerable<IIdentifiable> relationshipValueList = (IEnumerable<IIdentifiable>)relationshipAttr.GetValue(entity);
+                /// This adds support for resource-entity separation in the case of one-to-many. 
+                /// todo: currently there is no support for many to many relations.
+                if (relationshipAttr is HasManyAttribute hasMany) 
+                        relationshipValueList = GetEntityResourceSeparationValue(entity, hasMany) ?? relationshipValueList;
+                if (relationshipValueList == null) return null;
+                return GetTrackedManyRelationshipValue(relationshipValueList, relationshipAttr, ref wasAlreadyAttached);
+            }
+        }
+
+        private IList GetTrackedManyRelationshipValue(IEnumerable<IIdentifiable> relationshipValueList, RelationshipAttribute relationshipAttr, ref bool wasAlreadyAttached)
+        {
+            if (relationshipValueList == null) return null;
+            bool _wasAlreadyAttached = false;
+            var trackedPointerCollection = relationshipValueList.Select(pointer =>
+            {
+                var tracked = AttachOrGetTracked(pointer);
+                if (tracked != null) _wasAlreadyAttached = true;
+                return Convert.ChangeType(tracked ?? pointer, relationshipAttr.Type);
+            }).ToList().Cast(relationshipAttr.Type);
+            if (_wasAlreadyAttached) wasAlreadyAttached = true;
+            return (IList)trackedPointerCollection;
+        }
+
+        private IIdentifiable GetTrackedHasOneRelationshipValue(IIdentifiable relationshipValue, HasOneAttribute hasOneAttribute, ref bool wasAlreadyAttached)
+        {
+
+            var tracked = AttachOrGetTracked(relationshipValue);
+            if (tracked != null) wasAlreadyAttached = true;
+            return tracked ?? relationshipValue;
+        }
+
+        /// <inheritdoc />
         public async Task UpdateRelationshipsAsync(object parent, RelationshipAttribute relationship, IEnumerable<string> relationshipIds)
         {
             // TODO: it would be better to let this be determined within the relationship attribute...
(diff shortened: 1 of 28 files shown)


src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
```diff
@@ -147,217 +144,186 @@ public virtual async Task<TEntity> GetAndIncludeAsync(TId id, string relationshi
...
-        public virtual async Task<TEntity> UpdateAsync(TId id, TEntity entity)
+        [Obsolete("Use overload UpdateAsync(TEntity updatedEntity): providing parameter ID does no longer add anything relevant")]
+        public virtual async Task<TEntity> UpdateAsync(TId id, TEntity updatedEntity)
         {
-            var oldEntity = await GetAsync(id);
+            return await UpdateAsync(updatedEntity);
+        }
 
+        /// <inheritdoc />
+        public virtual async Task<TEntity> UpdateAsync(TEntity updatedEntity)
```
wisepotato: So, I get what's happening here, but can you use an example to signify what's happening here? The technical documentation is great, but a solid example for easy reading is great for when you haven't had your coffee yet.

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
```diff
@@ -470,5 +432,119 @@ public async Task<IReadOnlyList<TEntity>> ToListAsync(IQueryable<TEntity> entiti
...
         }
+
+
+        /// <summary>
+        /// Before assigning new relationship values (UpdateAsync), we need to
+        /// attach the current relationship state to the dbcontext, else 
+        /// it will not perform a complete-replace which is required for 
+        /// one-to-many and many-to-many.
+        /// </summary>
+        protected void LoadCurrentRelationships(TEntity oldEntity, RelationshipAttribute relationshipAttribute)
```
wisepotato: So here I would make it apparent that we're fetching actual database values

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
```diff
@@ -147,217 +144,186 @@ public virtual async Task<TEntity> GetAndIncludeAsync(TId id, string relationshi
...
-                        relationship.Key.SetValue(oldEntity, relationship.Value);
-                    }
-                    if ((relationship.Key.TypeId as Type).IsAssignableFrom(typeof(HasManyAttribute)))
-                    {
-                        await _context.Entry(oldEntity).Collection(relationship.Key.InternalRelationshipName).LoadAsync();
-                        var value = PreventReattachment((IEnumerable<object>)relationship.Value);
-                        relationship.Key.SetValue(oldEntity, value);
-                    }
-                }
+                LoadCurrentRelationships(oldEntity, relationshipAttr);
```
wisepotato: Could you refactor this into one class?

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
```diff
@@ -470,5 +432,119 @@ public async Task<IReadOnlyList<TEntity>> ToListAsync(IQueryable<TEntity> entiti
...
+            }
+        }
+
+        /// <summary>
+        /// assigns relationships that were set in the request to the target entity of the request
+        /// todo: partially remove dependency on IJsonApiContext here: it is fine to
+        /// retrieve from the context WHICH relationships to update, but the actual
+        /// values should not come from the context.
+        /// </summary>
+        private void AssignRelationshipValue(TEntity oldEntity, object relationshipValue, RelationshipAttribute relationshipAttribute)
```
wisepotato: Dont leave todos in summaries

src/JsonApiDotNetCore/Extensions/TypeExtensions.cs
```diff
@@ -8,6 +8,13 @@ namespace JsonApiDotNetCore.Extensions
 {
     internal static class TypeExtensions
     {
+        public static IEnumerable Cast(this IEnumerable source, Type type)
```
wisepotato: comment

src/JsonApiDotNetCore/Formatters/JsonApiReader.cs
```diff
@@ -62,6 +83,38 @@ public Task<InputFormatterResult> ReadAsync(InputFormatterContext context)
             }
         }
 
+        private bool CheckForId(object model)
```
wisepotato: comment

src/JsonApiDotNetCore/Formatters/JsonApiReader.cs
```diff
@@ -62,6 +83,38 @@ public Task<InputFormatterResult> ReadAsync(InputFormatterContext context)
...
+                if (string.IsNullOrEmpty(ro.Id)) return true;
+            }
+            else if (model is IIdentifiable identifiable)
+            {
+                if (string.IsNullOrEmpty(identifiable.StringId)) return true;
+            }
+            return false;
+        }
+
+        private bool CheckForId(IList modelList)
```
wisepotato: comment

src/JsonApiDotNetCore/Internal/InverseRelationships.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// This service is instantiated in the configure phase of the application.
+    /// 
+    /// When using a data access layer different from EF Core, and when using ResourceHooks
+    /// that depend on the inverse navigation property (BeforeImplicitUpdateRelationship),
+    /// you will need to override this service, or pass along the inverseNavigationProperty in
+    /// the RelationshipAttribute.
+    /// </summary>
+    public interface IInverseRelationships
+    {
+        void Resolve();
```
wisepotato: comment

src/JsonApiDotNetCore/Internal/InverseRelationships.cs
```diff
@@ -0,0 +1,65 @@
...
+        private readonly ResourceGraph _graph;
+        private readonly IDbContextResolver _resolver;
+
+        public InverseRelationships(IResourceGraph graph, IDbContextResolver resolver = null)
+        {
+            _graph = (ResourceGraph)graph;
+            _resolver = resolver;
+        }
+
+        public void Resolve()
```
wisepotato: inherit

src/JsonApiDotNetCore/Models/RelationshipAttribute.cs
```diff
@@ -15,8 +15,10 @@ protected RelationshipAttribute(string publicName, Link documentLinks, bool canI
         }
 
         public string PublicRelationshipName { get; internal set; }
-        public string InternalRelationshipName { get; internal set; } 
-        
+        public string InternalRelationshipName { get; internal set; }
+
```
wisepotato: empty space?

src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs
```diff
@@ -155,7 +155,11 @@ public object DocumentToObject(ResourceObject data, List<ResourceObject> include
                         continue;
                     var convertedValue = ConvertAttrValue(newValue, attr.PropertyInfo.PropertyType);
                     attr.SetValue(entity, convertedValue);
-                    _jsonApiContext.AttributesToUpdate[attr] = convertedValue;
+                    /// todo: as a part of the process of decoupling JADNC (specifically
```
wisepotato: add a reference or something like: #512  reference

src/JsonApiDotNetCore/Services/EntityResourceService.cs
```diff
@@ -154,7 +154,7 @@ public virtual async Task<TResource> UpdateAsync(TId id, TResource resource)
         {
             var entity = MapIn(resource);
 
-            entity = await _entities.UpdateAsync(id, entity);
```
wisepotato: why not immediatly into mapout?

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/UpdatingDataTests.cs
```diff
@@ -55,8 +55,9 @@ public async Task Response400IfUpdatingNotSettableAttribute()
 
             var content = new
             {
-                datea = new
```
wisepotato: nice.

src/JsonApiDotNetCore/Extensions/TypeExtensions.cs
maurei: fixed in c34d22ac1af94c50d735e09d1d99759200cd0d6f

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
maurei: fixed in bf520a06ba81bb0df9e08a0ec43fdd83982b78b2

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
maurei: Yes: it would be nice to have a helper class to make it all more readable. But leaving this as a stretch goal for the future. This shouldn't be a problem as it won't introduce any breaking changes, so we can refactor this whenever we like.

src/JsonApiDotNetCore/Models/RelationshipAttribute.cs
maurei: Diff seems to be misleading, empty space is not actually there

src/JsonApiDotNetCore/Services/EntityResourceService.cs
maurei: Easy to fix; but limiting this PR to upgrading the repository

src/JsonApiDotNetCore/Internal/InverseRelationships.cs
maurei: fixed in 457e93dbb38784220565d687c799acadf82350d1

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
maurei: fixed in 457e93dbb38784220565d687c799acadf82350d1

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
maurei: fixed in 457e93dbb38784220565d687c799acadf82350d1

src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs
maurei: fixed in 457e93dbb38784220565d687c799acadf82350d1

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
```diff
@@ -62,7 +62,8 @@ public class DefaultEntityRepository<TEntity, TId>
             _resourceDefinition = resourceDefinition;
         }
 
-        public DefaultEntityRepository(
+        public
```
wisepotato: what?

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
```diff
@@ -171,6 +172,13 @@ public virtual async Task<TEntity> CreateAsync(TEntity entity)
         /// <summary>
         /// Loads the inverse relationships to prevent foreign key constraints from being violated
         /// to support implicit removes, see https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/502.
+        /// 
+        /// example:
```
wisepotato: add in <example> tag

src/JsonApiDotNetCore/Extensions/TypeExtensions.cs
```diff
@@ -8,6 +8,32 @@ namespace JsonApiDotNetCore.Extensions
 {
     internal static class TypeExtensions
     {
+        public static void AddRange<T>(this IList list, IEnumerable<T> items)
```
wisepotato: Comment

maurei: This code wasn't added in this PR, don't know why the diff shows it as such

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
maurei: The explanation that follows isn't something that in the documentation should be generated as a code example. Edited it to be more readable.

src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs
```diff
@@ -172,8 +171,8 @@ public virtual async Task<TEntity> CreateAsync(TEntity entity)
         /// <summary>
         /// Loads the inverse relationships to prevent foreign key constraints from being violated
         /// to support implicit removes, see https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/502.
-        /// 
-        /// example: 
+        /// <para/>
```
wisepotato: para? `<example>`

PR #529, opened 2019-06-20
Improving testability and intuitivity of hooks 

maurei:
We found that when implementing resource hooks in our application, it wasn't very easy to unit test them. This was because 1) the resource helpers `AffectedRelationships`, `AffectedResources`, and `AffectedResourceDiff` were hard/impossible to mock/instantiate for unit testing, and because 2) these types don't give an intuitive feeling of the data structure they represent.

This PR fixes this by introducing following changes

* made the constructors of the affected resource/relationship helper classes public so they can be instantiated
* these public constructors now have clear dependencies
* got rid of `RelationshipProxy` where the use of `RelationshipAttribute` suffices; this should make it easier to instantiate `AffectedRelationships`.
* `AffectedRelationships`, `AffectedResources` and `AffectedResourceDiff` have been renamed to `RelationshipsDictionary`, `EntityHashSet` and `EntityDiffs` to make clear that (for the first two) you're actually just dealing with a dictionary and a hashset.


src/JsonApiDotNetCore/Hooks/IResourceHookContainer.cs
```diff
@@ -194,7 +140,69 @@ public interface IBeforeHooks<TEntity> where TEntity : class, IIdentifiable
         /// <returns>The transformed set of ids</returns>
         /// <param name="resourcesByRelationship">A helper that groups the entities by the affected relationship</param>
         /// <param name="pipeline">An enum indicating from where the hook was triggered.</param>
-        void BeforeImplicitUpdateRelationship(IAffectedRelationships<TEntity> resourcesByRelationship, ResourcePipeline pipeline);
+        void BeforeImplicitUpdateRelationship(IRelationshipsDictionary<TEntity> resourcesByRelationship, ResourcePipeline pipeline);
+    }
+
+
+
```
wisepotato: empty spaaace

src/JsonApiDotNetCore/Internal/TypeHelper.cs
```diff
@@ -1,11 +1,13 @@
...
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
+using JsonApiDotNetCore.Models;
 
 namespace JsonApiDotNetCore.Internal
 {
-    public static class TypeHelper
+    internal static class TypeHelper
```
wisepotato: is this not needed for unti tests??

test/UnitTests/Data/DefaultEntityRepository_Tests.cs
```diff
@@ -77,14 +77,15 @@ public async Task UpdateAsync_Updates_Attributes_In_AttributesToUpdate()
 
         private DefaultEntityRepository<TodoItem> GetRepository()
         {
+
```
wisepotato: empty spaaace

test/UnitTests/ResourceHooks/DiscoveryTests.cs
```diff
@@ -14,7 +14,7 @@ public class DummyResourceDefinition : ResourceDefinition<Dummy>
         {
             public DummyResourceDefinition() : base(new ResourceGraphBuilder().AddResource<Dummy>().Build()) { }
```
wisepotato: space not needed here

src/JsonApiDotNetCore/Internal/TypeHelper.cs
maurei: Nopes!

src/Examples/JsonApiDotNetCoreExample/Controllers/TodoCollectionsController.cs
```diff
@@ -33,7 +33,7 @@ public override async Task<IActionResult> PatchAsync(Guid id, [FromBody] TodoIte
             if (entity.Name == "PRE-ATTACH-TEST")
             {
                 var targetTodoId = entity.TodoItems.First().Id;
-                var todoItemContext = _dbResolver.GetDbSet<TodoItem>();
+                var todoItemContext = _dbResolver.GetContext().Set<TodoItem>();
```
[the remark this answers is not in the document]
maurei: Now it uses a EF core feature which wasn't there yet when JADNC introduced its own extension method to achieve the same  (`GetDbSet<T>`)

PR #532, opened 2019-06-21
fix: #528

### LINKED ISSUE #528: SetHasOneForeignKeyValue detects string as non-nullable type and throws exception
## Description
I have a 0..1 to 1 relationship using foreign key of type string. As string can be null so it should be fine to use it as a foreign key in this scenario but I'm getting an exception thrown from ```SetHasOneForeignKeyValue``` method:

"Cannot set required relationship identifier '{hasOneAttr.IdentifiablePropertyName}' to null because it is a non-nullable type."

I believe this is because the method uses this code for the check: ```Nullable.GetUnderlyingType(foreignKeyProperty.PropertyType)```
which returns null for type string.

That is because GetUnderlyingType only works as expected here for types derived from Nullable<T> which string is not.
https://referencesource.microsoft.com/#mscorlib/system/nullable.cs,132
...

## Environment

PiotrBrzezianski:
Allowing string foreign keys on optional HasOne relationship

Closes #528

#### BUG FIX
- [x] fix issue
- [ ] bump package version

### DIFF
--- a/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs
+++ b/src/JsonApiDotNetCore/Serialization/JsonApiDeSerializer.cs
@@ -252,7 +252,9 @@ namespace JsonApiDotNetCore.Serialization
 
                 // e.g. PATCH /articles
                 // {... { "relationships":{ "Owner": { "data": null } } } }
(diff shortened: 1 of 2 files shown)


maurei:
Looking good, thanks!

PR #549, opened 2019-08-26
Added value types support to the filters isnull & isnotnull.

### LINKED ISSUE #548: There is no good way to filter by the parent with isnull or isnotnull operators
## Description

There is no good way to filter by the parent with isnull or isnotnull operators.

If I have a class Child referencing the Parent, both use int as a Key, then I cannot use a filter query like this
`api/children?filter[parent.id]=isnull:`
— it fails to create the equality expression (because you can't compare int and null).

There are 2 bad workarounds for this right now:

1. Add the [Attr] attribute to the Parent property of the Child and use the following filter instead
`api/children?filter[parent]=isnull:`
It causes the loading of the whole Parent hierarchy into attributes.
2. Change the Key type from int to Nullable<int>. It's nonsense to have a nullable primary key.

## Environment

- JsonApiDotNetCore Version: the current master branch

unicrus:
Closes #548

### DIFF
--- a/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
+++ b/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
@@ -170,11 +170,29 @@ namespace JsonApiDotNetCore.Extensions
                     break;
                 case FilterOperations.isnotnull:
                     // {model.Id != null}
-                    body = Expression.NotEqual(left, right);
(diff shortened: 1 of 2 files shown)


wisepotato:
Hi i cant accept this without tests, also the tests are failing :)

unicrus:
@wisepotato , I fixed the failing tests and added 2 new ones.

PR #554, opened 2019-09-09
Decoupling context

maurei:
Work so far put in to get tests up with the current introduction of new decoupled services like `RequestManager`, etc.


maurei:
I accidentally pushed my serializers commits to this branch instead of to a separate serializer-decoupling branch. I rolled back and opened up a new PR  in #557 for the same commit as this PR (this one couldn't be merged anymore because GitHub does not allow PRs for force-pushed branches)

Issue #576, opened 2019-10-15
Future of this package?


house-md:
I would like to ask about future of this package. As I see from all the changes to the code and tests, it seems that this package completely lost it's original direction and purpose which was to be framework for easier implementition of JSON:API standard. Are there any reasons for these changes?

It seems that this package became a complete mess since version 3.1.0 and absolutely incompatible. Like with removal of JsonApiContext which was replaced with four or more different services? What purpose does this change have? And this is only beginning. For example why did You removed ResourceEntitySeparation and mapping logic or Bulk operations? Removed possibility to control if client can generate IDs. Dropped support for nested sparse fields because they can be Attributes and also Relationships. 

Current difference between v3 and v4 is more than 400 changed files. With current progress You are making it almost impossible to migrate from v3 to v4.

maurei:
Thank you for expressing your interests and concerns about the direction of JADNC.
<br/>

> I would like to ask about future of this package [...] seems that this package completely lost it's original direction and purpose [...] Are there any reasons for these changes?

In general, the goal of the framework is to become more production ready. For v4 this primarily means 
 1. to introduce proper testability of apps that are built on JADNC (apps that are not testable are not production ready, also see #251)
 2. to be able to implement authorization without having to entirely override/rewrite the service layer (the scope of JADNC would be very limited if it is too complicated to implement such a basic thing,  see #552, #547,  #473)
 3. to stabilize the core API  (i.e. fixing violations of the json:api spec and fixing/removing broken features in general).

At the same we are trying to keep the codebase simple and maintainable because of the very limited development resources we have available. We have insufficient capacity to fully support every broken/incomplete feature that is currently in v3.1 and as such we are forced to make a selection. We try to prioritise those features that are crucial to the json:api spec and the production-readiness of the framework, and as a result we will need to drop support for certain other features.
<br/>
>  Like with removal of JsonApiContext which was replaced with four or more different services? What purpose does this change have? 

The decoupling of the framework through the removal of `JsonApiContext` is a crucial step towards improved testability (and therefore production-readiness).`JsonApiContext` contains a high amount of strongly diverging responsibilities and it is injected in almost every service throughout the framework. This is a serious problem because it strongly violates Single Responsibility: it is a [god object](https://sourcemaking.com/antipatterns/the-blob) anti-pattern.

[...]

PR #579, opened 2019-10-16
Cleaning repository of remaining business logic

maurei:
This PR contains the the last touches of the `DefaultEntityRepository` for decoupling and consistency.

- Removed the last bit of business logic that should be in service layer (the retrieval of custom filter queries; moved this to `FilterQueryContext`). As such the repository layer is now free of business logic, i.e. no more dependency on `ResourceDefinition`. 
- Refactored the ef-core specific interface `DetachRelationshipPointers`. The service layer should not be responsible to instruct the repo to do such EF core specific stuff. This is now called and handled internally in the repo itself.
- Removed compound methods on repository layer. For example, the repository contained a `Get(TId)` and `Include(IQueryable<TResource>, RelationshipAttribute)` as well as a `GetAndIncludeRelationship(id, relationship)` method. Such compound methods that are basically a composition of other smaller other public methods did not add value.
     - as a consequence, all logic in the `EntityResourceService` that relied on these compound methods have been rerouted. 
- Replaces `SingleOrDefaultAsync` calls with `FirstOrDefaultAsync` calls, which performs better.
- ~~Restored any removed method with respect to v3.2 and marked them obsolete.~~

~~#### About the obsolete tags~~
~~This is in consideration of #576. The idea is the following: instead of directly deleting everything we will re-add the deleted methods so far but mark them obsolete. This does not mean that we will add full backward compatibility, this is technically not required by semver as we're upping the major version. But:~~
~~- when it's not too much work we should consider adding simple backward compatibility.~~
     -~~ eg in the case when a method simply needs to be rerouted, like with `GetAndIncludeRelationshipAsync(id, relationship)` -> `Get()` and `Include()`. ~~
~~- For ones that are too complicated to add backward compatibility we will directly set to `error: true`. Developers can then still be directed to the precise part of the migration guide instead of being surprised with completely unknown methods.~~
~~This approach would make migration to v4 *somewhat* easier as developers will not have to rewrite methods for which it is not strictly necessary and we can add directed hints where it *is* strictly necessary~~
~~Then some time later we could release v5 removing the legacy stuff completely.~~

Will not go through the hassle of introducing deprecations in hindsight and removing them again in v5. Instead will open a tracking issue proposing to make a v3.3 which would introduce these deprecations to make migration from v3 to v4 easier. If the demand is high enough, we will do it.


src/JsonApiDotNetCore/Models/ResourceDefinition.cs
```diff
@@ -99,9 +99,12 @@ public void HideFields(Expression<Func<TResource, dynamic>> selector)
         /// </example>
         public virtual QueryFilters GetQueryFilters() => null;
 
-        public bool HasCustomQueryFilter(string key)
+        public object GetCustomQueryFilter(string key)
```
wisepotato: i mean there must be a base class right?

src/JsonApiDotNetCore/Models/ResourceDefinition.cs
```diff
@@ -15,7 +15,7 @@ public interface IResourceDefinition
     {
         List<AttrAttribute> GetAllowedAttributes();
         List<RelationshipAttribute> GetAllowedRelationships();
-        bool HasCustomQueryFilter(string key);
+        object GetCustomQueryFilter(string key);
```
wisepotato: why is this returning object?

Issue #568, opened 2019-10-09
Consider simplifying constructor signatures in controller/repo/service layer


wisepotato:
Its only needed here:

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/8570b216c7ede3a7b5f9fc7a4912404a14926df5/src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs#L158-L160

So.. we might be able to get rid of that.

maurei:
#### Controller layer
 * Remove `ResourceGraph` dependency in controller

#### Service layer
 * Consider having a wrapping service to decrease size of constructors, allowing for easier extensibility
* Consider having internal constructors with signatures that allow for pipeline testing

#### Repo layer
 * Consider having a wrapping service to decrease size of constructors, allowing for easier extensibility
* Consider having internal constructors with signatures that allow for pipeline testing

maurei:
We could consider to do something like this.

Instead of
```c#
        public DefaultResourceService(
                ISortService sortService,
                IFilterService filterService,
                IIncludeService includeService,
                ISparseFieldsService sparseFieldsService,
                IPageService pageManager,
                IJsonApiOptions options,
                IResourceRepository<TResource, TId> repository,
                IResourceContextProvider provider,
                IResourceHookExecutor hookExecutor = null,
                ILoggerFactory loggerFactory = null)
        {
            _includeService = includeService;
            _sparseFieldsService = sparseFieldsService;
            _pageManager = pageManager;
            _options = options;
            _sortService = sortService;
            _filterService = filterService;
            _repository = repository;
            _hookExecutor = hookExecutor;
            _logger = loggerFactory?.CreateLogger<DefaultResourceService<TResource, TId>>();
            _currentRequestResource = provider.GetResourceContext<TResource>();
        }

public DefaultResourceService(IEnumerable<IQueryParameterService> queryServices,  ....)
```

```c#
        public DefaultResourceService(
                IEnumerable<IQueryParameterService> queryParameters
                IJsonApiOptions options,
                IResourceRepository<TResource, TId> repository,
                IResourceContextProvider provider,
                IResourceHookExecutor hookExecutor = null,
                ILoggerFactory loggerFactory = null)
        {
            _includeService = queryParameters.Single(qp => qp is IIncludedService);
            _sparseFieldsService =  queryParameters.Single(qp => qp is ISparseFieldsService);
[...]

PR #586, opened 2019-10-21
Separation of concerns ResourceGraph

### LINKED ISSUE #501: Calling 'AddJsonApi' bypass all the exception handler middlewares
## Description

Before using JsonApiDotNetCore, I had an ASP.NET Core middleware that had the responsibility to reformat all exceptions that occured in my controllers. This middleware was almost identical to [this one](http://anthonygiretti.com/2018/09/04/asp-net-core-2-1-middlewares-part1-building-a-custom-middleware/).

As soon as I started to use JsonApiDotNetCore by calling `AddJsonApi`, my middleware stopped working. This is because AddJsonApi registers an exception filter:

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/8570b216c7ede3a7b5f9fc7a4912404a14926df5/src/JsonApiDotNetCore/Extensions/IServiceCollectionExtensions.cs#L82

This exception filter handles **all** exceptions. My middleware is therefore unable to detect the exception.

Is there a way to disable this exception filter or to limit this exception filter to the routes that to use JSONAPI? My routes that don't use JSONAPI should still use my exception middleware.

## Environment

- JsonApiDotNetCore Version: 3.1.0.0

### LINKED ISSUE #569: Separation of concerns ResourceGraph
Currently there are two reasons why `IResourceGraph` is injected anywhere in the application:

- Retrieving `ContextEntity` instances
- Resolving relationships
	- abstracting away has-many-through stuff
	- resolving inverse relationships.

These two deviating responsibilities can be put behind their own interface:

- IContextEntityProvider
- IRelationshipResolver

The latter could potentially be merged with the current `IFieldExplorer` service.

maurei:
Closes #569

This PR contains work on `ResourceGraph`, which contained a bunch of deviating responsibilities. These have been moved to separated services/locations and as a result the resource graph now contains one clear task: to provide insight in which resources are exposed and how they are exposed (i.e. access to metadata in `AttrAttributes` and `RelationshipAttributes`).

For this, the following pieces of logic have been moved out of `ResourceGraph`

#### Getters of has many through relationship
```c#
// IResourceGraph.cs
object GetRelationship<TParent>(TParent resource, string propertyName); 
object GetRelationshipValue<TParent>(TParent resource, RelationshipAttribute relationship) where TParent : IIdentifiable; 
string GetRelationshipName<TParent>(string relationshipName); 
```
The above methods were helper methods that took care of retrieving the correct relationship values in the case of `HasManyThrough`. For all other field classes (`Attr`, `HasOneRelationship`, `HasManyRelationship`), this logic was not defined on the resource graph but contained in these classes itself. Now that responsibility for many-to-many relationship has also been moved to  the `HasManyThroughRelationship` class itself. It results in cleaner code throughout the framework because the following (which occurred in various places) is no longer needed:
```c#
if (relationship is HasManyThrough hm)
    value = _resourceGraph.GetRelationshipValue(entity, hm)
else 
   value =  relationship.GetValue(entity)
```
and the same for the setting of many-to-many relationships.

#### GetPublicAttributeName
```c#
// IResourceGraph.cs
string GetPublicAttributeName<TParent>(string internalAttributeName); 
```
This method was only referenced once from the controller layer, and here it was not strictly necessary. Refactored the related code in the controller which allowed the method to be removed from `IResourceGraph`. Also as a result the controller layer now no longer depends on `IResourceGraph` (which partially covers #568).

#### UsesDbContext
```c#
// IResourceGraph.cs
bool UsesDbContext { get; } // if the graph was built using a EF core dbcontext
```
The resource graph should not care nor expose how it was built: this metadata is only relevant during startup configuration. It was exposed only because of a design choice in the configuration flow when calling `.AddJsonApi( ... )`, and removing `UsesDbContext` from `IResourceGraph` required this configuration flow to be altered slightly:
* Previously, `IResourceGraphBuilder` and `IResourceGraph` were properties on `JsonApiOptions`, while simultaneously `IResourceGraphBuilder` contained a *static reference* to `JsonApiOptions`. 
* Now, `IResourceGraphBuilder` consumes a `JsonApiOptions` (as well as a few other things) and produces a `IResourceGraph`.

This altered startup flow is now wrapped in a dedicated class `JsonApiApplicationBuilder` instead of spread over various extension methods of `IServiceCollections`. As a result, `UsesDbContext` is no longer needed on the resource graph.  It also introduces configurability of the startup sequence: eg a developer can now override the global exception filter that is configured when calling `AddJsonApi()`, which closes #501. 

#### GetEntityFromControllerName
```c#
// IResourceGraph.cs
ContextEntity GetEntityFromControllerName(string pathParsed);
```
Another somewhat deviating responsibility was matching controllers to registered resources. This has been moved to a dedicated service: `IControllerResourceMapping` which is now consumed in the middleware. 

#### Remaining changes
As a result of removing the above methods, `IResourceGraph` contained no other responsibilities other than retrieving `ContextEntity`s. The `IFieldsExplorer` service had more or less the same responsibility: to get `ContextEntity`s and return their exposed fields. It was suitable to merge services into the `IResourceGraph`.


wisepotato:
fix merge conflict @maurei

maurei:
> fix merge conflict @maurei

Not anymore

PR #592, opened 2019-10-23
Simplify DefaultResourceService constructor

### LINKED ISSUE #568: Consider simplifying constructor signatures in controller/repo/service layer

maurei:
Usage of `IEnumerable<IQueryParameterService>` in `DefaultResourceService` constructor as opposed to individually injecting every member of that `IEnumerable`

Closes #568

### DIFF
--- a/src/JsonApiDotNetCore/Services/DefaultResourceService.cs
+++ b/src/JsonApiDotNetCore/Services/DefaultResourceService.cs
@@ -34,23 +34,19 @@ namespace JsonApiDotNetCore.Services
         private readonly ResourceContext _currentRequestResource;
 
         public DefaultResourceService(
-                ISortService sortService,
-                IFilterService filterService,
+                IEnumerable<IQueryParameterService> queryParameters,
                 IJsonApiOptions options,
-                IIncludeService includeService,
-                ISparseFieldsService sparseFieldsService,
-                IPageService pageManager,
                 IResourceRepository<TResource, TId> repository,
                 IResourceContextProvider provider,
                 IResourceHookExecutor hookExecutor = null,
                 ILoggerFactory loggerFactory = null)
         {
-            _includeService = includeService;
-            _sparseFieldsService = sparseFieldsService;
-            _pageManager = pageManager;
+            _includeService = queryParameters.FirstOrDefault(qp => qp is IIncludeService) as IIncludeService;
+            _sparseFieldsService = queryParameters.FirstOrDefault(qp => qp is ISparseFieldsService) as ISparseFieldsService;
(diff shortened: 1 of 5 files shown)


src/JsonApiDotNetCore/Services/DefaultResourceService.cs
```diff
@@ -34,23 +34,19 @@ public class DefaultResourceService<TResource, TId> :
...
-                IPageService pageManager,
                 IResourceRepository<TResource, TId> repository,
                 IResourceContextProvider provider,
                 IResourceHookExecutor hookExecutor = null,
                 ILoggerFactory loggerFactory = null)
         {
-            _includeService = includeService;
-            _sparseFieldsService = sparseFieldsService;
-            _pageManager = pageManager;
+            _includeService = queryParameters.FirstOrDefault(qp => qp is IIncludeService) as IIncludeService;
```
wisepotato: this is still a dependency and will still make testing a chore, are you sure about this?

wisepotato: also, might as well make a nice `get` function to make this a bit more readable: `queryParameters.Get<IIncludeService>();

wisepotato: or is this done to use the `IEnumerable` DI trick?

maurei: That's right: the various the `DefaultResourceService` still depend on the various query parameters. However, if we use `IEnumerable<IQueryParameterService>` as opposed to injecting every service individually, then:
* we would leverage the ability to access all query param services at once through that interface which is already available anyway, so no need to introduce/register an new `IQueryParameterAccessor` wrapper service for that
* Testing becomes a bit easier: eg if you want to test a part of the `DefaultResourceService` that only requires **one** queryparam service, you can just add **only that** service to this `IEnumerable`  instead of always having to mock all of them
* if in the future we decide to add a new built-in query parameter service, we will not have to change the `DefaultResourceService` constructor by adding it

maurei: With an extension method on `IEnumerable<IQueryParameterService>` it was possible to get a more readable getter:

```c#
_includeService = queryParameters.FirstOrDefault<IIncludeService>();
```

Issue #559, opened 2019-10-01
Upgrade to .net core 3.0 


maurei:
As title says

emouawad:
Not sure if this is related:

`System.Reflection.ReflectionTypeLoadException: 'Unable to load one or more of the requested types.
Method 'CommitAsync' in type 'JsonApiDotNetCore.Extensions.SafeTransactionProxy' from assembly 'JsonApiDotNetCore, Version=4.0.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.'`

on

          app.UseEndpoints(endpoints =>
            {
                endpoints.MapControllerRoute(
                    name: "default",
                    pattern: "{controller=Home}/{action=Index}/{id?}");
            });

wisepotato:
Working on this atm, will post back with updates.

wayne-o:
Would this have made it into 3.1.0?

maurei:
No, the .NET core 3.0 upgrade has been introduced in prerelease v4.0.0-alpha4

PR #593, opened 2019-10-23
.NET Core 2.0 -> 2.1

wisepotato:
First work in #559 

* upgrade .net core 2.0 to .net core 2.1 (ignore branch name)
* Cleanup in tests (warnings have been ignored until now)
* new startup format
* catch some exceptions that are too broad (exception -> InvalidOperationException)


.editorconfig
```diff
@@ -22,4 +22,8 @@ dotnet_naming_symbols.private_fields.applicable_kinds           = field
 dotnet_naming_symbols.private_fields.applicable_accessibilities = private
 
 dotnet_naming_style.prefix_underscore.capitalization = camel_case
-dotnet_naming_style.prefix_underscore.required_prefix = _
\ No newline at end of file
+dotnet_naming_style.prefix_underscore.required_prefix = _
+[*.cs]
+
+# CS0659: Type overrides Object.Equals(object o) but does not override Object.GetHashCode()
+dotnet_diagnostic.CS0659.severity = silent
```
maurei: This is to suppress the error in `RelationshipAttribute.cs`, right? Are we sure it's fine to just ignore it rather than solving the issue? I don't know: I always ignored it until now, never looked into it.

src/Examples/GettingStarted/Program.cs
```diff
@@ -10,17 +10,17 @@
...
-            BuildWebHost(args).Run();
-        }
-
-        public static IWebHost BuildWebHost(string[] args) =>
-            WebHost.CreateDefaultBuilder(args)
-                .UseStartup<Startup>()
-                .UseUrls("http://localhost:5001")
-                .Build();
+
+
```
maurei: minor styling: scrap one whitespace. No need to update this in this PR, do it in one of your followings and resolve this comment for now

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/CreatingDataTests.cs
```diff
@@ -340,7 +342,7 @@ public async Task CreateRelationship_ToManyWithImplicitRemove_IsCreated()
             var oldPersonDb = dbContext.People.AsNoTracking().Where(p => p.Id == currentPerson.Id).Include(e => e.TodoItems).Single();
             AssertEqualStatusCode(HttpStatusCode.Created, response);
             Assert.Equal(2, newPersonDb.TodoItems.Count);
-            Assert.Equal(1, oldPersonDb.TodoItems.Count);
+            Assert.Single(oldPersonDb.TodoItems);
```
maurei: 👍

.editorconfig
wisepotato: I'll take it out, so it stays as a warning next release

PR #601, opened 2019-10-30
Fix/pagesize

### LINKED ISSUE #599: No DefaultPageSize leads to division by zero error
## Description
Using a `Startup` without defining a `IJsonApiOptions.DefaultPagesize` leads to a `System.DivideByZeroException` upon a get request.

maurei:
Closes #599

Fixes a division by zero exception that was thrown for regular get requests in the case where no `DefaultPageSize` was set.

Also removes redundant `IPageService.DefaultPageSize` which wasn't referenced anywhere.

- [x] fix issue


wisepotato:
Let's first check if `WebApplicationFactory` is a valid replacement for putting the `Startup` in the Assembly `JsonApiDotNetCoreExample`, I would like seperation of concerns here. The rest of the PR is fine.

maurei:
> Let's first check if `WebApplicationFactory` is a valid replacement for putting the `Startup` in the Assembly `JsonApiDotNetCoreExample`, I would like seperation of concerns here. The rest of the PR is fine.

Tried to get this to work, but no luck. See https://github.com/aspnet/AspNetCore/issues/15373#issuecomment-548287685.

PR #616, opened 2019-11-05
Move EF core related building to extension methods

maurei:
This PR fixes the inability to use auto discovery when setting up a JADNC application with the ef-core minded `AddJsonApi<TDbContext>` overload. It does so by moving all EF Core related configuration to extension methods instead of them being hardcoded in the interface of `JsonApiApplicationBuilder`. This is also an important step towards decoupling EF core from the core of JADNC.

Inspired by https://dotnetcoretutorials.com/2017/01/24/servicecollection-extension-pattern/


maurei:
@wisepotato it seems that default interface members with `internal` access modifiers are not usable the way we expected it (unless you're going for explicit implementations), see https://stackoverflow.com/questions/58734707/using-static-internal-and-protected-access-modifiers-in-interfaces-c-sharp-8

Either way I've refactored the extensions methods to no longer need the internals of `ResourceGraphBuilder`, which resolves all of your comments

PR #617, opened 2019-11-06
Hooks discovery detection of implemented hooks issue

### LINKED ISSUE #615: Need guidance for resource hooks
## Description
I have been trying to get Resource Hooks working in v4.0.0-alpha3. I have followed the steps in the docs, but the hook is never hit.

The only way I have actually gotten it to work is to add an explicit call to `AddScoped<IResourceHookContainer<WeatherForecast>, WeatherForecastDefinition>()`, then inject the `IResourceHookContainer<WeatherForecast>` directly into the service. But that seems wrong - looking at the source it should be registered automatically using open generics.

I also run into problems trying to call `BeforeCreate` because I would then have to magically have a `IAffectedResources` object somehow. Which seems only to be available internally through some `TypeHelper` extension methods and internal derived classes.

Getting this to work is a blocker for me using this library in production code because I have to be able to authenticate and authorize at the request level.

Anyway, I have posted a simple repository that reproduces the behavior: https://github.com/crgolden/JsonApiDotNetCoreIssue. Please take a look and let me know what I'm doing wrong.

Thanks!
...

## Environment

- Other Relevant Package Versions: not sure

maurei:
Closes #615

There was a problem with detecting which hooks have been implemented for a particular resource when that resource and its associated resource definition were defined in different assemblies.

This PR:
- contains a fix for that issue: we're no longer expecting assemblies but instead relying on the fact that the hooks container *must* be registered with the DI container anyway, so we might as well resolve it from there
- fixed more occurrences of the `LoaDatabaseValue` typo
- sets the default value of `EnableResourceHooks` to true on `JsonApiOptions`.

### DIFF
--- a/src/JsonApiDotNetCore/Hooks/Discovery/HooksDiscovery.cs
+++ b/src/JsonApiDotNetCore/Hooks/Discovery/HooksDiscovery.cs
@@ -3,7 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using JsonApiDotNetCore.Internal;
 using JsonApiDotNetCore.Models;
-using JsonApiDotNetCore.Services;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace JsonApiDotNetCore.Hooks
 {
@@ -26,13 +26,14 @@ namespace JsonApiDotNetCore.Hooks
         public ResourceHook[] DatabaseValuesEnabledHooks { get; private set; }
         public ResourceHook[] DatabaseValuesDisabledHooks { get; private set; }
 
-        public HooksDiscovery(IScopedServiceProvider provider)
+        public HooksDiscovery(IServiceProvider provider)
         {
             _allHooks = Enum.GetValues(typeof(ResourceHook))
                             .Cast<ResourceHook>()
                             .Where(h => h != ResourceHook.None)
                             .ToArray();
-            var containerType = provider.GetService(_boundResourceDefinitionType)?.GetType();
+            using var scope = provider.CreateScope();
+            var containerType = scope.ServiceProvider.GetService(_boundResourceDefinitionType)?.GetType();
             if (containerType == null || containerType == _boundResourceDefinitionType)
                 return;
             DiscoverImplementedHooksForModel(containerType);
--- a/src/JsonApiDotNetCore/Hooks/Discovery/HooksDiscovery.cs
+++ b/src/JsonApiDotNetCore/Hooks/Discovery/HooksDiscovery.cs
@@ -32,19 +32,27 @@ namespace JsonApiDotNetCore.Hooks
                             .Cast<ResourceHook>()
                             .Where(h => h != ResourceHook.None)
                             .ToArray();
-            using var scope = provider.CreateScope();
-            var containerType = scope.ServiceProvider.GetService(_boundResourceDefinitionType)?.GetType();
-            if (containerType == null || containerType == _boundResourceDefinitionType)
-                return;
-            DiscoverImplementedHooksForModel(containerType);
+
+            Type containerType;
+            using (var scope = provider.CreateScope())
+            {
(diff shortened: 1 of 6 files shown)


src/JsonApiDotNetCore/Hooks/Discovery/HooksDiscovery.cs
```diff
@@ -12,78 +12,65 @@ namespace JsonApiDotNetCore.Hooks
...
         public ResourceHook[] ImplementedHooks { get; private set; }
         public ResourceHook[] DatabaseValuesEnabledHooks { get; private set; }
         public ResourceHook[] DatabaseValuesDisabledHooks { get; private set; }
 
-
-        public HooksDiscovery()
+        public HooksDiscovery(IServiceProvider provider)
         {
             _allHooks = Enum.GetValues(typeof(ResourceHook))
                             .Cast<ResourceHook>()
```
wisepotato: why is this called resourcehook and not resourcedefinition? The naming is a bit off

src/JsonApiDotNetCore/Hooks/Discovery/HooksDiscovery.cs
```diff
@@ -12,78 +12,65 @@ namespace JsonApiDotNetCore.Hooks
...
                             .Where(h => h != ResourceHook.None)
                             .ToArray();
-            DiscoverImplementedHooksForModel();
+            using var scope = provider.CreateScope();
```
wisepotato: by using this the scope extends to longer  longer than you need, the `DiscoverImplementedHooksForModel` does not need the scope, so I would limit it a bit more. By using this syntactic sugar it will extend as long as the local scope is available, I'd make it explicit.

src/JsonApiDotNetCore/Hooks/Discovery/HooksDiscovery.cs
```diff
@@ -12,78 +12,65 @@ namespace JsonApiDotNetCore.Hooks
...
             var implementedHooks = new List<ResourceHook>();
-            var enabledHooks = new List<ResourceHook>() { ResourceHook.BeforeImplicitUpdateRelationship } ;
-            var disabledHooks = new List<ResourceHook>();
-            Type targetType = null;
-            try
-            {
-                targetType = derivedTypes.SingleOrDefault(); // multiple containers is not supported
-            }
-            catch
+            var databaseValuesEnabledHooks = new List<ResourceHook> { ResourceHook.BeforeImplicitUpdateRelationship }; // this hook can only be used with enabled database values
```
wisepotato: stylecop: comments on previous line with newline before it.

src/JsonApiDotNetCore/Hooks/Execution/DiffableEntityHashSet.cs
```diff
@@ -90,7 +90,7 @@ public new HashSet<TResource> GetAffected(Expression<Func<TResource, object>> Na
 
         private HashSet<TResource> ThrowNoDbValuesError()
         {
-            throw new MemberAccessException("Cannot iterate over the diffs if the LoaDatabaseValues option is set to false");
+            throw new MemberAccessException("Cannot iterate over the diffs if the LoadDatabaseValues option is set to false");
```
wisepotato: suggestion: nameof

maurei: 👍

src/JsonApiDotNetCore/Hooks/Discovery/HooksDiscovery.cs
maurei: Fixed

src/JsonApiDotNetCore/Hooks/Discovery/HooksDiscovery.cs
maurei: It's an `enum` representing a hook on a resource definition

PR #625, opened 2019-11-18
Fix: top-level links missing when paging

### LINKED ISSUE #622: Bug: first/prev links not rendered on second page
## Description
When using paging, the `first` and `prev` links are not rendered on the second page.

Example:
First page is at URL /api/books (A), which contains a link to next page: /api/books/?page[number]=1 (B). That page contains a link to its next page: /api/books/?page[number]=2 (C).

Observed behavior:
Page C contains top-level links for `first` and `prev`, but they are missing on page B.

Expected behavior:
I think the correct behavior would be for `first` and `prev` to exist also on page B, because it is the second page.

I traced this down to `LinkBuilder.SetPageLinks(...)`, which conditionally sets them:
```csharp
if (_pageService.CurrentPage > 1)
{
    links.First = GetPageLink(primaryResource, 1, _pageService.PageSize);
    links.Prev = GetPageLink(primaryResource, _pageService.CurrentPage - 1, _pageService.PageSize);
}
```

I think the condition should be changed to > 0, as index 0 appears to be the first page.

## Environment
latest version of Develop branch

maurei:
- Closes bug reported in #622: values for top-level links of `Prev`, `Next` and `Self` were sometimes wrongly generated or missing
- Improved unit tests and added e2e test to reproduce the bug
- Fixed a failing test related to hooks


src/JsonApiDotNetCore/Serialization/Server/Builders/LinkBuilder.cs
```diff
@@ -31,10 +31,15 @@ public TopLevelLinks GetTopLevelLinks(ResourceContext primaryResource)
...
             if (ShouldAddTopLevelLink(primaryResource, Link.Self))
+            {
                 topLevelLinks = new TopLevelLinks { Self = GetSelfTopLevelLink(primaryResource.ResourceName) };
+            }
 
-            if (ShouldAddTopLevelLink(primaryResource, Link.Paging))
-                SetPageLinks(primaryResource, ref topLevelLinks);
+            if (ShouldAddTopLevelLink(primaryResource, Link.Paging) && _pageService.CanPaginate)
+            {
+
```
wisepotato: minor spacing

src/JsonApiDotNetCore/Serialization/Server/Builders/LinkBuilder.cs
```diff
@@ -48,16 +53,16 @@ public TopLevelLinks GetTopLevelLinks(ResourceContext primaryResource)
...
         {
             if (primaryResource.TopLevelLinks != Link.NotConfigured)
+            {
                 return primaryResource.TopLevelLinks.HasFlag(link);
+            }
+
             return _options.TopLevelLinks.HasFlag(link);
         }
 
-        private void SetPageLinks(ResourceContext primaryResource, ref TopLevelLinks links)
```
wisepotato: why this change?

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/PagingTests.cs
```diff
@@ -88,6 +89,69 @@ public async Task Can_Paginate_TodoItems_From_Start()
             Assert.Equal(expectedTodoItems, deserializedBody, new IdComparer<TodoItem>());
         }
 
+
+        [Fact]
+        public async Task Pagination_FirstPage_DisplaysCorrectLinks()
```
wisepotato: minor: I would  say

`Pagination_OnFirstPage_PageShowsNextAndLastLinks`

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/PagingTests.cs
```diff
@@ -88,6 +89,69 @@ public async Task Can_Paginate_TodoItems_From_Start()
...
+
+            // Act
+            var response = await Client.GetAsync(route);
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            var body = await response.Content.ReadAsStringAsync();
+            var links = JsonConvert.DeserializeObject<Document>(body).Links;
+            Assert.EndsWith("/api/v1/todoItems?page[size]=5&page[number]=1", links.Self);
+            Assert.Null(links.First);
```
wisepotato: i mean with the `EndsWith` you are implicitely checking for not-null ness right? Seems superfluous

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/PagingTests.cs
```diff
@@ -88,6 +89,69 @@ public async Task Can_Paginate_TodoItems_From_Start()
...
+            Assert.Null(links.Prev);
+            Assert.EndsWith("/api/v1/todoItems?page[size]=5&page[number]=2", links.Next);
+            Assert.EndsWith("/api/v1/todoItems?page[size]=5&page[number]=4", links.Last);
+        }
+
+        [Fact]
+        public async Task Pagination_SecondPage_DisplaysCorrectLinks()
```
wisepotato: `Pagination_OnSecondPage_PageShowsNextAndLastLinks`

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/PagingTests.cs
```diff
@@ -88,6 +89,69 @@ public async Task Can_Paginate_TodoItems_From_Start()
...
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            var body = await response.Content.ReadAsStringAsync();
+            var links = JsonConvert.DeserializeObject<Document>(body).Links;
+            Assert.EndsWith("/api/v1/todoItems?page[size]=5&page[number]=2", links.Self);
+            Assert.EndsWith("/api/v1/todoItems?page[size]=5&page[number]=1", links.First);
+            Assert.EndsWith("/api/v1/todoItems?page[size]=5&page[number]=1", links.Prev);
+            Assert.EndsWith("/api/v1/todoItems?page[size]=5&page[number]=3", links.Next);
+            Assert.EndsWith("/api/v1/todoItems?page[size]=5&page[number]=4", links.Last);
+        }
```
wisepotato: We now have `page=1,2` i would suggest `n-1` and `n` as tests. Can we maybe use a `Theory` and `InlineData`?

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/PagingTests.cs
maurei: It is not just testing first and last links. Changed to more descriptive top-level links

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/PagingTests.cs
maurei: I'm not sure what you're pointing at. The test asserts that `links.Self` ends with the indicated string, and `links.First` is null. There is no overlap there

src/JsonApiDotNetCore/Serialization/Server/Builders/LinkBuilder.cs
maurei: instantiation of `TopLevelLinks` no longer happens inside the `SetPageLinks` helper function but before that, i.e. we can just pass the variable instead of the reference to that variable

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/PagingTests.cs
maurei: It is not just testing first and last links. Changed to more descriptive "top-level links"

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/PagingTests.cs
maurei: [backward paging was already tested](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/19c8aa43227707c0e86293fc1a3c604fdc096077/test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/PagingTests.cs#L156). Will merge them together

Issue #627, opened 2019-11-21
Support for query parameters on nested/deep resource endpoints


bart-degreed:
## Description
For example, the URL:

`https://localhost:5001/api/users/123/orders?fields[orders]=purchase-date`

returns all attributes of orders, instead of only purchase-date.

maurei:
I think the malfunctioning of paging, inclusion and sparse field (and probably the rest of the query parameters too) for nested resources have a common origin. Thanks for pointing this out

maurei:
I looked into this and unfortunately this is all pretty complicated

### Include
This was easy to support and is addressed in #634 

### Filter
Doing a filter on nested resource routes essentially the same as performing a [EF Core filtered include which is not possible](https://github.com/aspnet/EntityFrameworkCore/issues/1833). Eg the following request
```
GET /articles/1/tags?filter[name]=jsonapi
```
would be translated to the currently not supported EF Core invocation
```c#
dbContext.People.Include(e => e.Tags.Where(t => t.Name.Contains("jsonapi")));
```

### Paging

I suspect this is also not possible, see the following passage ([source](https://stackoverflow.com/a/32646709/4441216))
> Currently, in EF you cannot filter or limit the Included related entities in any way.

Note that this is about EF, but if it wasn't supported in EF that is a good indicator that it isn't supported in EF Core either. I need to research this further to be sure, but even if it turns out that EF Core works smoothly with it, the same issue as with sparse field selection and sorting will be encountered (next paragraph).

### Field selection, sorting
I think it is feasible to build expression trees that perform a sql `SELECT` and `ORDER` on the dataset, but this is currently not straight forward to implement due to design choice in JADNC. The problem is that the current expression tree building as performed in the `IResourceRepository<Article>` assumes that the table for which to perform a `SELECT`/`ORDER` is the "base" table of the query. This will not work for nested resource endpoints because it will always involve an `INNER JOIN`.

We could go ahead and change the expression tree building to be more flexible and support this. But from a separation-of-concerns point of view, I don't think having the `IResourceService<Article>` call `Select` and `Sort` methods on `IResourceRepository<Article>` to perform a select on or sort by operation for fields that are defined on `Tags` is a good way forward. To properly implement this we need to seriously detangle this.

#### Temporary work around: in-memory
[...]

PR #634, opened 2019-11-22
Support query param operations on nested resources

### LINKED ISSUE #628: Bug: Fetch include in deep URL passes query validation, but is not applied
## Description
For example, the URL:

`https://localhost:5001/api/users/123/orders?include=orders.country`

returns all attributes of orders, but it does add an `included` section that contains the related country.

maurei:
This closed  #628 and partially addresses  #627

Please see #627 for a the relevant discussion about this

- [x] Support includes on nested route
~- [ ] Support filter on nested route~
~- [ ] Support sparse field selection on nested route~
~- [ ] Support paging on nested route~
- [x] Throw HTTP 400 when unsupported query params are used in nested resource routes

### DIFF
--- a/src/JsonApiDotNetCore/Services/DefaultResourceService.cs
+++ b/src/JsonApiDotNetCore/Services/DefaultResourceService.cs
@@ -274,7 +274,10 @@ namespace JsonApiDotNetCore.Services
             {
                 foreach (var inclusionChain in chains)
                 {
-                    inclusionChain.InsertRange(0, chainPrefix);
+                    if (chainPrefix != null)
+                    {
+                        inclusionChain.InsertRange(0, chainPrefix);
+                    }
                     entities = _repository.Include(entities, inclusionChain.ToArray());
                 }
             }
--- a/src/JsonApiDotNetCore/Services/DefaultResourceService.cs
+++ b/src/JsonApiDotNetCore/Services/DefaultResourceService.cs
@@ -80,7 +80,7 @@ namespace JsonApiDotNetCore.Services
             if (!IsNull(_hookExecutor, entity)) _hookExecutor.AfterDelete(AsList(entity), ResourcePipeline.Delete, succeeded);
             return succeeded;
         }
-        g
+        
         public virtual async Task<IEnumerable<TResource>> GetAsync()
         {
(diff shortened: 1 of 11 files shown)


src/JsonApiDotNetCore/Services/DefaultResourceService.cs
```diff
@@ -251,12 +256,34 @@ protected virtual IQueryable<TResource> ApplyFilter(IQueryable<TResource> entiti
         /// </summary>
         /// <param name="entities"></param>
         /// <returns></returns>
-        protected virtual IQueryable<TResource> ApplyInclude(IQueryable<TResource> entities)
+        protected virtual IQueryable<TResource> ApplyInclude(IQueryable<TResource> entities, IEnumerable<RelationshipAttribute> chainPrefix = null)
```
wisepotato: you deliberately pass a list, but you expect an `IEnumerable`, why not just pass an array then?

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/NestedResourceTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        [Theory]
+        [InlineData("filter[ordinal]=1")]
+        [InlineData("fields=ordinal")]
+        [InlineData("sort=ordinal")]
+        [InlineData("page[number]=1")]
+        [InlineData("page[size]=10")]
+        public async Task NestedResourceRoute_RequestWithUnsupportedQueryParam_ReturnsBadRequest(string queryParam)
+        {
+            // Act
+            var (body, response) = await Get($"/api/v1/people/1/assignedTodoItems?{queryParam}");
```
wisepotato: you could split this into an // Arrange part that prepares the task,

maurei: There isn't really a arrange part here

src/JsonApiDotNetCore/Services/DefaultResourceService.cs
maurei: Because of extensibility: if a developer extends the repository class it will be able to call the method with any list-like type. We're not following that approach everywhere yet, but we should be doing that I guess

PR #636, opened 2019-11-27
Return 404 or 200 on error/non existent resource

### LINKED ISSUE #620: NotFoundResult throws NullReferenceException
## Description
When an id is not found on a `GetAsync(id)` request, the controller properly returns a `NotFoundResult`. However, this seems to break after that and what is actually returned to the client is a `NullReferenceException`.

Most concerning from a security standpoint is that _**the stack trace is returned.**_

> The most common problem is when detailed internal error messages such as stack traces, database dumps, and error codes are displayed to the user (hacker). These messages reveal implementation details that should never be revealed.

Reference: [Improper Error Handling](https://www.owasp.org/index.php/Improper_Error_Handling)

I updated my [sample repo](https://github.com/crgolden/JsonApiDotNetCoreIssue) with a simple integration test (`NotFound`) that reproduces this behavior. Please advise if I am missing something.
...

## Environment

- Other Relevant Package Versions: not sure

### LINKED ISSUE #631: Breaking change: null data returned when resource not found
## Description
Earlier, when a resource does not exist, a 404 with an empty response body was returned. After fetch of latest master (with v4 merged in), I am now getting 404 with the next body:

Request URL: https://localhost/api/users/123

Response body:
```json
{
  "links": {
    "self": "https://localhost/api/users"
  },
  "data": null
}
```

Note the `self` link is wrong too: it misses the identity.

Traced this down to https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/a801d958d9640365f459e096b802ad5d90db33fa/src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs#L108, so it may be intentional.

wisepotato:
Closes #631

## Problem
Upon getting to a resource that is non-existent, the following object is returned:
```json
{
  "links": {
    "self": "https://localhost/api/users"
  },
  "data": null
}
```
## Proposed solution
When a `200` is applicable, i.e. the route exists, but nothing can be returned, i.e. when a user with id `1` exists and a GET request is sent to `http://example.com/users/1/books`. But it should return a 404 if user `1` does not exist, also for `http://example.com/users/1`.

We also require a descriptive error message, instead of just returning 404.

## Done

- [x] Requesting an existing base model with no to-one relationships should return `200` with `null` for data. ( GET `/api/v1/todoItems/412/oneToOnePerson`)
- [] Requesting an existing base model's relationships with a to-many relationship should return `200` with `[]` for data ( GET `/api/v1/todoItems/1/stakeHolders`)
- [ ] Requesting an existing base model with no to-one relationships should return `200` with `null` for data. ( GET `/api/v1/todoItems/412/relationships/oneToOnePerson`)
- [ ] Requesting an existing base model's relationships with a to-many relationship should return `200` with `[]` for data ( GET `/api/v1/todoItems/1/relationships/stakeHolders`)
- [x] Requesting an non-existing base model (todoItem) should return 404

### DIFF
--- a/src/JsonApiDotNetCore/Internal/Exceptions/ResourceNotFoundException.cs
+++ b/src/JsonApiDotNetCore/Internal/Exceptions/ResourceNotFoundException.cs
@@ -1,0 +1,13 @@
+using System;
+
+namespace JsonApiDotNetCore.Internal
+{
+    public class ResourceNotFoundException : Exception
+    {
+        private readonly ErrorCollection _errors = new ErrorCollection();
+
+        public ResourceNotFoundException()
+        { }
+
+    }
+}
--- a/src/JsonApiDotNetCore/Middleware/DefaultExceptionFilter.cs
+++ b/src/JsonApiDotNetCore/Middleware/DefaultExceptionFilter.cs
@@ -1,7 +1,9 @@
-﻿using JsonApiDotNetCore.Internal;
+using JsonApiDotNetCore.Internal;
+using JsonApiDotNetCore.Managers.Contracts;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.Extensions.Logging;
+using System.Collections.Generic;
 
 namespace JsonApiDotNetCore.Middleware
 {
@@ -10,19 +12,19 @@ namespace JsonApiDotNetCore.Middleware
     /// </summary>
     public class DefaultExceptionFilter : ActionFilterAttribute, IExceptionFilter
     {
+        private readonly ICurrentRequest _currentRequest;
         private readonly ILogger _logger;
 
-        public DefaultExceptionFilter(ILoggerFactory loggerFactory)
+        public DefaultExceptionFilter(ILoggerFactory loggerFactory, ICurrentRequest currentRequest)
         {
+            _currentRequest = currentRequest;
             _logger = loggerFactory.CreateLogger<DefaultExceptionFilter>();
         }
 
         public void OnException(ExceptionContext context)
         {
             _logger?.LogError(new EventId(), context.Exception, "An unhandled exception occurred during the request");
-
             var jsonApiException = JsonApiExceptionFactory.GetException(context.Exception);
-
             var error = jsonApiException.GetError();
             var result = new ObjectResult(error)
             {
--- a/src/JsonApiDotNetCore/Serialization/Server/Builders/LinkBuilder.cs
+++ b/src/JsonApiDotNetCore/Serialization/Server/Builders/LinkBuilder.cs
@@ -110,6 +110,10 @@ namespace JsonApiDotNetCore.Serialization.Server.Builders
         /// <inheritdoc/>
         public RelationshipLinks GetRelationshipLinks(RelationshipAttribute relationship, IIdentifiable parent)
         {
+            if(parent == null)
+            {
+                return null;
+            }
             var parentResourceContext = _provider.GetResourceContext(parent.GetType());
             var childNavigation = relationship.PublicRelationshipName;
             RelationshipLinks links = null;
--- a/src/JsonApiDotNetCore/Services/DefaultResourceService.cs
+++ b/src/JsonApiDotNetCore/Services/DefaultResourceService.cs
@@ -125,7 +125,7 @@ namespace JsonApiDotNetCore.Services
         }
 
         // triggered by GET /articles/1/relationships/{relationshipName}
-        public virtual async Task<TResource> GetRelationshipsAsync(TId id, string relationshipName)
+        public virtual async Task<(TResource model, bool emptyResults)> GetRelationshipsAsync(TId id, string relationshipName)
         {
             var relationship = GetRelationship(relationshipName);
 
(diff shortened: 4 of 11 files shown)


wisepotato:
@bart-degreed  anything you're missing?

src/JsonApiDotNetCore/Internal/Exceptions/ResourceNotFoundException.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace JsonApiDotNetCore.Internal
+{
+    public class ResourceNotFoundException : Exception
```
bart-degreed: This type appears to be unused and can be removed.

src/JsonApiDotNetCore/Middleware/DefaultExceptionFilter.cs
```diff
@@ -10,19 +12,19 @@ namespace JsonApiDotNetCore.Middleware
     /// </summary>
     public class DefaultExceptionFilter : ActionFilterAttribute, IExceptionFilter
     {
+        private readonly ICurrentRequest _currentRequest;
         private readonly ILogger _logger;
 
-        public DefaultExceptionFilter(ILoggerFactory loggerFactory)
+        public DefaultExceptionFilter(ILoggerFactory loggerFactory, ICurrentRequest currentRequest)
```
bart-degreed: The private field that is assigned from this parameter is never used.

src/JsonApiDotNetCore/Serialization/Server/Builders/LinkBuilder.cs
```diff
@@ -110,6 +110,10 @@ public ResourceLinks GetResourceLinks(string resourceName, string id)
         /// <inheritdoc/>
         public RelationshipLinks GetRelationshipLinks(RelationshipAttribute relationship, IIdentifiable parent)
         {
+            if(parent == null)
```
bart-degreed: I think there should be a space between the `if` keyword and the opening parenthesis.

src/JsonApiDotNetCore/Services/DefaultResourceService.cs
```diff
@@ -135,15 +137,36 @@ public virtual async Task<TResource> GetRelationshipsAsync(TId id, string relati
...
+            var baseQuery = _repository.Get(id);
+            var entityQuery = ApplyInclude(baseQuery, chainPrefix: new List<RelationshipAttribute> { relationship });
+
             var entity = await _repository.FirstOrDefaultAsync(entityQuery);
-            if (entity == null)
+
+
+            // lol
+            var relationshipValue = typeof(TResource).GetProperty(relationship.InternalRelationshipName).GetValue(entity) ;
+            var relEmpty = relationshipValue == null;
```
bart-degreed: Unused variable

src/JsonApiDotNetCore/Services/DefaultResourceService.cs
```diff
@@ -135,15 +137,36 @@ public virtual async Task<TResource> GetRelationshipsAsync(TId id, string relati
...
 
-            var entityQuery = ApplyInclude(_repository.Get(id), chainPrefix: new List<RelationshipAttribute> { relationship });
+            var baseQuery = _repository.Get(id);
+            var entityQuery = ApplyInclude(baseQuery, chainPrefix: new List<RelationshipAttribute> { relationship });
+
             var entity = await _repository.FirstOrDefaultAsync(entityQuery);
-            if (entity == null)
+
+
+            // lol
```
bart-degreed: Unclear comment

src/JsonApiDotNetCore/Services/DefaultResourceService.cs
```diff
@@ -135,15 +137,36 @@ public virtual async Task<TResource> GetRelationshipsAsync(TId id, string relati
...
+
+            if(relationshipValue == null)
             {
-                /// TODO: this does not make sense. If the **parent** entity is not found, this error is thrown?
-                /// this error should be thrown when the relationship is not found.
-                throw new JsonApiException(404, $"Relationship '{relationshipName}' not found.");
+                return null;
+            }
+            var listCast = (IList) relationshipValue;
+            if(listCast != null)
```
bart-degreed: Expression is always true. And should test for `IEnumerable` instead of `IList`.

src/JsonApiDotNetCore/Services/DefaultResourceService.cs
```diff
@@ -135,15 +137,36 @@ public virtual async Task<TResource> GetRelationshipsAsync(TId id, string relati
...
+            {
+                if(listCast.Count == 0)
+                {
+                    return null;
+                }
             }
 
+            //if (entity == null)
```
bart-degreed: Why was this block commented out? Should it be removed instead?

bart-degreed:
I did some more testing today and found that this PR throws a `NullReferenceException` on many-to-many relationships in `DefaultResourceService` at the next line:
```csharp
var relationshipValue = typeof(TResource).GetProperty(relationship.InternalRelationshipName).GetValue(entity) ;
```

in the debugger, `relationship.InternalRelationshipName` has value `JsonTags` (which is the json-exposed property, not what is backed by an EF Core entity. My model looks like this:
```csharp
public sealed class Pathway : Identifiable
{
    [NotMapped]
    [HasManyThrough("tags", nameof(Tags))]
    public List<Tag> JsonTags { get; set; }

    public List<PathwayTag> Tags { get; set; }
}
```
where `PathwayTag` is the join-table and `Tag` the exposed resource. Used URL: 
```
https://localhost/pathways/ABC/tags
```
So I tried to fix this by adding this:
```csharp
            object relationshipValue;
            if (relationship is HasManyThroughAttribute hasManyThroughRelationship)
            {
                relationshipValue = hasManyThroughRelationship.ThroughProperty.GetValue(entity);
            }
            else
            {
                relationshipValue = typeof(TResource).GetProperty(relationship.InternalRelationshipName).GetValue(entity);
            }
```

this works, unless there is no match found. In that case, this method returns `null`, which causes another `NullReferenceException` raised at the last line of: 
```csharp
        // triggered by GET /articles/1/{relationshipName}
        public virtual async Task<object> GetRelationshipAsync(TId id, string relationshipName)
        {
            var relationship = GetRelationship(relationshipName);
            var resource = await GetRelationshipsAsync(id, relationshipName);
            return relationship.GetValue(resource);
        }
```
because `resource` is `null`, so getting the relationship value fails.

bart-degreed:
Hi @wisepotato I believe I fixed all issues as part of #714. See the tests in FetchingRelationshipsTests and the last two of UpdatingRelationshipsTests. Would you be okay if we close this PR? If I've missed anything, please let me know.

PR #662, opened 2020-01-06
Attr property info

### LINKED ISSUE #650: Feature request: publicly expose AttrAttribute.PropertyInfo getter
## Description
The resource graph builder performs traversal of relations to resolve the matching `PropertyInfo` instance for filters like `?filter[user.id]=1`. It follows the resource.User property and points to its .Id property. I need access to that property in order to implement deep searching using Elasticsearch. Aside from the duplication, it does not seem easy/possible to re-implement that myself.

Furthermore, once this is exposed, it looks like `InternalAttributeName` can be removed. Which is a good thing, according to this documentation:
```
        /// <summary>
        /// Do not use this overload in your applications.
        /// Provides a method for instantiating instances of `AttrAttribute` and specifying
        /// the internal property name.
        /// The primary intent for this was to enable certain types of unit tests to be possible.
        /// This overload will be deprecated and removed in future releases and an alternative
        /// for unit tests will be provided.
        /// </summary>
        public AttrAttribute(string publicName, string internalName, bool isImmutable = false)
        {
            PublicAttributeName = publicName;
            InternalAttributeName = internalName;
            IsImmutable = isImmutable;
        }
```
Existing unit tests can then use .PropertyInfo.Name instead.

Please let me know if there are any objections to this. I'd be happy to submit a PR once approved.

## Environment
latest master branch

bart-degreed:
Closes #650

#### FEATURE
- Made `AttrAttribute.PropertyInfo` public and removed `AttrAttribute.InternalAttributeName`.
- Removed unused interface `IRelationshipField`.

### DIFF
--- a/benchmarks/Query/QueryParser_Benchmarks.cs
+++ b/benchmarks/Query/QueryParser_Benchmarks.cs
@@ -25,7 +25,7 @@ namespace Benchmarks.Query {
             var requestMock = new Mock<IRequestContext>();
             requestMock.Setup(m => m.GetRequestResource()).Returns(new ResourceContext {
                 Attributes = new List<AttrAttribute> {
-                    new AttrAttribute(ATTRIBUTE, ATTRIBUTE)
+                    new AttrAttribute(ATTRIBUTE)
                 }
             });
             var options = new JsonApiOptions();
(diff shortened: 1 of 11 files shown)


bart-degreed:
@maurei  I was unable to run all tests locally, because I do not have pgsql installed. If they do not run automatically during PR build, someone should try them before merging this in.

PR #663, opened 2020-01-06
Fixed spelling errors

bart-degreed:
/cc @maurei


bart-degreed:
These were identified by [ReSpeller](https://www.jetbrains.com/help/resharper/Spell_Checking.html), a spell checking engine that ships with Resharper. 

I have gone over the list twice. Despite the large set of affected files, reviewing this PR shouldn't take more than 5 minutes. It would be great if one of the repro maintainers could take a look and approve these for merge.

PR #670, opened 2020-01-22
Removed unused property RelationshipAttribute.EntityPropertyName

### LINKED ISSUE #649: Unused relationship property
## Description
`RelationshipAttribute` has constructor parameter `string mappedBy`, which is assigned to property `EntityPropertyName`. That property is never used. This may be a bug. But if the property is unneeded, I think it should be removed (along with the parameter) for clarity.

## Environment
latest master branch

bart-degreed:
Fixes #649.

Related to this, I found that `HasManyThroughAttribute` has a constructor parameter named `mappedBy`, whose value is passed to the base class string parameter called `inverseNavigationProperty`. I wonder if that is correct (sounds like something else entirely), but will leave that up to reviewers to decide.

Also note that the `inverseNavigationProperty` parameter is missing in the constructor documentation parameter list.

### DIFF
--- a/src/JsonApiDotNetCore/Models/Annotation/HasManyAttribute.cs
+++ b/src/JsonApiDotNetCore/Models/Annotation/HasManyAttribute.cs
@@ -12,7 +12,6 @@ namespace JsonApiDotNetCore.Models
         /// <param name="publicName">The relationship name as exposed by the API</param>
         /// <param name="relationshipLinks">Which links are available. Defaults to <see cref="Link.All"/></param>
         /// <param name="canInclude">Whether or not this relationship can be included using the <c>?include=public-name</c> query string</param>
-        /// <param name="mappedBy">The name of the entity mapped property, defaults to null</param>
         /// 
         /// <example>
         /// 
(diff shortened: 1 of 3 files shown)


maurei:
About the missing doc for `inverseNavigationProperty`: I'll add that in a separate PR once this has been merged (it would have been easier for me to add that adjustment if you create a branch in this repo instead of your fork -- I'm not sure if that requires any rights you don't have?)

bart-degreed:
It is located at: https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/src/JsonApiDotNetCore/Models/Annotation/HasManyThroughAttribute.cs#L46-L47

In the image below, Resharper shows the parameter name (`inverseNavigationProperty`) with argument `mappedby` being passed for it.

image

I do not think I have permission to create a branch on this repo. But if you tell me what to change, I'll push it to this PR.

maurei:
That looks like some weird behaviour of ReSharper. In the link you provided, the `mappedBy` parameter is passed as the 4th parameter to the base constructor, which [is actually the base mappedBy parameter](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/7015b212899e815c52a0d279c83798b03b322b52/src/JsonApiDotNetCore/Models/Annotation/HasManyAttribute.cs#L28), not `inverseNavigationProperty` as displayed in your screenshot. Unless I'm completely missing something here?

I'm assuming it's alright and will proceed merging.

bart-degreed:
Sorry for the confusion. The link pointed to the master branch, while the screenshot came from the PR branch. In master, the base class have 5 parameters, but in the PR it has 4. This looks like a refactoring mistake from my side. Let me create another PR to fix that.

Issue #632, opened 2019-11-21
Self link not correct when requesting single resource


maurei:
see https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/631#issuecomment-557135308

PR #672, opened 2020-01-23
Fixed invalid top-level self link

### LINKED ISSUE #646: Bug: Invalid top-level self link
## Description
When requesting the url `https://localhost/api/v3/users/123`, a response is returned with top-level self link `https://localhost/api/v3/users`. 

From the JSON:API spec:
```
The top-level links object MAY contain the following members:
- self: the link that generated the current response document.
```

This may be related to #632, but in this case it involves an existing resource.

bart-degreed:
Fixes #646 and #632.

Removed dependency on `ResourceContext` in the process, because it was no longer needed.
Additionally I changed the decision to render which links to be taken from the related entity (from Article in a request like /people/1/articles instead of Person), because I believe that is more correct than taking it from the primary entity, which is the old behavior.


bart-degreed:
Resolved merge conflicts.

src/JsonApiDotNetCore/Serialization/Server/Builders/LinkBuilder.cs
```diff
@@ -27,17 +28,19 @@ public class LinkBuilder : ILinkBuilder
         }
 
         /// <inheritdoc/>
-        public TopLevelLinks GetTopLevelLinks(ResourceContext primaryResource)
+        public TopLevelLinks GetTopLevelLinks()
         {
+            ResourceContext resourceContext = _currentRequest.GetRequestResource();
```
maurei: The goal initially was to have the `GetTopLevelLinks`, `GetResourceLinks` and `GetRelationshipLinks` ask themselves for all the parameters it needs to generate the links for easy testability and clarity. But yep: clearly just having the `ResourceContext` of the primary resource will not be enough to distinguish between `/people/1` vs `people/1/articles` vs `/people` in the case of self links, so we might as just rely on the already injected CurrentRequest. Good job on being able to remove the dependency from the base serializers!

bart-degreed: Thanks! It looks like the Travis build has succeeded, but here on GitHub it still shows as pending.

PR #674, opened 2020-01-24
Cleanup of public overloads for `ILoggerFactory`

bart-degreed:
Ensured consistent passing along of `ILoggerFactory` parameter downwards to base classes and dependencies. I ended up making them all non-optional, because it is easy to forget passing it (compiler does not warn).
Callers can still pass `null` as a value, but that should be very rare, since ASP.NET Core includes built-in logging that is automatically registered.

In the process, I added some missing namespaces in examples (types were in global namespace).

`BaseJsonApiController<T>` and `BaseJsonApiController<T, TId>`:
- Made types abstract with protected constructors, fixed related tests
- Ensure all overloads have `ILoggerFactory` parameter and its value is passed downwards
- Have overloads call each other, so that fields assignment occurs only once

`JsonApiCmdController<T>` and `JsonApiCmdController<T, TId>`, `JsonApiQueryController<T>` and `JsonApiQueryController<T, TId>`:
- Ensure all overloads have `ILoggerFactory` parameter and its value is passed downwards
- Changed `IResourceService<>` parameter into `IResourceQueryService<>` / `IResourceCmdService` (still allows to pass `IResourceService<>`, because it implements both)
- Aligned parameter names with base class

`JsonApiController<T>` and `JsonApiController<T, TId>`:
- **Bugfix where loggerFactory was assigned the value `null` while passing it downwards!**
- Ensure all overloads have `ILoggerFactory` parameter and its value is passed downwards

`DefaultResourceRepository<TResource>` and `DefaultResourceRepository<TResource, TId>`:
- Removed unneeded constructor overload

`DefaultExceptionFilter`:
- Added null conditional operator, because at logging time the logger is assumed to be possibly `null`.

All unit tests succeed with these changes.


src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs
```diff
@@ -9,10 +9,9 @@
 
 namespace JsonApiDotNetCore.Controllers
 {
-    public class BaseJsonApiController<T, TId>
-        : JsonApiControllerMixin
-        where T : class, IIdentifiable<TId>
+    public abstract class BaseJsonApiController<T, TId> : JsonApiControllerMixin where T : class, IIdentifiable<TId>
```
maurei: Abstract class: 👍

src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs
```diff
@@ -22,63 +21,26 @@ public class BaseJsonApiController<T, TId>
...
-            _getById = resourceService;
-            _getRelationship = resourceService;
-            _getRelationships = resourceService;
-            _create = resourceService;
-            _update = resourceService;
-            _updateRelationships = resourceService;
-            _delete = resourceService;
+        protected BaseJsonApiController(IJsonApiOptions jsonApiOptions, ILoggerFactory loggerFactory,
+            IResourceService<T, TId> resourceService)
+            : this(jsonApiOptions, loggerFactory, resourceService, resourceService, resourceService, resourceService,
```
maurei: Calling sibling constructors: 👍

src/Examples/GettingStarted/Controllers/ArticlesController.cs
```diff
@@ -2,15 +2,16 @@
...
 using JsonApiDotNetCore.Services;
+using Microsoft.Extensions.Logging;
 
 namespace GettingStarted
 {
     public class ArticlesController : JsonApiController<Article>
     {
-        public ArticlesController(
-            IJsonApiOptions jsonApiOptions,
+        public ArticlesController(IJsonApiOptions jsonApiOptions, ILoggerFactory loggerFactory,
```
maurei: Somewhat nitpicking, but good be consistent with code formatting. The formatting the arguments has been altered. I don't really mind which formatting we use, but let's keep it consistent. I have been using  the visual studio `Wrap every parameter` `>` ` Indent all parameters` formatting option and suggest you use that. We could also use the `Wrap every parameter` `>` `Align wrapped parameters` options as it uses one less line.

Throughout the PR such formatting alternations appear. I won't be commenting on every but do a quick sweep if you can 👍 

(Note the formatting isn't consistent yet in the *entire* codebase, but I try to. Feel free to clean that up anywhere you spot it)

src/Examples/JsonApiDotNetCoreExample/Services/CustomArticleService.cs
```diff
@@ -14,13 +14,12 @@ namespace JsonApiDotNetCoreExample.Services
 {
     public class CustomArticleService : DefaultResourceService<Article>
     {
-        public CustomArticleService(IEnumerable<IQueryParameterService> queryParameters,
-                                    IJsonApiOptions options,
-                                    IResourceRepository<Article, int> repository,
-                                    IResourceContextProvider provider,
-                                    IResourceHookExecutor hookExecutor = null,
-                                    ILoggerFactory loggerFactory = null)
```
maurei: See comment about optional dependency in main thread

src/Examples/ReportsExample/Services/ReportService.cs
```diff
@@ -2,36 +2,40 @@
 using System.Threading.Tasks;
 using JsonApiDotNetCore.Services;
 using Microsoft.Extensions.Logging;
+using ReportsExample.Models;
 
-public class ReportService : IGetAllService<Report>
+namespace ReportsExample.Services
```
maurei: 👍

src/JsonApiDotNetCore/Controllers/JsonApiCmdController.cs
```diff
@@ -3,27 +3,26 @@
...
-        where T : class, IIdentifiable<int>
+    public class JsonApiCmdController<T> : JsonApiCmdController<T, int> where T : class, IIdentifiable<int>
     {
-        public JsonApiCmdController(
-            IJsonApiOptions jsonApiOptions,
-            IResourceService<T, int> resourceService)
-            : base(jsonApiOptions, resourceService)
-        { }
+        public JsonApiCmdController(IJsonApiOptions jsonApiOptions, ILoggerFactory loggerFactory,
+            IResourceCmdService<T, int> cmdService)
```
maurei: 👍

src/JsonApiDotNetCore/Controllers/JsonApiController.cs
```diff
@@ -9,27 +9,22 @@ namespace JsonApiDotNetCore.Controllers
...
     public class JsonApiController<T, TId> : BaseJsonApiController<T, TId> where T : class, IIdentifiable<TId>
     {
-        /// <param name="jsonApiOptions"></param>
-        /// <param name="resourceService"></param>
-        /// <param name="loggerFactory"></param>
-        public JsonApiController(
-            IJsonApiOptions jsonApiOptions,
-            IResourceService<T, TId> resourceService,
-            ILoggerFactory loggerFactory = null)
-            : base(jsonApiOptions, resourceService, loggerFactory = null)
```
maurei: Whoops, good one catching this unintended null assignment!

src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs
```diff
@@ -90,6 +52,10 @@ public class BaseJsonApiController<T, TId>
             _update = update;
             _updateRelationships = updateRelationships;
             _delete = delete;
+
+            _logger = loggerFactory != null
+                ? loggerFactory.CreateLogger<BaseJsonApiController<T, TId>>()
+                : new Logger<BaseJsonApiController<T, TId>>(new LoggerFactory());
```
maurei: If ILoggerFactory is a required parameter, I don't think it can ever be null because it is a framework service rather than an application service and as such will always be injected.

If we go for ILoggerFactory being an optional parameter, then if it is null it will be so for a reason (the dev choosed to) in which case we shouldn't ignore that by instantiating a logger anyway

Either way, I think we shouldn't go about this way. 

(Also, it is a violation of SRP, see related comment in main thread)

bart-degreed: Yes, I agree we should change this. Either choose to always require callers to pass a logger/factory, or make it optional and then conditionally log depending on whether caller passed `null`.

Note that even when required, the logging can be suppressed per namespace/type in `appsettings.json`.

src/Examples/GettingStarted/Controllers/ArticlesController.cs
bart-degreed: I totally agree about being consistent with formatting. While working through the codebase, I found various conflicting styles of all sorts so I did not worry about it too much. I'm not comfortable with changing VS settings all the time because I work on different repositories, but will do it for now temporarily and go over my changes.

Fortunately there are better ways to deal with uniform style. In our team within Degreed I have setup highly automated tooling to ensure normalized formatting using free tools. I'm willing to share, but that's a story for another time...

bart-degreed:
Thanks for your review. Let me respond to your primary concerns.

> 1. **Optional parameter**.
> backward compatibility...

I agree we should make logging optional. It allows for easier tests setup and easier migration from earlier versions where no logging parameter existed. And I am okay with letting the caller choose whether they want to use logging, although I think it would be more common to just have a logger with only debug output be injected when callers are not interested in logging. I do however think it should be a required parameter (allowing caller to pass null) at a preferably fixed position (I chose right after the `IJsonApiOptions` parameter). Because it is just too easy to forget passing the value through to all dependencies otherwise, which is actually what happened in this repo. This is a breaking change, but the migration is straightforward and it targets the next major version of JADNC so that should not be a problem.

> For those who don't need logging it only adds noise to the constructor of their controller by having to null it out.

I believe forcing callers to choose whether they want logging or not improves the maintainability of code. The alternative, an optional parameter somewhere at the end, makes it easy to miss. In my opinion, the improvement in code maintainability outweighs the comfort of not passing anything, in the long term. I believe it is good to be explicit about dependencies. Logging becomes important when issues arise, so it's good to have it by default. Reducing logging noise is much easier than to inspect JADNC internals to determine what logging is missing.

> 2. Usage of ILogger<T>

Logging frameworks traditionally have had a bit of runtime magic around them, in that they are expected to automatically know who they are working for (the class name). The LoggerFactory nicely abstracts that concern away, making it possible to reuse a single dependency throughout the call stack. Having a different logger dependency all the time makes it very hard to manage dependencies without D/I. 

For example: `FooService` internally uses a `BarService`. The constructor for `FooService` needs to accept a `Logger<FooService>` plus a `Logger<BarService>` in order to instantiate its dependency. Image the long list of typed loggers you'd need in a real complex application (noise). `LoggerFactory` solves this.

I think the SRP point is valid, but also a bit puristic.
The official docs show both logger and factory examples.

The only strong reason I have against using typed loggers is for class hierarchies. Example:

```csharp
class Base
{
  public Base(ILogger<Base> logger) { ... }
}

class Derived : Base
{
  public Derived(ILogger<Derived> logger, ILogger<Base> baseLogger) : base(baseLogger) { }
}

```
We need a separate logger for each type in the hierarchy to get the correct class names. Using class names to track down where execution is coming from is so common in my experience that it is important to get it right. And losing the ability to pass custom category names is actually a good thing. Under what category/class name JADNC logs should be fixed, determined by JADNC and not library users. This is diagnostic information that may be shared on stackoverflow, issues etc. so it does not help if different users are using different categories to log the JADNC internal behavior.

maurei:
> The only strong reason I have against using typed loggers is for class hierarchies

Agree. Perhaps stick with `ILogger` then and leave it to the users to either directly resolve it normally from DI and pass it on to the base, or if more flexibility is required allow users to construct an `ILogger` themselves from `ILoggerFactory` (or equivalent) and pass that on to the base?

maurei:
>  I do however think it should be a required parameter (allowing caller to pass null) at a preferably fixed position (I chose right after the IJsonApiOptions parameter)

Given how c# handles optional parameters (at the end with a default value), placing it in the middle without a default value and surrounded by non-optional params it will not suggest to users it is optional. Users would need to know about JADNC internals that passing `null` is equivalent to opting out on logging and will not cause a nullref exeception. I feel it will be more clear to users if we handle optional params in the standard dotnetcore/c# way

> Because it is just too easy to forget passing the value through to all dependencies otherwise, which is actually what happened in this repo.

JADNC not passing it through was an error but any internal measures taken to ensure consistency here shouldn't ripple through to the user exposed API (should be achieved tests instead)

bart-degreed:
> Agree. Perhaps stick with ILogger then

I tried that, it does not work. How can the D/I container select which logger to create or reuse if nothing is specified? Note that once a logger instance is created, it is bound to a single type/category. That's why you should never share logger instances between types, not even in a type hierarchy. You need to either pass an `ILogger<T>` or `ILoggerFactory`.

Working example:
```csharp
    public class PeopleController : JsonApiController<Person>
    {
        private readonly ILogger<PeopleController> _logger;

        public PeopleController(IJsonApiOptions jsonApiOptions, ILoggerFactory loggerFactory,
            IResourceService<Person> resourceService)
            : base(jsonApiOptions, loggerFactory, resourceService)
        {
            _logger = loggerFactory.CreateLogger<PeopleController>();

            // ...

            _logger.LogInformation("Message from PeopleController.");
        }
    }

    public class JsonApiController<T, TId> : BaseJsonApiController<T, TId> where T : class, IIdentifiable<TId>
    {
        private readonly ILogger<JsonApiController<T, TId>> _logger;

        public JsonApiController(IJsonApiOptions jsonApiOptions, ILoggerFactory loggerFactory,
            IResourceService<T, TId> resourceService)
            : base(jsonApiOptions, loggerFactory, resourceService)
        {
            _logger = loggerFactory.CreateLogger<JsonApiController<T, TId>>();

            // ...

            _logger.LogInformation("Message from JsonApiController.");
        }
    }
```
Are we in agreement to use `ILoggerFactory` instead of `ILogger<T>`?

bart-degreed:
> Given how c# handles optional parameters (at the end with a default value), placing it in the middle without a default value and surrounded by non-optional params it will not suggest to users it is optional.

True, but realize that explicitly passing `null` for logging is very unlikely to happen in an ASP.NET Web Application project. See my recent push to this PR, where logging appears to be injected automatically without any configuration in Startup. This is how netcore3 works, so enforcing that `null` is passed for logging requires special effort anyway.

> Users would need to know about JADNC internals that passing null is equivalent to opting out on logging and will not cause a nullref exeception.

Like any public API parameter, it should be documented whether it is required, what value range is acceptable, etc. Making it optional is for convenience in tests only. And even if that gets unnoticed, a test likely passes a `Mock<>` that always returns `null` and we have the same effect: no logging occurs, and no `NullReferenceException`.

> I feel it will be more clear to users if we handle optional params the was as intended by c#

Optional parameters in C# are quite controversial. They were added to the language only after 10 years, being pushed back by the language designers until then. Eventually they were added in an effort to improve COM interop with the Word/Excel object models in VB, where optional parameters were used extensively. Calling them from C# required to pass `Missing.Value` about 14 times to match the COM signature. Some coding guidelines forbid the use of optional parameters entirely.

> JADNC not passing it through was an error but any internal measures taken to ensure consistency here shouldn't ripple through to the user exposed API (should be achieved with tests instead)

This does not just apply to this repo, but to any source tree that derives from the types here, such as controllers, services and repositories. By making them optional, it is easily missed. For example:

```csharp
    public class PeopleController : JsonApiController<Person>
    {
        public PeopleController(
            IJsonApiOptions jsonApiOptions,
            IResourceService<Person> resourceService) 
            : base(jsonApiOptions, resourceService/* oops, forgotten about logging: loggerFactory*/)
        { }
    }
```

This compiles just fine, while the library user may not have intended to disable logging in the entire type hierarchy. Its better to get a compile error, and realize you need to pass `null` to fix that, than to never be notified when doing it wrong accidentally.

src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs
bart-degreed: Fixed in recent commit https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/674/commits/1a0ea3e5f52288b21d838c8be2c98378975c6719.

bart-degreed:
https://ericlippert.com/2011/05/09/optional-argument-corner-cases-part-one/

maurei:
AH, now I get it; the goal is to have unique logging categories per *inheritance level* within a *single instance*. A use-case for this would be easily differentiating between internal JADNC logs versus custom application logs? In this case `ILoggerFactory` would make more sense indeed, as it provides both the dev and JADNC with their responsibility of creating their respective loggers. Therefore, usage of `ILoggerFactory`: ✅ 


About the optional parameter: I realised my approach will lead to inconsistency. With out-of-the-box usage of JADNC, the dev never touches the constructors of services/repositoriesand since (like you mentioned) the logging services are always available, these parameters will indeed *always* be populated and as such there is no reason for them to be marked as optional there. The only way to disable logging there is to set the loglevel to none, which I guess is also the recommended way in netcore3. Then for consistency the same approach for disabling logging should be pursued in other layers of the framework, i.e also the `JsonApiController`. Hence, non-optional argument: ✅

src/Examples/GettingStarted/Controllers/ArticlesController.cs
maurei: For now perhaps restore it to the previous state, and let us try to keep that as an example for classes throughout the code base?

maurei:
Probably we should extend this PR to also cover other places (services, repositories) where `ILoggerFactory` is referenced  and make the argument non-optional

bart-degreed:
> Ah now I get it; the goal is to have unique logging categories per...

Yes, exactly. The logging strategy (names, levels, stack traces, etc) from JADNC should be independent from the application logging strategy.

> About the optional parameter: I realised my approach...

Yes, having `null` loggers is the exception rather than the common scenario. Another way to deactivate logging would be to detach all providers.

Today I discovered [NullLoggerFactory.Instance](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.abstractions.nullloggerfactory?view=dotnet-plat-ext-3.1), exists, which implements the [null object design pattern](http://software-pattern.org/Null-Object). 

This makes me reconsider whether `ILoggerFactory` would be better off being a required parameter after all. See also [this guidance](https://blog.rsuter.com/logging-with-ilogger-recommendations-and-best-practices/#testing).

So we agree that `ILoggerFactory` should be a non-optional parameter, but passing `null` is allowed. While I can live with that, would you be okay to make it a required parameter instead?

> Probably we should extend this PR...

We have been discussing controllers, but this PR has targeted all usages of `ILoggerFactory` in the codebase from the start.

bart-degreed:
I reverted formatting changes. Assuming you agree with making `ILoggerFactory` a required parameter, this PR is ready for merge.

PR #675, opened 2020-01-24
Fixed all compiler warnings

bart-degreed:
Fixed all compiler warnings, except:
```
CS1591: Missing XML comment for publicly visible type or member
```
because it is a massive amount of work to fix all of those.

So I set builds to fail on compiler warnings in release mode, but with a suppression rule for CS1591. The suppression can be removed once all documentation has been updated. Until then, this should help to protect against accidentally introducing new build warnings.

I also corrected broken `RelationshipAttribute.Equals` implementation.


bart-degreed:
Resolved merge conflicts.

maurei:
Thanks for this. I myself have tried to use StyleCop for linting to prevent many of these warnings from occuring in the first place, but unfortunately it does not work properly yet with Visual Studio for Mac, [see this issue](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3058).

bart-degreed:
The reason I set warning-as-error for release builds only is to ensure warnings are resolved before commit to master, but do not get in your way when writing, refactoring and debugging code. For example, commenting out code or adding a temporary variable for inspection often causes compiler warnings like "unused variable" and "unreachable code", which is annoying. If anyone wants to prevent a failing CI build, they should check the warnings window in VS is empty before push.

For the same reason, I don't like StyleCop. It bothers me all the time with syntax-violation squiggles and I find that very distracting. When I'm ready to optimize and cleanup, I'll deal with the styling issues. Like I mentioned elsewhere, there are better tools out there to ensure a common style, while not getting in your way.

maurei:
Good point, turning it on for debugging would make it hard to iterate quickly during dev, I agree with that 👍.  I haven't come across any other projects in my career though for which this option was enabled, and also this project hasn't had it enabled in the past. As such I've gone ahead to add a section about this in the README for anyone wanting to get (back) into JADNC development.

PR #677, opened 2020-02-05
Repaired the broken benchmarks project

bart-degreed:
Removed `SortService.CheckIfProcessed()`, which was supposed to disallow multiple sort parameters. But it never worked. For example, if you pass `http://host/path?sort=a&sort=b`, it gets translated by `Microsoft.AspNetCore.Http.QueryCollection` into value "a,b". The check prevented to run benchmarks multiple times.

Also reactivated the build flag on all combinations of Platform/Configuration in Solution Configuration Manager for all projects.

Removed some benchmarks, for which no equivalent code exists anymore. But kept the benchmark on namespace-from-path extraction (and optimized a bit), because it nicely shows the perf difference between using strings and `ReadOnlySpan<char>`.

Also deleted old output markdown files from source control, as they differ per machine.

Finally, the stats from my laptop:

|                  Method |     Mean |     Error |    StdDev |
|------------------------ |---------:|----------:|----------:|
| DeserializeSimpleObject | 7.684 us | 0.0692 us | 0.0578 us |

|                Method |     Mean |     Error |    StdDev |
|---------------------- |---------:|----------:|----------:|
| SerializeSimpleObject | 7.871 us | 0.1011 us | 0.0946 us |

|         Method |         Mean |      Error |     StdDev |       Median |     Gen 0 |    Gen 1 | Gen 2 |  Allocated |
|--------------- |-------------:|-----------:|-----------:|-------------:|----------:|---------:|------:|-----------:|
|  AscendingSort |     2.680 us |  0.0211 us |  0.0449 us |     2.663 us |    1.0185 |   0.0076 |     - |     4.7 KB |
| DescendingSort |     2.728 us |  0.0258 us |  0.0527 us |     2.739 us |    1.0262 |   0.0076 |     - |    4.73 KB |
|   ComplexQuery | 3,258.409 us | 25.9838 us | 54.8087 us | 3,230.178 us | 1148.4375 | 281.2500 |     - | 5308.28 KB |

|            Method |      Mean |    Error |   StdDev |    Median |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------ |----------:|---------:|---------:|----------:|-------:|------:|------:|----------:|
|  UsingStringSplit | 244.66 ns | 0.783 ns | 1.702 ns | 244.60 ns | 0.1493 |     - |     - |     704 B |
| UsingReadOnlySpan |  69.72 ns | 2.638 ns | 5.680 ns |  66.09 ns | 0.0288 |     - |     - |     136 B |


maurei:
Thanks a lot for this! I never got into figuring out how the benchmarking works and never got to fixing it. I'm sure there is still lots to elaborate on in this field, that's something we can work on in the future.

PR #681, opened 2020-02-12
Added MaximumPageSize and MaximumPageNumber to IJsonApiOptions

### LINKED ISSUE #673: Feature request: max page size and number
## Description
For our current API to adopt JADNC (using database tables that contain millions of rows), we need the capability to constrain the maximum page size. Combined with client-based rate limiting, this protects us from callers issuing requests for potentially enormous responses, resulting in timeouts and/or database slowdowns on our side.

And given there are performance issues with paging through large sets ([SQL Server](https://sqlperformance.com/2015/01/t-sql-queries/pagination-with-offset-fetch), [Elasticsearch](https://rimdev.io/elasticsearch-sinking-in-deep-paging-quicksand/)), we need to limit the maximum value for page number too.

To support these, I propose to add the next properties to `IJsonApiOptions`:
- int? MaximumPageSize { get; }
- int? MaximumPageNumber { get; }

This enabled JADNC implementers to optionally set these in `Startup`, and they will be applied to all resources. An exception in thrown (resulting in 400 Bad Request) when a higher value is specified in the query string.

I'm willing to start work on a PR to implement this. Please let me know if this fits with the project vision. I've seen #217, which proposes a maximum page size per resource. That may still be implemented at a later time too, as a more fine-grained way to control this. But I think having a fallback default like proposed here does not conflict with the earlier proposal.

bart-degreed:
Fixes #673.

### DIFF
--- a/src/JsonApiDotNetCore/Configuration/IJsonApiOptions.cs
+++ b/src/JsonApiDotNetCore/Configuration/IJsonApiOptions.cs
@@ -19,6 +19,8 @@ namespace JsonApiDotNetCore.Configuration
         /// </example>
         bool IncludeTotalRecordCount { get; set; }
         int DefaultPageSize { get; }
+        int? MaximumPageSize { get; }
+        int? MaximumPageNumber { get; }
         bool ValidateModelState { get; }
         bool AllowClientGeneratedIds { get; }
         bool AllowCustomQueryParameters { get; set; }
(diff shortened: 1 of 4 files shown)


maurei:
Thats indeed pretty easy.  Thanks.

For some reason I was anticipating work on pagination of included (to-many) relationships, but I'm not sure why as that was never mentioned. I suspect though this is might still be relevant for you guys if ever deal with very large relationship collections being included.

PR #683, opened 2020-02-12
Updated build scripts to use auto-discovery

bart-degreed:
This ensures all projects in the solution are built in release mode and all unit tests in all projects in the solution are run.

Also removed outdated/broken additional solution files.


bart-degreed:
@maurei The failing AppVeyor builds puzzles me. Thoughts?

maurei:
Looks like you've figured it out. Let me know if this is the final state of the PR

bart-degreed:
I haven't been able to find the root cause of the failures (which worries me), but running the tests in a fixed order (workaround) seems to work. So this is ready for merge.

PR #685, opened 2020-02-13
Renamed Cmd to Command in class names

bart-degreed:
From the start I found it strange to have Cmd (not Command) and Query (not Qry) next to each other. So I renamed Cmd to the full version for clarity and consistency.


maurei:
That bothered me too, I agree. Thanks.

PR #688, opened 2020-02-14
Bugfix: Respect customizations on JsonSerializerSettings

### LINKED ISSUE #687: Bug: custom serializer settings are ignored
When setting custom serialization settings on `JsonApiOptions`, they are not used. In our scenario we have a few enums exposed as attributes and we want them to serialize into their text representation instead of numeric.

Example:
```csharp
public enum ScreenOrientation
{
    Horizontal, Vertical
}

public class DemoResource : Identifyable
{
    [Attr]
    public ScreenOrientation Orientation { get; set; }
}

public class Startup
{
    {
        services.AddJsonApi<AppDbContext>(options =>
        {
            options.SerializerSettings.Converters = new List<JsonConverter>
            {
                new StringEnumConverter()
            };
        });
    }
}
```

Putting `[JsonConverter(typeof(StringEnumConverter))]` on ScreenOrientation works, but the global setting does not.

Using the global setting should work, according to https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/docs/usage/options.md#custom-serializer-settings.

bart-degreed:
Bugfix: Respect customizations on JsonSerializerSettings when serializing response.

Fixes #687.

### DIFF
--- a/benchmarks/Serialization/JsonApiSerializerBenchmarks.cs
+++ b/benchmarks/Serialization/JsonApiSerializerBenchmarks.cs
@@ -1,5 +1,6 @@
 using System;
 using BenchmarkDotNet.Attributes;
+using JsonApiDotNetCore.Configuration;
 using JsonApiDotNetCore.Internal.Contracts;
 using JsonApiDotNetCore.Managers;
 using JsonApiDotNetCore.Query;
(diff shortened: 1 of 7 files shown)


bart-degreed:
I've come to realize that `JsonSerializerSettingsNullValueHandlingScope` was a very bad idea: Changing options on the shared instance will likely affect concurrent requests. As described [here](https://stackoverflow.com/questions/38230326/copy-jsonserializersettings-from-jsonserializer-to-new-jsonserializer), locking is not going to help.

The right way to deal with this is to have the settings copied into a new serializer instance. There is no Clone method, but calling `JsonSerializer.Create(settings)` does perform the copy internally. As an optimization, we may even cache the `JsonSerializer` instance (it is [thread-safe](https://stackoverflow.com/questions/36186276/is-the-json-net-jsonserializer-threadsafe)).

bart-degreed:
Replaced by #720.

PR #694, opened 2020-02-19
Fixed inefficient cache usage.

### LINKED ISSUE #450: EFCore based Memory Leak
## Description
There is a memory leak, based on the way EF Core cache queries, especially queries with parameters automatically generated with such a librairy.

The following test (on GetTests.cs) shows just that, with `cachedQueryCount` being equal to 1000 instead of 1.

```
        [Fact]
        public async Task Memory_Leak()
        {
            for (int i = 0; i < 1000; i++)
            {
                await _fixture.SendAsync("GET", $"/api/v1/courses?filter[number]:eq={i}", null);
            }

            int cachedQueryCount = 0;

            var cache = _fixture.Server.GetService<IMemoryCache>() as MemoryCache;
            var entriesProperty = typeof(MemoryCache).GetProperty("EntriesCollection", BindingFlags.NonPublic | BindingFlags.Instance);
            var entries = entriesProperty.GetValue(cache) as ICollection;
            var items = new List<string>();
            if (entries != null)
            {
                foreach (var item in entries)
                {
                    var methodInfoVal = item.GetType().GetProperty("Value");
                    var val = methodInfoVal.GetValue(item) as ICacheEntry;
                    if (val?.Value == null)
                    {
                        continue;
                    }
                    var contentType = val.Value.GetType();
                    var gens = contentType.GenericTypeArguments;
                    if (gens.Length == 2 && gens[0] == typeof(QueryContext) && gens[1] == typeof(IAsyncEnumerable<CourseEntity>))
                    {
                        cachedQueryCount++;
                    }
                }
            }

            Assert.Equal(1, cachedQueryCount); //Except it's 1000
        }
```

## Environment

bart-degreed:
By turning constant expressions (whose value comes from query string) into parameters, the parsed query can be reused for varying query string parameters.

Fixes #450.

### DIFF
--- a/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
+++ b/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
@@ -247,7 +247,7 @@ namespace JsonApiDotNetCore.Extensions
             PropertyInfo relationProperty = null;
             PropertyInfo property = null;
             MemberExpression left;
-            ConstantExpression right;
+            Expression right;
 
             // {model}
             var parameter = Expression.Parameter(concreteType, "model");
@@ -287,8 +287,8 @@ namespace JsonApiDotNetCore.Extensions
                     // convert the incoming value to the target value type
                     // "1" -> 1
                     var convertedValue = TypeHelper.ConvertType(filter.Value, property.PropertyType);
-                    // {1}
-                    right = Expression.Constant(convertedValue, property.PropertyType);
+
+                    right = CreateTupleAccessForConstantExpression(convertedValue, property.PropertyType);
                 }
 
                 var body = GetFilterExpressionLambda(left, right, filter.Operation);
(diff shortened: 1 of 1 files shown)


bart-degreed:
Additional locations where Expression.Constant is used:

- 'in' filter https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/668f071edf3093db8c0b4b17cd8c7ccb3b0b01aa/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs#L215, https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/668f071edf3093db8c0b4b17cd8c7ccb3b0b01aa/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs#L223 These operate on `IList`, which is translated to a SQL 'in' clause by EF Core. Although these seems to be a workaround at https://stackoverflow.com/questions/52055583/entity-framework-select-in-not-using-parameters, I feel this is better left to the EF Core team to optimize.

- 'isnull'/'isnotnull' filter https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/668f071edf3093db8c0b4b17cd8c7ccb3b0b01aa/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs#L284 Changing this makes some tests fail that try to pass `null` for a non-nullable ID. Having one additional cache entry for `null` should not be a big problem.

- ? https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/668f071edf3093db8c0b4b17cd8c7ccb3b0b01aa/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs#L389 I found no way to get that code executed. Maybe dead code?

PR #699, opened 2020-02-21
Updated deserialization to include error details and log at Info leve…

### LINKED ISSUE #655: Invalid Patch request body format required (most likely user error)
## Description
Either I am doing something incorrectly (likely), or there is an inconsistency with PATCH request required format (unlikely).  I also realize that I am operating with an alpha version, and as such things might be goofy.

I am having a few issues, but the salient issue for this post is that I can't get a PATCH request to result in a deserialized type with the format specified in the json:api spec 1.0

Expected format per spec:
`{
	"data": {
		"type": "users",
		"id": 10,
		"attributes": {
			"userType": "Employee"
		}
	}
}`
This format actually results in a null type being presented to the JsonApiController PatchAsync method.

Actual format required to get a deserialized type that contains the value presented in the userType property (:
`{
    "type": "users",
    "id": "10",
    "userType": "Employee"
}`

All other imagined variations of the request body format either result in null types or types with default values that do not reflect the values passed in the request body.
...

## Environment

- Targeted Version: .netcore3.0
- JsonApiDotNetCore Version: 4.0.0-alpha4
-- Microsoft.AspNetCore.Mvc.NewtonsoftJson: 3.1.0
-- Newtonsoft.Json: 12.0.3

bart-degreed:
…l instead or Error.

I suspect that adding an error to ModelState gets discarded if model state validation is turned off in settings, so I think it is better to throw, which results into a json-error-object.

Fixes #655.


bart-degreed:
I'm addressing this as part of a larger future PR that implements error handling differently.

PR #701, opened 2020-03-04
Eager loading

### LINKED ISSUE #652: How to include nested EF entities in include
## Description

I'm using AutoMapper as I expose through a `Resource` information that I take from several classes that require EF navigation. See `src.EntityTypeNameType.NameType.Code` in the mapping that I use..

```c#
 CreateMap<Model.EntityName, DTO.EntityName>()
    .ForMember(dest => dest.NameType,
        opts => opts.MapFrom(src => src.EntityTypeNameType.NameType.Code));
```

What happens is that with the `DefaultEntityRepository` I see only the base include is included unless I force it.

My idea is to override the `IQueryable<Entity> Include(...)` and create a Dictionary of the nested includes required that need to be loaded to query all the tables in the relationship.

This is more a design question or feature-request rather than a bug. Does it make sense the solution I'm planning?

btw, congrats for this project looks very nice! ;)

## Environment

- JsonApiDotNetCore Version: v3.1.0
- Other Relevant Package Versions: AutoMapper

bart-degreed:
Fixes #652.


bart-degreed:
This is ready for review. The code turned out a lot nicer than I expected it would!

src/JsonApiDotNetCore/Builders/ResourceGraphBuilder.cs
```diff
@@ -179,6 +179,39 @@ protected virtual List<RelationshipAttribute> GetRelationships(Type entityType)
         protected virtual Type GetRelationshipType(RelationshipAttribute relation, PropertyInfo prop) =>
             relation.IsHasMany ? prop.PropertyType.GetGenericArguments()[0] : prop.PropertyType;
 
+        private List<EagerLoadAttribute> GetEagerLoads(Type entityType, int recursionDepth = 0)
+        {
+            if (recursionDepth >= 500)
```
wisepotato: why 500? Is there a better way of detecting this?

bart-degreed: Just some high number that is unlikely to be ever hit, unless there are circular dependencies. This check makes the code throw instead of entering an endless loop.

bart-degreed: We could run a scan upfront that builds a dependency tree. But that is expensive. This way seems more appropriate to me to indicate the resource modelling is invalid. This is not something API-users will hit, but for the developer creating the model.

PR #704, opened 2020-03-06
Change tracking for patch updates, to improve json:api spec compliance

### LINKED ISSUE #577: Response when updating attributes of resource not conform spec
The [specs](https://jsonapi.org/format/#crud-updating-responses) for updating a resource state:

> 200 OK
> If a server accepts an update but also changes the resource(s) in ways other than those specified by the request (for example, updating the updated-at attribute or a computed sha), it MUST return a 200 OK response. The response document MUST include a representation of the updated resource(s) as if a GET request was made to the request URL.
> 
> A server MUST return a 200 OK status code if an update is successful, the client’s current attributes remain up to date, and the server responds only with top-level meta data. **In this case the server MUST NOT include a representation of the updated resource(s).**

JsonApiDotNetCore does not comply with the piece in **bold text**. Currently, after an update the response for the update pipeline **always** returns a full representation of the resource on the server. 

In the default implementation of the pipelines, how a resource will change will always be fully specified by the request, in which case the server **must never** return an updated representation.

Users can easily add business logic that introduces additional changes, and only in this case the server **must** return an updated representation. 

Implementing this part of the spec is challenging because it is nearly impossible for JADNC to detect if the business logic as introduced by the developer does anything like this. Hence it would probably be a more feasible approach to create a service that is responsible for signalling this to the internals of the framework. Such service would then explicitly be called by the developer.

Perhaps this issue can be tackled together with #536.
Eg we could introduce a helper service to take care of updates other than those specified by the request
```c#
updatedArticle.updatedAt = DateTime.Now();
_updateHelper.MarkAsUpdated(a => a.updatedAt);
```
the `_updateHelper` can then signal this to the repository layer which would solve #536, as well as to the serialization layer which would solve the current issue. This way we wont have to clutter the interface of the repository layer and could custom serialization layer opt out on such signal.

bart-degreed:
I have added some calculated properties to the examples, in order to have tests for:
- nothing outside the patch request changed (returns empty data)
  - `KebabCaseFormatterTests.KebabCaseFormatter_Update_IsUpdated` has no side effects in updating `KebabCasedModel` attributes
  - `ManyToManyTests.Can_Update_Many_To_Many` has no side effects in updating `Article` tags
  - `ManyToManyTests.Can_Update_Many_To_Many_With_Complete_Replacement` has no side effects in updating `Article` tags
  - `ManyToManyTests.Can_Update_Many_To_Many_With_Complete_Replacement_With_Overlap` has no side effects in updating `Article` tags
- exposed attribute that was not in PATCH request changed (returns attributes)
  - `UpdatingDataTests.PatchResource_ModelWithEntityFrameworkInheritance_IsPatched` updates `User.Password` property, which updates exposed `LastPasswordChange` attribute
  - `UpdatingDataTests.Patch_Entity_With_HasMany_Does_Not_Include_Relationships` updates `Person.FirstName` property, which updates exposed `Initials` attribute
  - `TodoItemsControllerTests.Can_Patch_TodoItemWithNullable` does not update exposed `TodoItem.AlwaysChangingValue` attribute
- exposed attribute that was in PATCH request changed (returns attributes)
  - `TodoItemsControllerTests.Can_Patch_TodoItem` updates `TodoItem.AlwaysChangingValue` attribute

Also updated a few places where an empty list was allocated each time with a cached empty array instance.

Fixes #577.


src/JsonApiDotNetCore/Services/DefaultResourceService.cs
```diff
@@ -183,25 +188,36 @@ public virtual async Task<object> GetRelationshipAsync(TId id, string relationsh
...
+            if (!IsNull(_hookExecutor, databaseEntity))
             {
-                _hookExecutor.AfterUpdate(AsList(entity), ResourcePipeline.Patch);
-                entity = _hookExecutor.OnReturn(AsList(entity), ResourcePipeline.Patch).SingleOrDefault();
+                _hookExecutor.AfterUpdate(AsList(databaseEntity), ResourcePipeline.Patch);
+                _hookExecutor.OnReturn(AsList(databaseEntity), ResourcePipeline.Patch);
             }
-            return entity;
+
+            _repository.FlushFromCache(databaseEntity);
```
maurei, line 215: Would there be any implementation beyond EF core for such repository method?

maurei: I also noticed that UpdateAsync does not return an entity anymore. Could we let the repo take care of detaching and re-getting the entity internally and returning it so that the service layer does not need to worry about it?

bart-degreed: > Would there be any implementation beyond EF core for such repository method?
I haven't checked, but expect NHibernate to provide something similar.

> Could we... so that the service layer does not need to worry about it?
From the service layer perspective: usually it is okay to use cached data, but in the change tracking flow there is an explicit need for a fresh fetch (for json:api compliance). An alternative would be to add a boolean flag "forceFresh" to all repository methods, which is more intrusive. The advantage of the extra method is it can just be left empty for non-caching implementations.

I'd prefer the keep the repository simple: just update the entity and save it. The change tracking is a json:api feature that is unrelated to the underlying data source, so I think the logic should be driven from the service layer. I've attempted to extract and delegate the logic for the most, but if you know of better ways I'd be interested.

PR #705, opened 2020-03-06
Add operators : Not Like,  Start With, Not Start with, End With, Not …

absymylyar:
I Added the operators Not Like,  Start With, Not Start with, End With, Not End With

### DIFF
--- a/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
+++ b/src/JsonApiDotNetCore/Extensions/IQueryableExtensions.cs
@@ -149,7 +149,28 @@ namespace JsonApiDotNetCore.Extensions
                     body = Expression.GreaterThanOrEqual(left, right);
                     break;
                 case FilterOperation.like:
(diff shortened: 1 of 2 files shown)


bart-degreed:
Nice addition! A few remarks:
- Do they all translate into SQL or do they execute client-side?
- I believe it would be better to spell out the operator names instead of the abbreviations for clarity
- Unit tests are missing
- Documentation needs to be updated

But I'm quite new on the project, so before putting a lot of work into this, I'd recommend to wait and see what the project owners feel about the new operators.

@maurei thoughts?

absymylyar:
I've been using those operators for 3 months now on my own project, angular 7 with DevExtrem Datagrid.
I can plug it directly to the API, translating the grid's filter to JsonApi format...

absymylyar:
I'm working also to use DateTime and TimeSpan as "HasOne" relationship, and make their properties (year, month, day, dayofweek, etc...) available for filtering and sorting.

bart-degreed:
Superseded by #792.

PR #714, opened 2020-04-07
Error handling and logging

### LINKED ISSUE #631: Breaking change: null data returned when resource not found

### LINKED ISSUE #638: JsonApiException namespace issue
The current `JsonApiException` is situated in the `JsonApiDotNetCore.Internal` namespace, it should be in the `JsonApiDotNetCore.Internal.Exceptions`, as we have a folder for that.

As discussed in #620, where we discussed that namespaces are a bit of a mess at the moment, we should take some time to correct this.

### LINKED ISSUE #655: Invalid Patch request body format required (most likely user error)

### LINKED ISSUE #689: Review logging
## Description
Now that we have the right logging dependencies in place, we should review when, what and at which level logging occurs within JADNC.

Official guidance on logging levels: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.2#log-level

I'm thinking of:
- Do not log events at **error** level which cannot be solved by us, but are functional errors that do not require investigation. Like resource-not-found, invalid-attribute-name, resource-already-exists, input validation errors etc.
- Basic trace/debug logging of JSON request/response, along with qualified URL
- Log query string parsing results, maybe

Additions are welcome!

### LINKED ISSUE #709: Bug: crash on invalid relationship name
When getting a misspelled relationship name, the code crashes at [CurrentRequestMiddleware.GetCurrentEntity()](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/ec5f8d560209738288f8907d37d3b319cfdd34e7/src/JsonApiDotNetCore/Middleware/CurrentRequestMiddleware.cs#L232) due to recent changes.

Repro steps:
- Run the unit test `Can_Get_TodoItem_WithOwner` to have a record in database
- Start project JsonApiDotNetCoreExample
- GET http://localhost:1479/api/v1/todoItems/1?include=owner and see todoItems resource followed by included owner
- GET http://localhost:1479/api/v1/todoItems/1/owner and observe successful response with person
- GET http://localhost:1479/api/v1/todoItems/1/ownerX and observe HTTP 500

bart-degreed:
This PR is mostly about error handling and logging.

- Ensured all errors go through an extensibility point (`IExceptionHandler`), which makes logging and exception handling customizable (see `CustomExceptionHandler` in tests). `DefaultExceptionHandler` provides virtual methods to adapt logging level, log message and the conversion from Exception to Error response objects.
- Converted action results into rich `JsonApiException` objects and added/updated tests to assert on the returned error objects.
- Fixed broken model state validation.
- Fixes around returning response body with NotFound for resources and relationships.
- Improved selection for which parser handles a query string parameter, allowing for better error messages.
- Fixes on handling of omitNull/omitDefault query string overrides.
- Improvements for handling `DisableQueryAttribute`.
- Added trace logging on controller/service/repository flow + JSON requests/responses.
- Fix in handling zero page size; set default page size to nonzero.

Fixes #631
Fixes #689
Fixes #655
Fixes #709
Fixes #638


src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs
```diff
@@ -61,107 +61,97 @@ public abstract class BaseJsonApiController<T, TId> : JsonApiControllerMixin whe
...
-                return UnprocessableEntity();
+                throw new InvalidRequestBodyException(null, null, null);
 
             if (!_jsonApiOptions.AllowClientGeneratedIds && !string.IsNullOrEmpty(entity.StringId))
-                return Forbidden();
+                throw new ResourceIdInPostRequestNotAllowedException();
 
             if (_jsonApiOptions.ValidateModelState && !ModelState.IsValid)
-                return UnprocessableEntity(ModelState.ConvertToErrorCollection<T>());
+                throw new InvalidModelStateException(ModelState, typeof(T), _jsonApiOptions);
```
maurei, line 117: `JsonApiOptions` is used for only one of its properties. Perhaps we can remove it as a dependency and only pass a boolean to reflect this option?

src/JsonApiDotNetCore/Middleware/ConvertEmptyActionResultFilter.cs
```diff
@@ -0,0 +1,30 @@
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.AspNetCore.Mvc.Infrastructure;
+
+namespace JsonApiDotNetCore.Middleware
+{
+    public sealed class ConvertEmptyActionResultFilter : IAlwaysRunResultFilter
```
maurei, line 7: 👍

src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs
bart-degreed: How to inject a boolean property?

bart-degreed: Update: I'm going to merge this, so I can work on related things. Will address this later.

bart-degreed: Fixed in #716.

PR #720, opened 2020-04-10
Newtonsoft serializer settings

### LINKED ISSUE #687: Bug: custom serializer settings are ignored

### LINKED ISSUE #697: Usage of OmitNull and OmitDefault query string parameters
These seem poorly supported.

I could not find any documentation for them. The test `OmitDefaultServiceTests` passes `omitNull` query string parameter to `OmitDefaultService` (which should be `OmitNullService`).
I found no tests at all for the behavior of omitDefault query string parameter.

### LINKED ISSUE #718: Proposal: use NamingStrategy from Newtonsoft.Json
Today we have our own `IResourceNameFormatter` to apply casing conventions for resource names, attributes and query string parameters. And we have custom implementations for camel case and kebab case.

Instead of having our own, I'd like to reuse the ones that ship with the Newtonsoft.Json library (currently kebab case, camel case and snake case). This would enable the next setup using options:

```csharp
options.SerializerSettings = new JsonSerializerSettings
{
    ContractResolver = new DefaultContractResolver
    {
        NamingStrategy = new KebabCaseNamingStrategy()
    }
};
```

Today we already heavily depend on the Newtonsoft.Json library, which I don't expect to change anytime soon as System.Text.Json is relatively young and [lacks support for many features](https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/664#issuecomment-585135599).

So I think now is not the time to abstract away from Newtonsoft.Json, which you'll need anyway if you want to configure things like enum string conversion, culture settings, tracing, timezone handling, dealing with circular references etc. Once System.Text.Json reaches feature parity, we'll have a better understanding how to support multiple serialization libraries and the right way to abstract settings.

bart-degreed:
Cleanup and fixes for setting custom serializer behavior. The casing convention is now set there too.
Renamed omitNull/omitDefault query string parameters to nulls/defaults to prevent casing convention issues (note the parameter value is inverted now) and extended existing partial tests for them.

Fixes #687.
Fixes #718.
Fixes #697.


src/JsonApiDotNetCore/Graph/ResourceNameFormatters/IResourceNameFormatter.cs
```diff
@@ -1,27 +0,0 @@
-using System;
```
maurei, line 1: Using NewtonSoft serializer settings is a good idea, but perhaps still worth still rely on this service `IResourceNameFormatter` and abstract away the call of the newtonsoft contract resolver in there? This way a dev can still switch entirely switch out this newtonsoft casing conventions for anything else. Also it will allow for easier decoupling of newton soft in the future

test/UnitTests/Builders/DocumentBuilderBehaviour_Tests.cs
```diff
@@ -1,3 +1,5 @@
+// TODO: Why is this file commented out?
+
```
maurei, line 2: I think this is a legact v3.x test file, can be removed

test/UnitTests/Builders/MetaBuilderTests.cs
```diff
@@ -1,3 +1,5 @@
+// TODO: Why is this file commented out?
+
```
maurei, line 2: I think this is a legact v3.x test file, can be removed

src/JsonApiDotNetCore/Graph/ResourceNameFormatters/IResourceNameFormatter.cs
maurei: Also, I liked the idea/clarity of having all "formatting related logic" in grouped together in one injectable/overridable service

bart-degreed: I'd prefer to not abstract away when only 1 implementation exists. At the time we consider supporting multiple serializers, we should look at how they deal with casing conventions and come up with the right abstraction that enables the scenarios we want to support. This sounds like [YAGNI](https://www.itexico.com/blog/software-development-kiss-yagni-dry-3-principles-to-simplify-your-life) to me.

Devs that want to use custom casing conventions can inherit from NamingStrategy and override ResolvePropertyName to apply their own logic. Then plug it into the settings like any naming strategy.

Furthermore, this enables to override naming convention via attributes, as shown [here](https://www.newtonsoft.com/json/help/html/NamingStrategyAttributes.htm). Having a single naming conventions point sounds better to me than having one for JADNC and one for Newtonsoft.JSON in case you need it.

maurei: 👍

Issue #734, opened 2020-05-08
Review public API surface


bart-degreed:
Things I'd like to change:

- [x] Rename 'entity' parameters to 'resource' in DefaultResourceService
- [x] Rename DefaultResourceRepository to EntityFrameworkCoreRepository (and update variable/parameter names likewise)
- [x] Move all types required for basic usage into JADNC's non-internal namespace. Advanced usage, like custom Service/Repository/ResourceDefinition and query filters can be in Internal namespaces. This should clarify our intent on taking breaking changes.
- [x] Make `options.DefaultPageSize` nullable (disallow setting to 0)
- [x] Rename T in `JsonApiController<T>` to TResource
- [x] Rename IQueryParameterParser to IQueryStringParser (it parses all types, not a single one)
- [x] Replace `IEnumerable<>` / `List<>` with `IReadOnlyCollection<>`
- [x] Review null checks
- [x] Validate input parameters on public members, validate options.
- [x] Rename CurrentRequest to JsonApiRequest

Other
- [x] Remove the various .gitignore files that are spread around in the repo, and update the root version from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore - #743
- [x] Cleanup launchsettings.json, enable https - #744
- [x] Cleanup excessive logging in AppVeyor - #744

maurei:
>  Rename DefaultResourceRepository to DefaultEntityRepository (and update 
variable/parameter names likewise)

What would be the motivation for this?

bart-degreed:
It is an implementation of a resource repository that uses ORM entities. Other implementations could be named AmazonResourceService, CachingRedisRepository etc.

PR #741, opened 2020-05-11
Documentation updates for v4

### LINKED ISSUE #706: Lifetime scope in Getting Started
## Description
The step-by-step guide at https://json-api-dotnet.github.io/JsonApiDotNetCore/getting-started/step-by-step.html instructs to add the following code to Startup.ConfigureServices:

```csharp
public IServiceProvider ConfigureServices(IServiceCollection services)
{
    // add the db context like you normally would
    services.AddDbContext<AppDbContext>(options =>
    { // use whatever provider you want, this is just an example
        options.UseNpgsql(GetDbConnectionString());
    }, ServiceLifetime.Transient);

    // add jsonapi dotnet core
    services.AddJsonApi<AppDbContext>();
    // ...
}
```

I wonder why `ServiceLifetime.Transient` needs to be passed. because it is not in the official EF Core 3 documentation. Official sample at https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/data/ef-rp/intro/samples/cu30/Startup.cs

Unless the scope setting is required for some reason, I think it would be better to remove it, so the context can be reused within a single request. 

See also: https://stackoverflow.com/questions/37507691/entity-framework-core-service-default-lifetime.

bart-degreed:
Fixes #706.


docs/usage/extensibility/repositories.md
```diff
@@ -1,68 +1,66 @@
...
-    services.AddScoped<IEntityRepository<Article>, AuthorizedArticleRepository>();
+    services.AddScoped<IResourceRepository<Article>, AuthorizedArticleRepository>();
     // ...
 }
 ```
 
 A sample implementation that performs data authorization might look like this.
 
-All of the methods in the DefaultEntityRepository will use the Get() method to get the DbSet<TEntity> so this is a good method to apply scoped filters such as user or tenant authorization.
+All of the methods in the DefaultResourceRepository will use the Get() method to get the DbSet<TEntity>, so this is a good method to apply scoped filters such as user or tenant authorization.
```
fdlane, line 17: did you keep the `TEntity` here because on the `DbSet`?  btw, huge thank you for all this work.

bart-degreed: Yes. This is just a first pass over the documentation, type parameters may change before stable release. Thoughts at https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/734

docs/usage/resource-graph.md
```diff
@@ -47,6 +46,18 @@ public void ConfigureServices(IServiceCollection services)
 }
 ```
 
+Be aware that the previous command does not inject resource definitions and service layer overrides. You can it combine with auto-discovery to register them.
```
fdlane, line 49: consider changing `You can it combine ...` to `You can combine it ...`

bart-degreed: nice catch!

docs/usage/resources/hooks.md
```diff
@@ -455,9 +455,9 @@ public override void BeforeImplicitUpdateRelationship(IAffectedRelationships<Per
 }
 ```
 
-## Using Resource Hooks without EF Core
+## Using Resource Hooks without Entity Framework Core
 
-If you want to use Resource Hooks without EF Core, there are several things that you need to consider that need to be met. For any resource that you want to use hooks for:
+If you want to use Resource Hooks without Entity Framework Core, there are several things that you need to consider that need to be met. For any resource that you want to use hooks for:
 1. The corresponding resource repository must fully implement `IEntityReadRepository<TEntity, TId>`
```
fdlane, line 461: Should this be `ResourceReadRepository` now?

docs/usage/resources/resource-definitions.md
```diff
@@ -48,38 +47,24 @@ Content-Type: application/vnd.api+json
 ### Single Attribute
 
 ```c#
-public class ModelResource : ResourceDefinition<Model>
+public class UserResource : ResourceDefinition<User>
 {
-    protected override List<AttrAttribute> OutputAttrs()
-        => Remove(m => m.AccountNumber);
+    public UserResource(IResourceGraph resourceGraph) : base(resourceGraph)
```
fdlane, line 52: This type of change brings so much clarity to the docs for someone like me who does not spend every day in C#. Thank you

docs/usage/resources/hooks.md
bart-degreed: Yes, thanks.

docs/usage/resources/resource-definitions.md
bart-degreed: Do you mean that the constructor is included in the docs, or that the implementation was changed?

docs/usage/sorting.md
```diff
@@ -1,21 +1,25 @@
...
 
 ```http
 GET /api/articles?sort=-author HTTP/1.1
-Accept: application/vnd.api+json
+```
+
+## Multiple
+
+```http
+GET /api/articles?sort=author,-pageCount HTTP/1.1
```
fdlane, line 22: is this `sort=author&sort=-pageCount`

docs/usage/resources/resource-definitions.md
fdlane: Just meant your use of real-world models like `User` and not using expression body syntax for examples.

docs/usage/sorting.md
fdlane: I don't think multiple sorts in this version is working. Maybe doing something wrong but I've lost this functionality.

docs/usage/resources/resource-definitions.md
bart-degreed: Thanks, that is great feedback. I've updated other examples to not use expression-bodied syntax too.

docs/usage/sorting.md
bart-degreed: Thanks for reporting. Tracked at #745.

bart-degreed: `sort=author&sort=-pageCount` is identical to `sort=author,-pageCount`.
ASP.NET Core merges multiple parameters with the same name automatically.

fdlane: Thank you, I had not realized.  

Neither seems to work when trying to execute multiple sorts on the same entity.  I understand that the nested sorts are not working, I just wanted to make sure that we called out multiple sorts on same entity is not working either.

bart-degreed: What do you mean with nested sorts? Like I  said, #745 tracks multiple sorts on the same entity.

docs/usage/resources/resource-definitions.md
```diff
@@ -90,13 +75,16 @@ _since v3.0.0_
...
 {
-    protected override PropertySortOrder GetDefaultSortOrder()
-        => new PropertySortOrder {
-            (t => t.Prop, SortDirection.Ascending),
-            (t => t.Prop2, SortDirection.Descending),
+    public override PropertySortOrder GetDefaultSortOrder()
+    {
+        return new PropertySortOrder
+        {
+            (account => account.LastLoginTime, SortDirection.Descending),
```
fdlane, line 84: I think this should be...

```C#
        (account => new { account.LastLoginTime }, SortDirection.Descending),
        (account => new { account.UserName }, SortDirection.Ascending)
```

bart-degreed: Are you sure? The next is in the unit tests:
```c#
        public override PropertySortOrder GetDefaultSortOrder()
            => new PropertySortOrder {
                (t => t.Prop, SortDirection.Ascending)
            };
```

fdlane: Seems to work fine for Strings, but for other data types the following error is returned.

From the detail: `The type must be a NewExpression.Example: article => new { article.Author };`

```json
{
    "errors": [
        {
            "id": "ba97f100-e76c-44b3-b20b-b1c4b2777a2c",
            "status": "500",
            "title": "An unhandled error occurred while processing this request.",
            "detail": "The expression returned by 't => Convert(t.InvoiceDate, Object)' for 'JsonApiDotNetCore.Internal.ResourceGraph' is of type System.Linq.Expressions.UnaryExpression and cannot be used to select resource attributes. The type must be a NewExpression.Example: article => new { article.Author };"
        }
    ]
}
````

Changing to the following seems to resolve the issue.

```c#
      public override PropertySortOrder GetDefaultSortOrder() => new PropertySortOrder
      {
        (t => new { t.InvoiceDate }, SortDirection.Descending)
      };
```

fdlane: Updated the test with the property `SomeDate` and added a sort and the test now fails.

```C#
    public class Model : Identifiable
    {
        [Attr] public string AlwaysExcluded { get; set; }
        [Attr] public string Password { get; set; }
        [Attr] public string Prop { get; set; }
        [Attr] public DateTime SomeDate { get; set; }
    }

    public override PropertySortOrder GetDefaultSortOrder()
            => new PropertySortOrder {
                (t => t.Prop, SortDirection.Ascending),
                (t => t.Password, SortDirection.Descending),
                (t => t.SomeDate, SortDirection.Descending)
            };
```

Do you want me to update the test with a pr?

bart-degreed: The error message you're getting is misleading. See https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/754 for the fix. Would be great if you could review that. Perhaps some testing with your API for additional types may reveal move problems.

fdlane: @bart-degreed, thank you.  We are testing.

fdlane: @bart-degreed closing the loop here also and noting that #754 fixed our api sorting.  Thank you.

bart-degreed:
@fdlane When you're done, can you please use the Review button to approve/request changes/etc. Also, for next time 'Start a review' / 'Finish review' works better than single comments, because each new comment sends out a new notification email.

fdlane:
@bart-degreed  Certainly.  Sorry I flooded email.

PR #754, opened 2020-05-14
Fixed: error setting default sort for non-string properties

bart-degreed:
Fixes the problem described at https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/741#discussion_r424720227.

The issue is that the specified expression gets wrapped in a type conversion expression on non-string properties, making the code unable to 'see' through the single field.

Unfortunately this code is shared between GetDefaultSortOrder and HideFields, making it possible to specify multiple fields in a single ordering (which makes no sense at all).

### DIFF
--- a/src/JsonApiDotNetCore/Internal/ResourceGraph.cs
+++ b/src/JsonApiDotNetCore/Internal/ResourceGraph.cs
@@ -84,7 +84,9 @@ namespace JsonApiDotNetCore.Internal
 
             var targeted = new List<IResourceField>();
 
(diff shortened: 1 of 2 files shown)


bart-degreed:
@fdlane does this solve your issue?

fdlane:
@bart-degreed  the non-string properties now sort with correctly.  Thank you!

```C#
      public override PropertySortOrder GetDefaultSortOrder() => new PropertySortOrder
      {
        (t => t.InvoiceDate, SortDirection.Descending)
      };
```

PR #755, opened 2020-05-14
Fixed: sorting on multiple attributes

### LINKED ISSUE #745: Bug: sorting on multiple fields does not work
A query like /api/articles?sort=authorName,-pageCount should sort on author, and within that subset sort in pageCount in descending order.

QueryableExtensions contains two methods:
```csharp
public static IOrderedQueryable<TSource> Sort<TSource>(this IQueryable<TSource> source, SortQueryContext sortQuery) { ... }

public static IOrderedQueryable<TSource> Sort<TSource>(this IOrderedQueryable<TSource> source, SortQueryContext sortQuery) { }
```

But the second one (building query based upon a parent ordering) is never called.

bart-degreed:
Fixes #745.


fdlane:
@bart-degreed your branch works for me. Thank you

PR #765, opened 2020-05-19
JADNC: Add required on post validator.

### LINKED ISSUE #472: PATCH with ValidateModelState enabled and [Required] attribute
If I provide ValidateModelState =true, then every PATCH failed on validation 
on properties with [Required] attribute.
For example I have three properties with [Required], but sending patch to one of them only. Then
two of them failed on validation.

I found that AspNetCore uses JsonPatchDocument for this use case, but maybe I missed something.
Do you know how to deal with it? 

- JsonApiDotNetCore Version: 3.0.1-alpha4

sasman0001:
Fixes #472

This feature is a custom validator which causes a resource property to be RequiredOnPost, and also validates that the resource property is not updated with a null value on Patch.

### DIFF
--- a/src/JsonApiDotNetCore/Serialization/Server/RequestDeserializer.cs
+++ b/src/JsonApiDotNetCore/Serialization/Server/RequestDeserializer.cs
@@ -45,7 +45,8 @@ namespace JsonApiDotNetCore.Serialization.Server
                 var property = attr.PropertyInfo;
                 var requiredOnPost = property.GetCustomAttributes(typeof(RequiredOnPostAttribute), false);
                 if (requiredOnPost?.FirstOrDefault() != null && attr.GetValue(entity) == null)
-                    throw new InvalidOperationException($"Attribute {attr.PublicAttributeName} is required and therefore cannot be updated to null.");
+                    throw new InvalidRequestBodyException("Changing the value of a required attribute to null is not allowed.", 
+                        $"Attribute '{attr.PublicAttributeName}' is required and therefore cannot be updated to null.", null);
 
                 _targetedFields.Attributes.Add(attr);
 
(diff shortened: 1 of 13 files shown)


src/JsonApiDotNetCore/Serialization/Server/RequestDeserializer.cs
```diff
@@ -36,16 +37,19 @@ protected override void AfterProcessField(IIdentifiable entity, IResourceField f
...
-                        "Changing the value of the requested attribute is not allowed.",
-                        $"Changing the value of '{attr.PublicAttributeName}' is not allowed.", null);
-                }
+                           "Changing the value of the requested attribute is not allowed.",
+                           $"Changing the value of '{attr.PublicAttributeName}' is not allowed.", null);
+
+                var property = attr.PropertyInfo;
+                var requiredOnPost = property.GetCustomAttributes(typeof(RequiredOnPostAttribute), false);
+                if (requiredOnPost?.FirstOrDefault() != null && attr.GetValue(entity) == null)
+                    throw new InvalidRequestBodyException("Changing the value of a required attribute to null is not allowed.",
```
bart-degreed, line 48: How would throwing here interact with ModelState validation, which returns a list of all fields that are invalid? 

Instead of throwing, can we add to the ModelStateDictionary, which picks up all validations? See usage of InvalidModelStateException, which aggregates them and combines them into a single error response.

bart-degreed:
Does this PR solve https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/472#issuecomment-462479751?

sasman0001:
No, this does not resolve https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/472#issuecomment-462479751. I have converted back to draft because I want to give another approach a try. I did make the requested changes on this branch though.

sasman0001:
Decided to go another direction, found here: https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/781

PR #766, opened 2020-05-20
Fixed: RelativeLinks

### LINKED ISSUE #763: Namespace not honored when RelativeLinks=true
## Description

In the GettingStarted App if you set `RelativeLinks = true;`  the links in response do not include the namespace. Should it not return `"self": "api/people"`?

```c#
services.AddJsonApi<SampleDbContext>(options =>
    {
        options.Namespace = "api";
        options.RelativeLinks = true;
     });
```

```json

{
    "links": {
        "self": "/people"
    },
    "data": [
        {
            "type": "people",
            "id": "1",
            "attributes": {
                "name": "John Doe"
            },
            "relationships": {
                "articles": {
                    "links": {
                        "self": "/people/1/relationships/articles",
                        "related": "/people/1/articles"
...
```
...

## Environment

- Other Relevant Package Versions: core 3.1

fdlane:
Closes #763 

Added commented out `options.RelativeLinks = true;` to the GettingStarted example project.


bart-degreed:
> Added commented out options.RelativeLinks = true; to the GettingStarted example project.

Why?

The solution (adding a / to the base path) does not seem to have anything to do with absolute/relative links. I'd expect the / to be already in the base path if needed.

We should have an integration test for relative links. you can do:

```c#
var options = (JsonApiOptions)_factory.GetService<IJsonApiOptions>();
options.UseRelativeLinks = true;
```

from within a test, before executing a request. Another approach would be to do something similar to `NoDefaultPageSizeStartup`.

fdlane:
@bart-degreed  adding the option in the GettingStarted was an attempt to help new users.  I reverted the change.

Tests have been added for true/false.  

The change I made seems to work with the responses in the Tests, GettingStarted app, and my apps tested locally.  

I reviewed the links being created in the following blocks and it seems to behave as expected now.

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/ac5b100b37246d7c8f1f24d1d9b0da79543c32d6/src/JsonApiDotNetCore/Serialization/Server/Builders/LinkBuilder.cs#L178

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/ac5b100b37246d7c8f1f24d1d9b0da79543c32d6/src/JsonApiDotNetCore/Serialization/Server/Builders/LinkBuilder.cs#L183

~https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/ac5b100b37246d7c8f1f24d1d9b0da79543c32d6/src/JsonApiDotNetCore/Serialization/Server/Builders/LinkBuilder.cs#L188

bart-degreed:
> I reviewed the links being created in the following blocks and it seems to behave as expected now.

That the code works does not mean it is the right solution. The base path should be set in middleware and be correct in all cases. I've added tests and fixed the problem to show what I had in mind. Let me know if you agree with this.

I also tried to cleanup the `GetCustomRoute` method, but found no good way up till now. Not needed for this PR, though.

fdlane:
@bart-degreed, so much better.  Thank you.

bart-degreed:
@fdlane Thanks for working on this!

Issue #730, opened 2020-05-01
Extract IIdentifiable and annotation attributes (Attr, HasOne, etc.) into separate package


ikirkpat:
## Description
Forgive me if this is a duplicate request but I couldn't find an answer...

We have a set of models that we're moving to an external package for other non-JADNC projects to use. They were originally in our JADNC app and had JADNC schema attributes on them (IIdentifiable, Attr, HasOne, etc...). We need to be able to apply a JADNC schema on them from our JADNC app while they live in another project that does not reference JADNC.

If there already is a solution, would someone point me to it?

I know EFCore solves this by providing their [Fluent Api](https://www.entityframeworktutorial.net/efcore/fluent-api-in-entity-framework-core.aspx). Something like this would be a huge help in helping us decouple our models from the infrastructure layer of JADNC.

We are using JADNC 3.1 with .NET core 2.1

bart-degreed:
This was discussed a while ago [here](https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/536#issuecomment-507174551).

bart-degreed:
Related to #661.

bart-degreed:
Unfortunately there is currently no way to accomplish this. A fluent API is being added in #771, but even with that in place, you'd still need to reference JADNC because all resources need to implement IIdentifiable.

To accomplish what you want, we'd need to extract IIdentifiable, Attr, HasOne etc. into a separate NuGet package, which you can then reference from your models project (instead of the full JADNC package).

yaniv-yechezkel:
The #771 PR takes care of mapping attributes fluently, and replaces the need for Annotations, in order to keep the Domain clean.

I don't really have an issue with referencing JADNC from the Domain, as I do so with Fluent NHibernate or EF where I end up placing Mappings/Configurations in the same assembly as the Domain.

My main concern are the Annotations, which I wanted to remove.

As for IIdentifiable, for most Domain implementations you already have concepts like Entity and Value Object, and Entity also have the concept of Id. 
I've seen I can implement IIdentifiable explicitly on the Entity so the JADNC can use StringId at runtime, and it looks fine.

In order to eliminate the need to implement IIdentifiable, and accomplish this using fluent mapping, in a similar way of mapping a property, it will require more drastic change in both discovery during graph building, and in controller's implementations.

bart-degreed:
[...]

PR #771, opened 2020-05-23
Feature/fluent mapping

yaniv-yechezkel:
Added support for Fluent API for Resource Mapping

Including:
Resource Name
Attributes and Capabilities
Relationships
Eager Loads
Links

Extended EF Demo with Models and Resource Mappings

### DIFF
--- a/src/Examples/JsonApiDotNetCoreExample/Controllers/CategoriesController.cs
+++ b/src/Examples/JsonApiDotNetCoreExample/Controllers/CategoriesController.cs
@@ -1,0 +1,18 @@
+using JsonApiDotNetCore.Configuration;
+using JsonApiDotNetCore.Controllers;
+using JsonApiDotNetCore.Services;
+using JsonApiDotNetCoreExample.Models;
+using Microsoft.Extensions.Logging;
+
+namespace JsonApiDotNetCoreExample.Controllers
+{
+    public sealed class CategoriesController: JsonApiController<Category>
+    {
+        public CategoriesController(
+            IJsonApiOptions jsonApiOptions,
+            ILoggerFactory loggerFactory,
+            IResourceService<Category> resourceService)
+            : base(jsonApiOptions, loggerFactory, resourceService)
+        { }
+    }
+}
--- a/src/Examples/JsonApiDotNetCoreExample/Controllers/ProductsController.cs
+++ b/src/Examples/JsonApiDotNetCoreExample/Controllers/ProductsController.cs
@@ -1,0 +1,24 @@
+using JsonApiDotNetCore.Configuration;
+using JsonApiDotNetCore.Controllers;
+using JsonApiDotNetCore.Data;
+using JsonApiDotNetCore.Services;
+using JsonApiDotNetCoreExample.Models;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+using System.Collections.Generic;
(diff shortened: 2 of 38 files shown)


wisepotato:
what exactly is being added here? Why are there no tests?

yaniv-yechezkel:
I've talked with Bart about adding Fluent Mapping for Resources as an alternative for Annotations
Wanted your perspective regarding the approach taken so far...

bkoelman:
@yaniv-yechezkel Correct me if I'm wrong, but I understood that the goal here is to have no NuGet reference to JADNC from your Models project (onion architecture). Using fluent mappings would enable that.

My first thought is that I like the general approach you have taken. There's many little things to improve (including we need to have tests) which we can go over at a later time. For now, the AppDomain scanning looks expensive and I think it should accept a list of assemblies/types instead, coming from Startup.Configure method. Thanks for your efforts so far!

@maurei thoughts?

yaniv-yechezkel:
Thanks @bart-degreed. 

Yes, The main goal it to keep the Domain clean.
So Fluent mapping can now replace all the Annotations, but I still have Identifiable, and it looks like it is coupled to Discovery, so it would be a challenge...

Providing a list of Assemblies would be much better, and Tests... of course!

Sure. It's a great project.

yaniv-yechezkel:
@bart-degreed I've managed to handle Resource Mapping registration during Discovery.
So it only looks for types in the Assemblies from Startup, and there's also an option to register specific Mappings.
In had to play with Application Builder Setup to make things happen in the desired order. 
Let me know what you think.

bart-degreed:
About IIdentifiable: see my comment at https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/730#issuecomment-635241309.

I've create an issue (#776) for the design this PR is supposed to implement. I'm curious about your opinion, so feel free to add comments/remarks/questions there.

bart-degreed:
Based on my proposal, I don't think we need to have assembly/type scanning at all. The building steps can simply be extracted into a separate class if the goal is to keep Startup.cs clean.

yaniv-yechezkel:
From my understating, there's already assembly/type scanning happening. This is the discovery.
I didn't add any new concept here, just relied on the current flow.
Startup remains the same. No changes at all.

bart-degreed:
You are correct that discovery also picks up resources, aside from services and definitions. This is all convention-based. I meant we should not add **extra** discovery steps (which you did at https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/771/files#diff-fbd8f7ce570dc8180eb86d101f854b1dR249).

yaniv-yechezkel:
I totally agree with you there should not be an extra step.
Sorry, the link redirects me to JsonApiApplicationBuilder.cs, but I'm not sure as to which line are you referring...

bart-degreed:
I intended to reference this line in JsonApiApplicationBuilder:
```c#
_services.TryAddSingleton<IResourceMappingService>(sp => new ResourceMappingService(_services));
```

yaniv-yechezkel:
Thanks.

It is not a separate step.

It is used in Discovery for registering the Resource Mappings provided from Startup, and then for resolving Resource Mapping when building the Graph.

When Graph Builder is looking for defining Attributes, Relationships etc on a given Resource, it asks the Resource Service for available Mapping. At this stage Resource Mapping Service relies on them to already be registered on the Container...

PR #779, opened 2020-06-04
Feature/fluent api for building resource graph

yaniv-yechezkel:
Add support for Fluent API for building Resource Graph

### DIFF
--- a/src/JsonApiDotNetCore/Builders/IResourceGraphBuilder.cs
+++ b/src/JsonApiDotNetCore/Builders/IResourceGraphBuilder.cs
@@ -2,6 +2,7 @@ using System;
 using JsonApiDotNetCore.Internal;
 using JsonApiDotNetCore.Internal.Contracts;
 using JsonApiDotNetCore.Models;
+using JsonApiDotNetCore.Fluent;
 
 namespace JsonApiDotNetCore.Builders
 {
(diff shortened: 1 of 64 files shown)


bart-degreed:
This requires some big cleanups first. Did you even look at the diff?

Also it is not following the suggested approach:

```
How would that work?

- When builder.Resource<>() is called and the resource is not part of the graph, first it runs existing attribute scanning logic
- Next, it processes the chained methods, overriding the registration that was added from attributes
```

which would prevent the massive code duplication in this PR.

To clarify, if a user writes:

```c#
builder.Resource<WorkItem>()
    .Attribute(workItem => workItem.Title)
    .PublicName("work-title")
```

it would call:

```c#
var articleContext = resourceGraphBuilder._resources.SingleOrDefault(x => x.ResourceType == typeof(WorkItem));

if (articleContext == null) { } // add empty Article context

var titleAttribute = articleContext.GetAttributes(x => x.PropertyName == "Title");
titleAttribute.PublicName = "some"
```

Finally, I find the tests hard to read using the specification pattern that you've introduced. I'd prefer to keep the AAA-pattern that's used in all other tests.

bart-degreed:
I'm closing this out, as it seems abandoned to me. Feel free to reopen if you intend to work more on this and address review feedback.

PR #782, opened 2020-06-15
Should not parse filter values as nested filters

bjornharrtell:
I have a case where filter values can contain commas, which I noted was not accepted by current implementation.

Removing the parsing of values and special case for array filters seem a bit too invasive, however it seems to be working. I'm currently doing more evaluation of this change but would welcome input.


bart-degreed:
Hi @bjornharrtell, what a coincidence. I'm currently working on a major rewrite of query string parameter handling that provides way more flexible querying capabilities compared to what we offer today. In there, I already removed the comma limitation.

bjornharrtell:
Hopefully superceded by https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/792.

bart-degreed:
Yes, it contains tests for using a comma inside text. Note though, you'll need to escape quotes by doubling them like this:

?filter=equals(name,'Jack O'' Brian')

bart-degreed:
Superseded by #792.

PR #784, opened 2020-06-16
fix: gitignore macOS file systems and scoped client serializer extension

### LINKED ISSUE #783: AddClientSerialization should register (de)serializers as scoped services
## Description
the `AddClientSerialization` extension on `IServiceCollection` registers client-side (de)serializers as singleton services, but `ResponseDeserializer` depends on `IResourceFactory` which is a scoped service, i.e. ResponseDeserializer should be registered as a scoped service too.

## Environment

maurei:
- Closes #783
- Also adds macOS system files (.DS_STORE) to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,6 +97,9 @@ StyleCopReport.xml
 *.svclog
 *.scc
 
+# MacOS file systems
+**/.DS_STORE
+
 # Chutzpah Test files
 _Chutzpah*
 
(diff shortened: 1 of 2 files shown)


maurei:
@bart-degreed I think I would prefer that. However, `DefaultResourceFactory` depends on `IServiceProvider`, which currently is a scoped service provided because `IResourceFactory` is scoped. Registering it as a singleton will result in `IServiceProvider` being the root service provider, from which no scoped services can directly be resolved. This might cause problems right [here](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/423f45f36dd62ad59dfae755d29e62fa8bae88df/src/JsonApiDotNetCore/Internal/IResourceFactory.cs#L75), what do you think?

bart-degreed:
Good point, that is going to cause problems indeed. Because what typically will be injected as a resource parameter is DbContext, which is scoped. Sounds like we'll need to make them all scoped then, which is what you initially proposed.

PR #813, opened 2020-08-28
Allow setting Property when creating AttAttribute

bjornharrtell:
I have a use case where part of my model attributes are not static, so I need a way to create AttAttribute in runtime. To properly do that I need to be able to tie a AttrAttribute to a PropertyInfo which this PR adds as an alternative constructor.

### DIFF
--- a/src/JsonApiDotNetCore/Models/Annotation/AttrAttribute.cs
+++ b/src/JsonApiDotNetCore/Models/Annotation/AttrAttribute.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Reflection;
 using JsonApiDotNetCore.Internal;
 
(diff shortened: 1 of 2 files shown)


bjornharrtell:
Alternatively could make the setter for Property non-internal?

maurei:
Hi Bjorn, thanks for your efforts. Would you mind elaborating on your use-case with an example? Preferably if you'd open up an issue for type of background information so we can discuss it further there

bjornharrtell:
@maurei that's reasonable, I've made an attempt in #814.

bjornharrtell:
See https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/814 for the discussion which led this to be closed.

PR #821, opened 2020-09-09
Fix: incorrect resource type comparison

### LINKED ISSUE #789: Type match incorrect for PATCH requests
It seems logic is at fault at:

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/36bc4896a4072b0fe330fe69e65aed78ec6f573c/src/JsonApiDotNetCore/Middleware/DefaultTypeMatchFilter.cs#L33-L34

For PATCH requests that will compare the path id parameter not the entity.

maurei:
Fixes #789 

- No longer relies on `ActionDescriptor` to get an expected type, but retrieves it from `IJsonApiRequest` instead


bart-degreed:
@maurei Can you resolve merge conflicts?

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
```diff
@@ -16,17 +21,20 @@ namespace JsonApiDotNetCore.Serialization
...
+            IJsonApiRequest jsonApiRequest,
+            IResourceContextProvider resourceContextProvider,
             ILoggerFactory loggerFactory)
         {
             if (loggerFactory == null) throw new ArgumentNullException(nameof(loggerFactory));
 
             _deserializer = deserializer ?? throw new ArgumentNullException(nameof(deserializer));
-            _request = request ?? throw new ArgumentNullException(nameof(request));
+            _jsonApiRequest = jsonApiRequest ?? throw new ArgumentNullException(nameof(jsonApiRequest));
+            _resourceContextProvider = resourceContextProvider;
```
bart-degreed, line 37: Missing null check

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
```diff
@@ -134,5 +166,27 @@ private async Task<string> GetRequestBody(Stream body)
...
+            }
+
+            return model == null ? new Type[0] : new[] { model.GetType() };
+        }
+
+        private Type GetEndpointResourceType()
+        {
+            return _jsonApiRequest.Kind == EndpointKind.Primary
+                ? _jsonApiRequest.PrimaryResource.ResourceType 
+                : _jsonApiRequest.SecondaryResource?.ResourceType;
```
bart-degreed, line 189: I'm not 100% sure if there are cases where this can fail, but I think using secondary with fallback to primary is more robust. So without checking `Kind` first. Because `Kind != Primary` does not always mean `Kind.Secondary`, it could mean `Kind.Relationship` too.

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
```diff
@@ -16,17 +21,20 @@ namespace JsonApiDotNetCore.Serialization
...
-            IJsonApiRequest request,
+            IJsonApiRequest jsonApiRequest,
+            IResourceContextProvider resourceContextProvider,
             ILoggerFactory loggerFactory)
         {
             if (loggerFactory == null) throw new ArgumentNullException(nameof(loggerFactory));
 
             _deserializer = deserializer ?? throw new ArgumentNullException(nameof(deserializer));
-            _request = request ?? throw new ArgumentNullException(nameof(request));
+            _jsonApiRequest = jsonApiRequest ?? throw new ArgumentNullException(nameof(jsonApiRequest));
```
bart-degreed, line 36: I did a solution-wide refactoring in #808 to rename all parameters and fields of type `IJsonApiRequest` from `jsonApiRequest` to `request` (and the same for options) and would like to keep it that way.

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
```diff
@@ -63,22 +71,46 @@ public async Task<InputFormatterResult> ReadAsync(InputFormatterContext context)
...
+            ValidateIncomingResourceType(context, model);
+            
             return await InputFormatterResult.SuccessAsync(model);
         }
 
+        private void ValidateIncomingResourceType(InputFormatterContext context, object model)
+        {
+            if (context.HttpContext.IsJsonApiRequest() && IsPatchOrPostRequest(context.HttpContext.Request))
+            {
+                var endPointResourceType = GetEndpointResourceType();
```
bart-degreed, line 83: `endPointResourceType` should have lowercase 'p' (same as in method name)

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
```diff
@@ -63,22 +71,46 @@ public async Task<InputFormatterResult> ReadAsync(InputFormatterContext context)
...
+                var bodyResourceTypes = GetBodyResourceTypes(model);
+
+                foreach (var bodyResourceType in bodyResourceTypes)
+                {
+                    if (bodyResourceType != null && endPointResourceType != null && bodyResourceType != endPointResourceType)
```
bart-degreed, line 88: I don't think `bodyResourceType` can ever be null, so comparing both without the null checks should be fine. If not, entering the loop only when `endPointResourceType != null` (check upfront) seems more appropriate, as the condition never changes when iterating through the loop.

bart-degreed (changes requested):
Nice, Looks great!

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
maurei: Good catch, will add

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
maurei: Good catch, will modify

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
maurei: Good catch, will modify

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
maurei: Your right! the `bodyResourceType` null check is redundant because when the resource type from the body resolves to `null`, this is already caught [here](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/f27e020418c5f08762f57a85f221f3aa698019b8/src/JsonApiDotNetCore/Serialization/JsonApiReader.cs#L182).

Will process both suggestions

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
maurei: At this stage `Kind` can only be `Relationship` or `Primary, because there can be no json:api POST or PATCH request on a secondary endpoint. `POST /todoItems/1/assignee` is not supported

A fallback on the primary resource when the secondary resource is `null` will also not suffice, because it is possible to have a request on a `Kind.Relationship` endpoint with an unknown relationship, in which case the `null` is actually desired and should not lead to a fallback to the primary resource. See [Fails_On_Unknown_Relationship](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/f27e020418c5f08762f57a85f221f3aa698019b8/test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/UpdatingRelationshipsTests.cs#L774).

bart-degreed: Good points! So this already was the best approach after all, let's leave it like that.

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
```diff
@@ -134,5 +170,27 @@ private async Task<string> GetRequestBody(Stream body)
...
         }
+        
+        private bool IsPatchOrPostRequest(HttpRequest request)
+        {
+            return request.Method == HttpMethods.Patch || request.Method == HttpMethods.Post;
+        }
+
+        private IEnumerable<Type> GetBodyResourceTypes(object model)
+        {
+            if (model is ICollection<IIdentifiable> resourceCollection)
```
bart-degreed, line 181: This can be `IEnumerable<T>` now.

bart-degreed (changes requested):
Looking good, one more small change.

src/JsonApiDotNetCore/Serialization/JsonApiReader.cs
maurei: Fixed!

bart-degreed (approved):

PR #826, opened 2020-09-14
Break up AttrCapabilities.AllowChange

### LINKED ISSUE #801: Add AttrCapabilities.AllowCreate
## Description
This would work similar to `AttrCapabilities.AllowUpdate`: we would block POST requests that set the attribute with HTTP 400. 
Intended for attributes that are set automatically from code or the database, such as creation dates.

bart-degreed:
Before, `AttrCapabilities.AllowChange` was used for both POST and PATCH requests (although documentation said it was only for PATCH).

This commit adds a new flag and changes the meaning of the old one:
- AttrCapabilities.AllowCreate affects only POST
- AttrCapabilities.AllowChange affects only PATCH

Fixes #801.

### DIFF
--- a/docs/usage/resources/attributes.md
+++ b/docs/usage/resources/attributes.md
@@ -50,9 +50,21 @@ public class User : Identifiable<int>
 }
 ```
 
-### Mutability
+### Createability
 
-Attributes can be marked as mutable, which will allow `PATCH` requests to update them. When immutable, an HTTP 422 response is returned.
+Attributes can be marked as creatable, which will allow `POST` requests to assign a value to them. When sent but not allowed, an HTTP 422 response is returned.
+
+```c#
+public class Person : Identifiable<int>
+{
+    [Attr(Capabilities = AttrCapabilities.AllowCreate)]
+    public string CreatorName { get; set; }
+}
+```
+
+### Changeability
+
+Attributes can be marked as changeable, which will allow `PATCH` requests to update them. When sent but not allowed, an HTTP 422 response is returned.
 
 ```c#
 public class Person : Identifiable<int>
--- a/docs/usage/resources/attributes.md
+++ b/docs/usage/resources/attributes.md
@@ -50,7 +50,7 @@ public class User : Identifiable<int>
 }
 ```
 
-### Createability
+### Creatability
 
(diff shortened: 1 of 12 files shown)


test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/CreatingDataTests.cs
```diff
@@ -325,6 +325,39 @@ public async Task CreateResource_UnknownResourceType_Fails()
             Assert.Contains("Request body: <<", errorDocument.Errors[0].Detail);
         }
 
+        [Fact]
+        public async Task CreateResource_Blocked_Fails()
```
maurei, line 329: We can be more explicit on how it fails: 422

docs/usage/resources/attributes.md
```diff
@@ -50,9 +50,21 @@ public class User : Identifiable<int>
 }
 ```
 
-### Mutability
+### Createability
```
maurei, line 53: [I think it should be creatability](https://english.stackexchange.com/questions/400406/create-createable-update-updateable-delete-deletable), as spelled on the next line

maurei (changes requested):

docs/usage/resources/attributes.md
bart-degreed: Wow, that surprises me! When I wasn't sure about the right spelling, I looked it up and found changeability. And then I assumed the same would apply to creation. Turns out it does not, so I will adapt.

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/CreatingDataTests.cs
bart-degreed: I considered adding that, which means breaking with the style of existing tests in this file. Then I decided to not add it, knowing we're going to give all tests a bit overhaul in the future. Are you okay with leaving it like this for now?

maurei: Makes sense! Agreed.

maurei (approved):

Issue #696, opened 2020-02-19
How to deserialize a class with a relationship that is a derived class?


marioGomezDev:
## Description
My domain are set up like this:

```
abstact class Person {
	...
}

class Student {
	...
}

class Teacher{
	...
}

class Article {
	Person Author
}

```
When I post an Article, it fails to deserialize the specific Teacher or Student class, how can I fix it?
...

## Environment

- JsonApiDotNetCore Version: 4.0.0.0

maurei:
Thanks for reporting this issue. There might be some issues with the (de)serialization of inherited properties, but to be sure what is going wrong and to see if we need to patch anything in the internals, we would need some more details. Preferably if you could open up a reproduction repository for us to step into and debug, or even better would be a PR with a few tests that reproduce the issue in the JsonApiExample project

marioGomezDev:
In this repository, you can see the issue
[https://github.com/marioGomezDev/DeserializeIssue.git](url)

Test example:
[https://www.getpostman.com/collections/d8f6c4df7407152a9f0b](url)

bart-degreed:
Hi @marioGomezDev, The repro repository your provided targets JADNC v3, so I [upgraded](https://github.com/marioGomezDev/DeserializeIssue/pull/1) it to .NET Core 3.1 and JADNC 4-beta1. After that, I was able to reproduce your issue using the next request:

```http
POST https://localhost:44308/articles HTTP/1.1

{
  "data": {
    "type": "articles",
    "attributes": {
      "article-prop": "AP-new"
    },
    "relationships": {
      "author": {
        "data": {
          "type": "students",
          "id": "1"
        }
      }
    }
  }
}
```

This crashes in the deserialization process, because it tries to create an instance of abstract Person class. Instead it should look at the actual type ("students" in this case) and so create an instance of Student. I have a preliminary fix at #828.

For context, the next request shows the sample data:

```http
GET https://localhost:44308/articles?include=author HTTP/1.1
```
```json
{
  "links": {
    "self": "https://localhost:44308/articles?include=author"
  },
  "data": [
    {
      "type": "articles",
      "id": "1",
      "attributes": {
        "article-prop": "ArticleProp 1"
      },
      "relationships": {
        "author": {
          "links": {
            "self": "https://localhost:44308/articles/1/relationships/author",
            "related": "https://localhost:44308/articles/1/author"
          },
          "data": {
            "type": "students",
            "id": "1"
          }
        }
[...]

PR #828, opened 2020-09-14
Fix: crash when deserializing post with relationship to abstract base class

bart-degreed:
I'm not sure if this is a good solution, but at least this removes the crash when trying to create an instance of abstract base class (see #696). There are likely more places we need to change.

### DIFF
--- a/src/JsonApiDotNetCore/Repositories/RepositoryRelationshipUpdateHelper.cs
+++ b/src/JsonApiDotNetCore/Repositories/RepositoryRelationshipUpdateHelper.cs
@@ -114,6 +114,7 @@ namespace JsonApiDotNetCore.Repositories
 
             var newLinks = relationshipIds.Select(x =>
             {
+                // TODO: [#696] Potential location where we crash if the relationship targets an abstract base class.
                 var link = _resourceFactory.CreateInstance(relationship.ThroughType);
                 relationship.LeftIdProperty.SetValue(link, TypeHelper.ConvertType(parentId, relationship.LeftIdProperty.PropertyType));
                 relationship.RightIdProperty.SetValue(link, TypeHelper.ConvertType(x, relationship.RightIdProperty.PropertyType));
--- a/src/JsonApiDotNetCore/Resources/Annotations/HasManyThroughAttribute.cs
+++ b/src/JsonApiDotNetCore/Resources/Annotations/HasManyThroughAttribute.cs
@@ -137,6 +137,7 @@ namespace JsonApiDotNetCore.Resources.Annotations
                 List<object> throughResources = new List<object>();
                 foreach (IIdentifiable identifiable in (IEnumerable)newValue)
                 {
+                    // TODO: [#696] Potential location where we crash if the relationship targets an abstract base class.
                     object throughResource = resourceFactory.CreateInstance(ThroughType);
                     LeftProperty.SetValue(throughResource, resource);
                     RightProperty.SetValue(throughResource, identifiable);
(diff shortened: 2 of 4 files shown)


src/JsonApiDotNetCore/Serialization/BaseDeserializer.cs
```diff
@@ -168,15 +168,19 @@ private IIdentifiable ParseResourceObject(ResourceObject data)
             var rio = (ResourceIdentifierObject)relationshipData.Data;
             var relatedId = rio?.Id;
 
+            var resourceContext = relationshipData.SingleData == null
```
maurei, line 171: I think this can be simplified?

```c#
var relatedResourceType = relationshipData.SingleData == null
    ? attr.RightType
    : ResourceContextProvider.GetResourceContext(relationshipData.SingleData.Type).ResourceType;

// this does not make sense in the following case: if we're setting the dependent of a one-to-one relationship, IdentifiablePropertyName should be null.
var foreignKeyProperty = resourceProperties.FirstOrDefault(p => p.Name == attr.IdentifiablePropertyName);

if (foreignKeyProperty != null)
    // there is a FK from the current resource pointing to the related object,
    // i.e. we're populating the relationship from the dependent side.
    SetForeignKey(resource, foreignKeyProperty, attr, relatedId, relatedResourceType);

SetNavigation(resource, attr, relatedId, relatedResourceType);
```

maurei: Also: isn't `ResourceContextProvider.GetResourceContext(relationshipData.SingleData.Type).ResourceType;` always just equal to `relationshipData.SingleData.Type`?  Because `relationshipData.SingleData.Type` is parsed from the request body, so if I'm not mistaken this will never equal the abstract base type.

src/JsonApiDotNetCore/Serialization/BaseDeserializer.cs
```diff
@@ -232,6 +237,7 @@ private void SetNavigation(IIdentifiable resource, HasOneAttribute attr, string
             {   // if the relationship is set to null, no need to set the navigation property to null: this is the default value.
                 var relatedResources = relationshipData.ManyData.Select(rio =>
                 {
+                    // TODO: [#696] Potential location where we crash if the relationship targets an abstract base class.
```
maurei, line 240: I think this is valid. I think we'll have to introduce the equivalent approach as for SetHasOneRelationship

src/JsonApiDotNetCore/Serialization/Client/Internal/ResponseDeserializer.cs
```diff
@@ -87,6 +87,7 @@ protected override void AfterProcessField(IIdentifiable resource, ResourceFieldA
         /// </summary>
         private IIdentifiable ParseIncludedRelationship(RelationshipAttribute relationshipAttr, ResourceIdentifierObject relatedResourceIdentifier)
         {
+            // TODO: [#696] Potential location where we crash if the relationship targets an abstract base class.
```
maurei, line 90: True

src/JsonApiDotNetCore/Repositories/RepositoryRelationshipUpdateHelper.cs
```diff
@@ -114,6 +114,7 @@ private async Task UpdateManyToManyAsync(IIdentifiable parent, HasManyThroughAtt
 
             var newLinks = relationshipIds.Select(x =>
             {
+                // TODO: [#696] Potential location where we crash if the relationship targets an abstract base class.
```
maurei, line 117: Interesting. This one seems a bit more difficult to resolve

src/JsonApiDotNetCore/Resources/Annotations/HasManyThroughAttribute.cs
```diff
@@ -137,6 +137,7 @@ public override void SetValue(object resource, object newValue, IResourceFactory
                 List<object> throughResources = new List<object>();
                 foreach (IIdentifiable identifiable in (IEnumerable)newValue)
                 {
+                    // TODO: [#696] Potential location where we crash if the relationship targets an abstract base class.
```
maurei, line 140: Same for this one

maurei (changes requested):

maurei:
Superceded by #833 833

Issue #671, opened 2020-01-23
Relations being validated on POST


Lobosque:
## Description
I am trying to POST an `enterprise` resource. while posting it, I need to associate it to a `partner`:

```javascript
{
  "data": {
    "type": "enterprise",
    "attributes": {
      "name": "Ember",
      "alias": "Emba",
      "documentNumber": "1234567"
    },
    "relationships": {
      "partner": {
        "data": { "type": "partner", "id": 10 }
      },
    }
  }
}
```

And these are the models:
```
    [Resource("enterprise")]
    public class Enterprise : Identifiable, IEntity
    {
        [Column("EnterpriseId")]
        public override int Id { get; set; }

        [Required]
        [Attr]
        public string Name  { get; set; }

        [Required]
        [Attr]
        public string Alias { get; set; }

        [Required]
        [Attr]
        public string DocumentNumber { get; set; }

        public int PartnerId { get; set; }
        [HasOne("partner")]
        public virtual EnterprisePartner Partner { get; set; }
    }

    [Resource("partner")]
    public class EnterprisePartner : Identifiable, IEntity
    {
        [Required]
        [Attr]
        public string Token { get; set; }

        [Required]
        [Attr]
        public string Name { get; set; }

        [Attr]
        public string Alias { get; set; }
    }
```
I am getting the following return: 
```
{
    "errors": [
        {
            "title": "ArgumentNullException",
            "detail": "Value cannot be null. (Parameter 'element')",
            "status": "500"
        }
    ]
}
```

While debugging, I found that the partner attributes are being validated, and of course not passing since I am not sending any value:
```
[0] [string]:"Partner.Name"
[1] [string]:"Partner.Token"
```
I expect that while Posting, I am just associating the entities, not trying to create a new one. Should I be doing this in a different way?
The final error happens when the code tries to access `Partner.Name` attribute from `Enterprise`, but the actual problem seems to be the validation of the related resource. 

## Environment

- JsonApiDotNetCore 4.0.0alpha

maurei:
Hi @Lobosque. If I'm understanding correctly you're trying to use create a resource and relate it to an already existing resource [this part of the json:api spec](https://jsonapi.org/format/#crud-creating), but the usage of .NET Core's `[Required]` attributes is messing up the pipeline?

bart-degreed:
I haven't tried, but this may have been fixed in #781, which is now in master. @Lobosque can you confirm?

[...]

PR #829, opened 2020-09-14
Fixed: crash when building ModelState error response

bart-degreed:
It may happen that `propertyName` contains something like 'Parent.Child.Some', so `property` becomes `null` and we crash on the next line with a `NullReferenceException`. This fix returns the original value if we are unable to translate between private/public names.

Found this while trying to find the root cause for #671.


bart-degreed:
Superseded by #831.

PR #831, opened 2020-09-15
Fixed: Do not run ModelState validation for relationships in POST/PATCH

### LINKED ISSUE #671: Relations being validated on POST

bart-degreed:
Fixes #671.


bart-degreed:
@maurei One test is disabled, I found no way to get that one working.
Tagging @sasman0001 because she implemented IsRequired in the past and may be aware of a solution.

Ideas are welcome! :)

If we find no solution, perhaps its better to take that test out.

sasman0001:
Hey Bart. I branched from your branch to try and see if I could come up with any ideas. But when I tried to push what I came up with, it says I don't have access. So I will just describe it below:

In `IsRequiredAttribute.cs` I added the following to the `ShouldSkipValidationForResource` method:

```     
                if (TypeHelper.IsOrImplementsInterface(validationContext.ObjectType, typeof(IHasParent)))
                {
                    var parent = (IHasParent)validationContext.ObjectInstance;
                    return parent.HasParent;
                }
```

then in `Enterprise.cs ` I user the IHasParent interface and implement it as follows:

```
       public bool HasParent { get; set; }

        private Enterprise _parent;

        [HasOne]
        public Enterprise Parent
        {
            get => _parent;
            set
            {
                _parent = value;
                _parent.HasParent = true;
            }
        }
```
The `IHasParent.cs` only has ` bool HasParent { get; set; }` within it.

When I make these changes and remove the skip on test `When_patching_resource_with_self_reference_it_must_succeed` all tests succeed.

src/JsonApiDotNetCore/Middleware/HttpContextExtensions.cs
```diff
@@ -5,42 +5,44 @@ namespace JsonApiDotNetCore.Middleware
...
 
-        internal static void DisableValidator(this HttpContext httpContext, string propertyName, string model)
+        internal static void DisableRequiredValidator(this HttpContext httpContext, string propertyName, string model)
         {
             if (httpContext == null) throw new ArgumentNullException(nameof(httpContext));
             if (propertyName == null) throw new ArgumentNullException(nameof(propertyName));
             if (model == null) throw new ArgumentNullException(nameof(model));
 
-            var itemKey = $"JsonApiDotNetCore_DisableValidation_{model}_{propertyName}";
+            var itemKey = _disableRequiredValidatorKey + $"_{model}_{propertyName}";
```
maurei, line 35: Use `$"{_disableRequiredValidatorKey}_{model}_{propertyName}";` for consistent string interpolation

src/JsonApiDotNetCore/Middleware/HttpContextExtensions.cs
```diff
@@ -5,42 +5,44 @@ namespace JsonApiDotNetCore.Middleware
...
-        internal static bool IsValidatorDisabled(this HttpContext httpContext, string propertyName, string model)
+        internal static bool IsRequiredValidatorDisabled(this HttpContext httpContext, string propertyName, string model)
         {
             if (httpContext == null) throw new ArgumentNullException(nameof(httpContext));
             if (propertyName == null) throw new ArgumentNullException(nameof(propertyName));
             if (model == null) throw new ArgumentNullException(nameof(model));
 
-            return httpContext.Items.ContainsKey($"JsonApiDotNetCore_DisableValidation_{model}_{propertyName}") ||
-                   httpContext.Items.ContainsKey($"JsonApiDotNetCore_DisableValidation_{model}_Relation");
+            return httpContext.Items.ContainsKey(_disableRequiredValidatorKey + $"_{model}_{propertyName}");
```
maurei, line 45: Use `$"{_disableRequiredValidatorKey}_{model}_{propertyName}";` for consistent string interpolation

maurei (changes requested):

src/JsonApiDotNetCore/Resources/Annotations/IsRequiredAttribute.cs
```diff
@@ -11,22 +11,52 @@ namespace JsonApiDotNetCore.Resources.Annotations
...
+            {
+                return ValidationResult.Success;
+            }
+
+            return base.IsValid(value, validationContext);
         }
 
-        /// <inheritdoc />
-        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        private static bool ShouldSkipValidationForResource(ValidationContext validationContext, IJsonApiRequest request)
```
maurei, line 33: Is there a particular reason for this being a static method? If there is no objection, I would prefer an instance method.

Even though this method _can be_ static perfectly fine, I find it more readable to use an instance method here because its callees only live within the same class and are instance methods too, and because there is no static usage of `IsRequiredAttribute` (since `Attribute`s are always instantiated by aspnet core)

maurei (changes requested):

maurei (changes requested):
The result of @sasman0001 will lead to a passing test. But my objection is that it only shows a workaround that a developer can implement in the case of self-referencing models. It does not introduce a solution at the level of the library itself. I would be in favour of the latter.

We could reflect on to object which is to be validated and detect if it is self-referencing. Caching that result we can then skip the validation when appropriate.

I've made a PR to your branch so you can check out the fix.

On the side: is there any reason for your branches to live on your own fork? It would be easier for me to work with your code if your branches lived in this repository.

sasman0001:
That makes sense. Good point @maurei.

src/JsonApiDotNetCore/Middleware/HttpContextExtensions.cs
bart-degreed: Good catch, thanks!

src/JsonApiDotNetCore/Resources/Annotations/IsRequiredAttribute.cs
bart-degreed: The sole reason is that I used the 'extract method' refactoring, which determined that the method can be static so that's what it applied. I have no strong preference in this case and will adapt.

bart-degreed:
@maurei Although creative, I felt the exact same objection about @sasman0001's solution. So I decided to wait a bit and see if you'd come up with an alternative solution. I think your approach is solid and I'm going to integrate that.

About the forking, you asked me that last week too! So let me repeat my reply: This stems from the time I wasn't a JADNC project contributor and had limited permissions like everyone else. At that time, I created some branches on my fork that I cannot close yet. I'm hoping to ditch my fork in the near future (it makes sync easier), but until then I do not want to have *two* JANDC repositories on my system: it is just too easy to be accidentally working in the wrong one.

src/JsonApiDotNetCore/Resources/Annotations/IsRequiredAttribute.cs
```diff
@@ -11,6 +14,9 @@ namespace JsonApiDotNetCore.Resources.Annotations
     /// </summary>
     public sealed class IsRequiredAttribute : RequiredAttribute
     {
+        // TODO: We need put this cache in some context, as it is not thread-safe (parallel requests) and grows indefinitely.
```
maurei, line 17: Good catch

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ModelStateValidation/SystemDirectory.cs
```diff
@@ -0,0 +1,38 @@
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using JsonApiDotNetCore.Resources;
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.ModelStateValidation
+{
+    public sealed class SystemDirectory : Identifiable
```
maurei, line 8: This example is a lot more intuitive, nicely done.

maurei (approved):

Issue #737, opened 2020-05-11
IHasMeta does not work


bart-degreed:
Example:

```csharp
    public class Person : Identifiable, IHasMeta
    {
        [Attr]
        public string FirstName { get; set; }
        
        [Attr]
        public string LastName { get; set; }

        public Dictionary<string, object> GetMeta()
        {
            return new Dictionary<string, object>
            {
                {"copyright", "Copyright 2018 Example Corp."},
                {"authors", new[] {"John Doe"}}
            };
        }
    }
```

```
curl https://localhost:44354/people
```

```json
{"links":{"self":"https://localhost:44354///people"},"data":[{"type":"people","id":"1","attributes":{"firstName":"John","lastName":"Simmons"},"links":{"self":"https://localhost:44354///people/1"}},{"type":"people","id":"2","attributes":{"firstName":"Pete","lastName":"Bell"},"links":{"self":"https://localhost:44354///people/2"}}]}
```

bart-degreed:
Same for `IRequestMeta`:

```csharp
public class RequestMetaService : IRequestMeta
{
    public Dictionary<string, object> GetMeta()
    {
        return new Dictionary<string, object>
        {
            {"copyright", "Copyright 2018 Example Corp."},
            {"authors", new string[] {"John Doe"}}
        };
    }
}
```

bart-degreed:
Current behavior is that `IRequestMeta` adds top-level meta to the response document unconditionally, while `IHasMeta` (which was replaced with `IResourceDefinition.GetMeta` in #832) adds top-level meta only when the requested resource type matches.

We've discussed to change this, because:
- The json:api spec allows meta to occur per resource _instance_ (non top-level), which we currently do not support but seems more useful.
- We suspect that the matching of resource type does not work properly. For example, there is no obvious requested resource type for a request like `/blogs/1/relationships/articles`, because it functionally targets a relationship between two resources of potentially different types.

Our goal is to change this to:
1. Support top-level meta unconditionally (typically used for API version, copyright notice etc) -- as-is
2. Support nested meta per resource instance (use case: some articles may have a discount, product recalls etc) -- requires that the resource instance is passed to `IResourceDefinition.GetMeta` calls and it is invoked (multiple times) during the serialization process.

PR #832, opened 2020-09-16
Generic resource definitions

### LINKED ISSUE #806: Generic resource definitions
### Current situation
The current way how resource definitions work is that JADNC provides a generic base class, which users can derive from for a specific resource type. JADNC contains scanning logic that picks up such non-generic (resource-bound) classes and calls into them from various places.

Example:

```c#
public class OrderLineDefinition : ResourceDefinition<OrderLine>
{
    // ...
}
```

When no resource-bound class is found for a specific resource, then resource callbacks are skipped for that resource type. This is accomplished via a non-generic `IResourceDefinition` interface, which is a type intended for internal use only to track which callbacks are available. It put me on the wrong track when I started using JADNC: I expected this was something for me to implement (similar to services and repositories), but it is not.

### Proposal

I'd like to make resource definitions a first-class pluggable extensibility point, similar to services and repositories, so that resource definitions are *always* called into. We would provide an interface along with a default generic implementation (that contains no logic). Users can then implement the interface themselves and register their generic type -and- derive from our default generic implementation and register that -and- declare resource-bound definitions, which may derive from their own generic base class. 

Example:
```c#
namespace JADNC
{
    public interface IResourceDefinition<TResource>
    {
        // ...
    }

    public class JsonApiResourceDefinition<TResource> : IResourceDefinition<TResource>
    {
        // ...
    }
}

namespace ApiDeveloper
{
    public class CustomResourceDefinition<TResource> : JsonApiResourceDefinition<TResource>
    {
        // ...
    }

    public class OrderLineResourceDefinition : CustomResourceDefinition<OrderLine>
    {
        // ...
    }
}
```

### Use case

The use case I have in mind for this is authorization. Assume the system stores whether users have read/write access per resource type. While it is possible to annotate controller methods with attributes, this would not be sufficient. It would only work for primary requests like `/articles` and `/articles/1`, but not secondary requests like `/articles/1/author`. The controller would need to contain custom logic that inspects the `relationship` parameter and runs the permission check for that too. It is even debatable whether a permission check on articles should be done in that case, because the request is not targeting an article.

But it gets worse: a request like `/articles?include=author` would not get checked for access to the author resource. The current way to deal with that would be to override `ResourceDefinition.OnApplyIncludes` for all resource types in the project. A similar override for `ResourceDefinition.OnApplyFilter` is needed to properly check a request like `/articles?filter=any(revisions)` to ensure the caller has read permission on the revisions table.

If it were possible to plug in a generic resource definition, it could handle the permission check for all resource types in the project. Resource-bound classes would then call into the base class to ensure authorization is always applied.

bart-degreed:
Fixes #806.

Made resource definitions a first-class pluggable extensibility point, similar to resource services and repositories. So developers now can derive from `JsonApiResourceDefinition`, implement `IResourceDefinition`, or insert base classes in the type hierarchy to share common logic. Such a base class can then be registered as an open generic, so that all resources will use that (unless the container finds a non-generic class, which is a better match) instead of the built-in `JsonApiResourceDefinition`.

To accomplish this, I had to split up `ResourceDefinition` into two:
- `JsonApiResourceDefinition` is highly pluggable like described above
- `ResourceHooksDefinition` contains just the hook callbacks

This way, the recursive resolve logic for resource hooks remains as-is. Since it did not use `IResourceDefinitionProvider`, that type has been replaced with `IResourceDefinitionAccessor`, which invokes the requested callbacks directly on the found class. What used to be `IResourceDefinition` (an internal type to invoke callbacks on) no longer exists.

Additionally replaced `IHasMeta` with a method on `IResourceDefinition` and only scan for hook definitions when they are enabled. I intend to update documentation on `IHasMeta` (which was already wrong) later, as part of #737 (which may change these again).


src/JsonApiDotNetCore/Configuration/ServiceDiscoveryFacade.cs
```diff
@@ -96,12 +101,12 @@ internal void DiscoverResources()
...
                 {
-                    AddResource(assembly, descriptor);
+                    AddResource(descriptor);
                 }
             }
         }
-        
-        internal void DiscoverInjectables()
+
+        internal void DiscoverInjectables(bool enableResourceHooks)
```
maurei, line 109: Alternatively we could pass `IJsonApiOptions` to the constructor of `ServiceDiscoveryFacade` and read the option from there. That feels a little bit cleaner to me. Thought?

maurei (changes requested):
Partial review, will continue later today

src/JsonApiDotNetCore/Configuration/ServiceDiscoveryFacade.cs
bart-degreed: Sounds good, I've change it.

maurei (approved):

docs/usage/meta.md
```diff
@@ -1,43 +1,60 @@
 # Metadata
 
-Non-standard metadata can be added to your API responses in two ways: Resource and Request Meta. In the event of a key collision, the Request Meta will take precendence.
+Top-level custom metadata can be added to your API responses in two ways: globally and per resource type. In the event of a key collision, the resource meta will take precendence.
```
maurei, line 3: I think the word "custom" is redundant here

docs/usage/meta.md
```diff
@@ -1,43 +1,60 @@
...
 
-## Resource Meta
+## Global Meta
 
-Resource Meta is metadata defined on the resource itself by implementing the `IHasMeta` interface.
+Global metadata can be added by injecting a service that implements `IResponseMeta`.
```
maurei, line 7: Is injecting the right terminology here? I usually refer to _injecting_ as adding a service to a constructors signature. Did you mean _registering_?

docs/usage/meta.md
```diff
@@ -1,43 +1,60 @@
...
+This is useful if you need access to other injected services to build the meta object.
```
maurei, line 8: _registered_ services

maurei (changes requested):

docs/usage/meta.md
bart-degreed: It is, thanks.

docs/usage/meta.md
bart-degreed: Agreed.

docs/usage/meta.md
bart-degreed: Agreed.

maurei (approved):

PR #833, opened 2020-09-18
Fix: crash when deserializing post with relationship to abstract base class

### LINKED ISSUE #696: How to deserialize a class with a relationship that is a derived class?

maurei:
Fixes #696 

Introduces support for inheritance in resource models


src/JsonApiDotNetCore/Serialization/BaseDeserializer.cs
```diff
@@ -168,15 +169,19 @@ private IIdentifiable ParseResourceObject(ResourceObject data)
             var rio = (ResourceIdentifierObject)relationshipData.Data;
             var relatedId = rio?.Id;
 
+            var relatedResourceType = relationshipData.SingleData == null
```
bart-degreed, line 172: I find this name confusing: related to what? It seems to hold a `System.Type` (not the public type name, as in "articles"), so I think `relationshipType` would be a better fit.

src/JsonApiDotNetCore/Serialization/BaseDeserializer.cs
```diff
@@ -187,7 +192,8 @@ private IIdentifiable ParseResourceObject(ResourceObject data)
         /// Sets the dependent side of a HasOne relationship, which means that a
         /// foreign key also will to be populated.
         /// </summary>
-        private void SetForeignKey(IIdentifiable resource, PropertyInfo foreignKey, HasOneAttribute attr, string id)
+        private void SetForeignKey(IIdentifiable resource, PropertyInfo foreignKey, HasOneAttribute attr, string id,
```
bart-degreed, line 195: The comment above this line reads "foreign key also will to be populated". It is not part of the diff, but while you're at it, could you please correct this sentence as well?

src/JsonApiDotNetCore/Serialization/BaseDeserializer.cs
```diff
@@ -168,15 +169,19 @@ private IIdentifiable ParseResourceObject(ResourceObject data)
...
             var foreignKeyProperty = resourceProperties.FirstOrDefault(p => p.Name == attr.IdentifiablePropertyName);
 
             if (foreignKeyProperty != null)
                 // there is a FK from the current resource pointing to the related object,
                 // i.e. we're populating the relationship from the dependent side.
-                SetForeignKey(resource, foreignKeyProperty, attr, relatedId);
+                SetForeignKey(resource, foreignKeyProperty, attr, relatedId, relatedResourceType);
 
-            SetNavigation(resource, attr, relatedId);
+            SetNavigation(resource, attr, relatedId, relatedResourceType);
```
bart-degreed, line 184: In this file, I would have expected similar changes in `SetHasManyRelationship`, which also calls `CreateInstance`. I marked that as a TODO in the closed PR. Is this an oversight?

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Controllers.cs
```diff
@@ -0,0 +1,28 @@
+using JsonApiDotNetCore.Configuration;
+using JsonApiDotNetCore.Controllers;
+using JsonApiDotNetCore.Services;
+using Microsoft.Extensions.Logging;
+
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance
+{
+    public sealed class MalesController : JsonApiController<Male>
```
bart-degreed, line 8: Please put these in separate files, to make it easier to find them.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models.cs
```diff
@@ -0,0 +1,49 @@
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations.Schema;
+using JsonApiDotNetCore.Resources;
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance
+{
+    public abstract class Person : Identifiable { }
+    
+    public sealed class Male : Person
```
bart-degreed, line 10: Same here, please put them in separate files.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models.cs
```diff
@@ -0,0 +1,49 @@
...
+        [Attr]
+        public string FemaleProperty { get; set; }
+    }
+    
+    public sealed class Placeholder : Identifiable
+    {
+        [HasOne]
+        public Person OneToOnePerson { get; set; }
+        
+        public int? OneToOnePersonId { get; set; }
```
bart-degreed, line 27: I think this property can be removed. ID properties are only needed for join relationships (many-to-many via HasManyThrough).

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models.cs
```diff
@@ -0,0 +1,49 @@
...
+        
+        [HasMany]
+        public List<Person> OneToManyPersons { get; set; }
```
bart-degreed, line 30: I'd prefer to switch to a more humanly understandable model. For example, Ferrari and Lamborghini inherit from Car, or Cat and Dog which are both Animals.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
```diff
@@ -0,0 +1,302 @@
...
+
+        public ResourceInheritanceTests(IntegrationTestContext<TestableStartup<ResourceInheritanceDbContext>, ResourceInheritanceDbContext> testContext)
+        {
+            _testContext = testContext;
+        }
+
+        [Theory]
+        [InlineData("males", 0)]
+        [InlineData("females", 1)]
+        public async Task Can_create_resource_with_one_to_one_relationship_that_has_inheritance(string type, int index)
```
bart-degreed, line 28: I think we should plan a meeting and decide on test naming. Here, I'd prefer the `When_[scenario]_it_should_[expected-outcome]` pattern.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
```diff
@@ -0,0 +1,302 @@
...
+            
+            responseDocument.SingleData.Should().NotBeNull();
+            responseDocument.SingleData.Relationships["oneToOnePerson"].Should().NotBeNull();
+            responseDocument.SingleData.Relationships["oneToOnePerson"].SingleData.Type.Should().Be(type);
+            responseDocument.SingleData.Relationships["oneToOnePerson"].SingleData.Id.Should().Be(_persons[index].StringId);
+        }
+
+        [Theory]
+        [InlineData("males", 0)]
+        [InlineData("females", 1)]
```
bart-degreed, line 72: This looks like testing the same thing twice to me. Assuming the scenario is to show a base class with a derived type works.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
```diff
@@ -0,0 +1,302 @@
...
+            responseDocument.SingleData.Relationships["oneToManyPersons"].ManyData[0].Id.Should().Be(_persons[0].StringId);
+            responseDocument.SingleData.Relationships["oneToManyPersons"].ManyData[1].Type.Should().Be("females");
+            responseDocument.SingleData.Relationships["oneToManyPersons"].ManyData[1].Id.Should().Be(_persons[1].StringId);
+        }
+        
+        [Fact]
+        public async Task Can_patch_one_to_many_relationship_that_has_inheritance_through_relationship_endpoint()
+        {
+            // Arrange            
+
```
bart-degreed, line 167: Please remove this extra empty line.

bart-degreed (changes requested):

bart-degreed:
I would have expected changes in `ResponseDeserializer.ParseIncludedRelationship` too, they were marked as a place we need to update.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Controllers.cs
maurei: In my opinion it only makes it harder to find them because controllers, models and test files are all mixed up in one folder. I'll separate them over different files but then will put them in children folders instead.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models.cs
maurei: This wasn't so easy to come up with. I tried to eg. elaborate on the teacher/student/article example from [the repro repository](https://github.com/bart-degreed/DeserializeIssue/tree/upgrade-netcore31-jadnc4beta1). It has an intuitive one-to-many relationship. But when trying to expand t with a one-to-one and many-to-many relationship that are all conceptually different and still intuitive, I kept ending up having weird relationships in examples that no longer fully made sense. Having only partially intuitive examples, I prefer having an abstract one. If you have a fully intuitive examples with the three relationships and using inheritance, I would be happy to include it

src/Examples/JsonApiDotNetCoreExample/Controllers/KebabCasedModelsController.cs
```diff
@@ -6,13 +6,13 @@
...
-            : base(options, loggerFactory, resourceService)
-        { }
+    public sealed class KebabCasedModelsController : JsonApiController<KebabCasedModel>	
+    {	
+        public KebabCasedModelsController(	
+            IJsonApiOptions options,	
+            ILoggerFactory loggerFactory,	
+            IResourceService<KebabCasedModel> resourceService)	
+            : base(options, loggerFactory, resourceService)	
+        { }
```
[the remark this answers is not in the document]
maurei: 👍

src/JsonApiDotNetCore/Repositories/EntityFrameworkCoreRepository.cs
```diff
@@ -264,15 +264,18 @@ private IEnumerable GetTrackedManyRelationshipValue(IEnumerable<IIdentifiable> r
...
+            
             var trackedPointerCollection = TypeHelper.CopyToTypedCollection(relationshipValues.Select(pointer =>
             {
-                // convert each element in the value list to relationshipAttr.DependentType.
                 var tracked = AttachOrGetTracked(pointer);
                 if (tracked != null) newWasAlreadyAttached = true;
-                return Convert.ChangeType(tracked ?? pointer, relationshipAttr.RightType);
+                
+                // we should recalculate the target type for every iteration because types may vary. This is possible with resource inheritance
+                return Convert.ChangeType(tracked ?? pointer, (tracked ?? pointer).GetType());
```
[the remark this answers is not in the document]
maurei: 👍

src/JsonApiDotNetCore/Serialization/BaseDeserializer.cs
maurei: To me `relationshipType` is unclear because a "relationship" refers to the coupling between two resources, eg _person_ has many articles_. But what we're after here is the _resource type_ of the RHS of the relationship, which is embedded in but not the same as `relationshipType`

src/JsonApiDotNetCore/Serialization/BaseDeserializer.cs
maurei: I think you missed the changes on lines 242-243 of this file

src/JsonApiDotNetCore/Serialization/BaseDeserializer.cs
maurei: 👍

src/JsonApiDotNetCore/Serialization/BaseDeserializer.cs
maurei: But I see now I've been inconsistent in this variable naming, I've used `relationshipType` elsewhere. I'm going to update it to `relationshipType`

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models.cs
maurei: 👍

src/JsonApiDotNetCore/Repositories/EntityFrameworkCoreRepository.cs
```diff
@@ -264,15 +264,18 @@ private IEnumerable GetTrackedManyRelationshipValue(IEnumerable<IIdentifiable> r
...
             bool newWasAlreadyAttached = false;
+            
             var trackedPointerCollection = TypeHelper.CopyToTypedCollection(relationshipValues.Select(pointer =>
             {
-                // convert each element in the value list to relationshipAttr.DependentType.
                 var tracked = AttachOrGetTracked(pointer);
                 if (tracked != null) newWasAlreadyAttached = true;
-                return Convert.ChangeType(tracked ?? pointer, relationshipAttr.RightType);
+                
+                // we should recalculate the target type for every iteration because types may vary. This is possible with resource inheritance
```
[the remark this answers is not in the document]
maurei: 👍

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
maurei: We discussed this privately. Refactored tests

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
maurei: I've refactored the tests significantly and decided to stick with the `Can_....` format. Let me know if you still think the `When...` pattern in the current form of the tests

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models.cs
maurei: We discussed this in private. Tests have been refactored.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models.cs
maurei: Removed it

maurei:
> I would have expected changes in `ResponseDeserializer.ParseIncludedRelationship` too, it was marked as a place that needs to be updated in #828.

Thanks for the heads up. Has been addressed now too.

bart-degreed (changes requested):
Will look at tests later.

src/JsonApiDotNetCore/Serialization/Client/Internal/ResponseDeserializer.cs
```diff
@@ -87,19 +87,20 @@ protected override void AfterProcessField(IIdentifiable resource, ResourceFieldA
...
                 return relatedInstance;
 
-            var resourceContext = ResourceContextProvider.GetResourceContext(relatedResourceIdentifier.Type);
-            if (resourceContext == null)
+            if (relatedResourceContext == null)
                 throw new InvalidOperationException($"Included type '{relationshipAttr.RightType}' is not a registered json:api resource.");
 
-            SetAttributes(relatedInstance, includedResource.Attributes, resourceContext.Attributes);
-            SetRelationships(relatedInstance, includedResource.Relationships, resourceContext.Relationships);
+            SetAttributes(relatedInstance, includedResource.Attributes, relatedResourceContext.Attributes);
```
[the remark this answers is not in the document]
maurei: Good catch!

src/JsonApiDotNetCore/Resources/ResourceFactory.cs
```diff
@@ -116,6 +116,11 @@ private static ConstructorInfo GetLongestConstructor(Type type)
         {
             ConstructorInfo[] constructors = type.GetConstructors().Where(c => !c.IsStatic).ToArray();
 
+            if (constructors.Length == 0)
+            {
+                throw new InvalidOperationException($"No public constructors found for '{type.FullName}'.");
```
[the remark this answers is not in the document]
maurei: 👍

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
maurei: Removed

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
```diff
@@ -0,0 +1,315 @@
...
+            });
+        }
+
+
+        [Fact]
+        public async Task Can_patch_resource_with_to_one_relationship_through_relationship_link()
+        {
+            // Arrange
+            var person = new Male();
+            var cat = new Cat();
```
bart-degreed, line 91: Did you mean `var man`?

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
```diff
@@ -0,0 +1,315 @@
...
+
+            await _testContext.RunOnDatabaseAsync(async dbContext =>
+            {
+                await dbContext.AddRangeAsync(person, cat);
```
bart-degreed, line 95: I'm not sure what's better. Calling `dbContext.AddRange()` enables passing different entity types is a single call. That's short to write, be at the cost of losing static typing (you can pass anything, such as `DateTime`). That's basically trading compile-time errors for run-time errors. I think I prefer using the typed members, such as `dbContext.Animals.AddRange()`, which better reveals intent (and is more efficient in theory, but likely not measurable).

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
```diff
@@ -0,0 +1,315 @@
...
+            var (httpResponse, responseDocument) = await _testContext.ExecutePostAsync<Document>(route, requestBody);
+
+            // Assert
+            httpResponse.Should().HaveStatusCode(HttpStatusCode.Created);
+
+            await _testContext.RunOnDatabaseAsync(async dbContext =>
+            {
+                var assertPerson = await dbContext.People
+                    .Include(p => p.Pet)
+                    .Where(p => p.Id.Equals(int.Parse(responseDocument.SingleData.Id))).FirstAsync();
```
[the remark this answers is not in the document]
bart-degreed: In general, calling `something.Equals` should be avoided when `operator==` is implemented. It is more readable, but more importantly does not throw when `something` is `null`.

test/UnitTests/Serialization/SerializationTestsSetupBase.cs
```diff
@@ -1,7 +1,9 @@
 using Bogus;
 using JsonApiDotNetCore.Configuration;
+using JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance;
 using Microsoft.Extensions.Logging.Abstractions;
 using UnitTests.TestModels;
+using InheritancePerson = JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance.Person;
```
bart-degreed, line 6: I find it very confusing that we have two incompatible definitions of Person *and* that we need both of them in the same file. Better would be to rename ResourceInheritance.Person to Human so they are really distinct.

test/UnitTests/UnitTests.csproj
```diff
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Examples\JsonApiDotNetCoreExample\JsonApiDotNetCoreExample.csproj" />
+    <ProjectReference Include="..\JsonApiDotNetCoreExampleTests\JsonApiDotNetCoreExampleTests.csproj" />
```
bart-degreed, line 14: I think this dependency is wrong. Unit-tests should not depend on types defined in another test project. It is only a matter of time before we need circular dependencies if we go down that path.

test/UnitTests/Serialization/Client/ResponseDeserializerTests.cs
```diff
@@ -335,5 +334,59 @@ public void DeserializeList_DeeplyNestedIncluded_CanDeserialize()
...
+                {
+                    Type = "cats",
+                    Id = "20",
+                    Attributes = new Dictionary<string, object> { {"feline", "true" }, { "scaredOfDogs", "true" } }
+                },
+            };
+            var body = JsonConvert.SerializeObject(content);
+
+            // Act
+            var result = _deserializer.DeserializeSingle<Male>(body);
```
bart-degreed, line 370: This verifies deserialize behavior, but what if `DeserializeSingle` is called with `<Person>`? My point is that this does not prove much. So I'd prefer to have an integration test instead, with has an assertion on the json text output (see `Registered_IResponseMeta_Adds_TopLevel_Meta` for example).

For background why I believe we should have mostly integration tests, see https://kentcdodds.com/blog/write-tests and https://tyrrrz.me/blog/unit-testing-is-overrated.

bart-degreed (changes requested):

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/DocumentTests/LinksWithoutNamespaceTests.cs
```diff
@@ -35,24 +32,24 @@ public async Task GET_RelativeLinks_True_Without_Namespace_Returns_RelativeLinks
             var options = (JsonApiOptions) _testContext.Factory.Services.GetRequiredService<IJsonApiOptions>();
             options.UseRelativeLinks = true;
 
-            var person = new Person();
+            var blog = new Blog();
```
bart-degreed, line 35: Why replace Person with Blog here? Does it have to do with namespace clashes?

bart-degreed (changes requested):

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/Animal.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance
+{
+    public abstract class Animal : Identifiable
+    {
+        [Attr]
+        public bool Feline { get; set; } 
+        
+        [Attr]
+        public bool IsDomesticated { get; set; } 
+    }
```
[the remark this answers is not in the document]
RichardsonWTR: What about Animal/Dog/Bird? 
Easy to understand and realistic at the same time

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/Content.cs
```diff
@@ -0,0 +1,6 @@
+using JsonApiDotNetCore.Resources;
+
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance
```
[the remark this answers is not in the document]
maurei: Will fix

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/Content.cs
```diff
@@ -0,0 +1,6 @@
...
+{
+    public abstract class Content : Identifiable { }
+}
```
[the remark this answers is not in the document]
maurei: I agree on `ContentItem` vs `Content`.

I don't agree that adding extra properties is useful. I think the goal is to make example cases in the tests more intuitive by choosing a set of models that have an understandable inheritance structure rather than having classes like `BaseResource`, `DerivedResource1`, `DerivedResource2`. For this purpose having `ContentItem`, `Video`, and `Book` is sufficient. Only relationship properties are assigned in the tests, not attribute properties, so I don't think adding any unused attribute properties will be of any value or make the tests anymore intuitive. It only adds unused clutter.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/ContentPerson.cs
```diff
@@ -0,0 +1,13 @@
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance
+{
+    public sealed class ContentPerson
```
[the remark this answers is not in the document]
maurei: I don't think so.  It is customary to name a jointable entity like this, see [EF cores standard example](https://docs.microsoft.com/en-us/ef/core/modeling/relationships?tabs=fluent-api%2Cfluent-api-simple-key%2Csimple-key#many-to-many). It also doesn't make sense to name it `PersonFavoriteContentItem` because this type could be reused for other many-to-many relationships between `ContentItem` and `Person`.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/Animal.cs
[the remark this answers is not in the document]
bart-degreed: I'd prefer something that developers actually do model, I've never seen the need to model animals in applications. More realistic, for example, would be BankAccount { Amount } that can be a CheckingAccount { AtmWithdrawalLimit} or SavingsAccount { InterestRate }.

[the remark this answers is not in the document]
maurei: I think what is required for intuitive understanding when reading the tests is an intuitive inheritance structure. `Animal` with derived types `Cat` and `Dog` suffice for that. Properties are not important because relationship assignment the only thing that happens in the test, so I think we could even just remove the properties. Either way, coming up with a different set with models with other properties won't make the test examples any more intuitive, and I don't think a change here is necessary.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
```diff
@@ -0,0 +1,315 @@
...
+using Xunit;
+
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance
+{
+    public sealed class ResourceInheritanceTests : IClassFixture<IntegrationTestContext<
+        TestableStartup<ResourceInheritanceDbContext>, ResourceInheritanceDbContext>>
+    {
+        private readonly
+            IntegrationTestContext<TestableStartup<ResourceInheritanceDbContext>, ResourceInheritanceDbContext>
+            _testContext;
```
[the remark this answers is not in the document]
maurei: I don't think `Inheritance` quite covers it. Inheritance of what? It could be anything.

test/UnitTests/UnitTests.csproj
maurei: The alternative would be to copy paste the models to the unit test project, I don't like that. It would be solving for something that is currently not even an issue (circular dependencies). If it would ever hit us, we can choose to create a separate class project for our test models.

test/UnitTests/Serialization/SerializationTestsSetupBase.cs
maurei: I agree, will improve

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
maurei: Yep

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/Content.cs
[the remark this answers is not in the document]
maurei: Will change to `JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance.Models`

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
```diff
@@ -0,0 +1,315 @@
...
+
+            _testContext.RunOnDatabaseAsync(async dbContext =>
+            {
+                await dbContext.ClearTableAsync<ContentPerson>();
+                await dbContext.ClearTableAsync<Book>();
+                await dbContext.ClearTableAsync<Video>();
+                await dbContext.ClearTableAsync<Cat>();
+                await dbContext.ClearTableAsync<Dog>();
+                await dbContext.ClearTableAsync<Female>();
+                await dbContext.ClearTableAsync<Male>();
```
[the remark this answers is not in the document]
maurei: > Effectively the tables are purged before the first test executes
This is not true, it is executed before every test in the encapsulating test. 

So it is just a clean up that executes before every test, which is desired. Putting it in the constructor prevents code duplication, which is not a bad practise.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
```diff
@@ -0,0 +1,315 @@
...
+        
+        [Fact]
+        public async Task Can_create_resource_with_to_one_relationship()
+        {
+            // Arrange
+            var cat = new Cat();
+            
+            await _testContext.RunOnDatabaseAsync(async dbContext =>
+            {
+                await dbContext.AddAsync(cat);
```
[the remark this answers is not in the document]
maurei: Will adjust

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
maurei: Will change

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
maurei: No. It's not about the `person`, its about the `cat` being assigned. Could have been a `Female` as well. It could be changed to `man` or `male` but I don't think it makes much of a difference and I don't think it requires a change.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
```diff
@@ -0,0 +1,315 @@
...
+            // Assert
+            httpResponse.Should().HaveStatusCode(HttpStatusCode.Created);
+
+            await _testContext.RunOnDatabaseAsync(async dbContext =>
+            {
+                var assertPerson = await dbContext.People
+                    .Include(p => p.Pet)
+                    .Where(p => p.Id.Equals(int.Parse(responseDocument.SingleData.Id))).FirstAsync();
+                
+                assertPerson.Pet.GetType().Should().Be(cat.GetType());
```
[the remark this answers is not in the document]
maurei: Will add

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
```diff
@@ -0,0 +1,315 @@
...
+            });
+        }
```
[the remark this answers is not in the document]
maurei: Will add

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
maurei: I agree that it in general it is more safe to just the strongly typed version. However this overload also allows for more brevity, which is nice when trying to create readable tests. Safety risks or unclear intent or are not much of a concern here because the method call and the associated model declarations are very closely together.

test/UnitTests/Serialization/Client/ResponseDeserializerTests.cs
maurei: Calling `DeserializeSingle<Person>(body)` is not possible because the type parameter must be a reference type. 

It would be different if this method call is stuffed away in other code (eg a client library) which calls this method reflectively and which could break at runtime if `Person` was supplied. In this scenario a integration would be useful, but there currently is no encapsulating code like that.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
bart-degreed: Are you sure? https://xunit.net/docs/shared-context

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
bart-degreed: My point exactly: this test is about the cat. So declaring a male, named person, leads to believe something related to inheritance for person/man is going on (but it is not relevant at all), distracting from the purpose of this test.

Thinking about this: wouldn't it be more correct to have Man and Woman instead of Male and Female?

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/DocumentTests/LinksWithoutNamespaceTests.cs
maurei: There tests that failed in the example test project because these tests added the example project assembly ([like this here](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/10dd16a00f1b28bb96a495a710c114e48dea5507/test/JsonApiDotNetCoreExampleTests/Acceptance/KebabCaseFormatterTests.cs#L38) and my PR introduces a `PeopleController` in the example project. This leads to the routing convention throwing an error because tries to register the same route twice. `Blog` rather than `Person` can be used equally well for `LinksWithoutNamespaceTests` while it fixes this problem.

test/UnitTests/Serialization/Client/ResponseDeserializerTests.cs
bart-degreed: `Person` _is_ a reference type, so it is possible. And I don't understand what encapsulating code you're missing that prevents you from writing an integration test. It also resolves the issue of a bad project reference vs duplicating types.

test/UnitTests/UnitTests.csproj
bart-degreed: Having the right dependencies is not just to prevent circular references, it is about expressing layered architecture. Tests should not depend on each other.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
bart-degreed: Makes sense. So what should be our convention then?
1. Always call dbContext.Add/AddRange and never call dbContext.Animals.Add/AddRange
2. Only call dbContext.AddRange for mixed types, in all other cases use dbContext.Animals.Add/AddRange
3. Anything else?

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/Content.cs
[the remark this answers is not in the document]
bart-degreed: I'm not saying the tests should be assigning those properties. But this is about modelling something, not a procedure that contains dead code. To make the model more understandable and available for use, such properties add value. If in the future I wanted to add a test that works on a property, I'd need to invent one first (and possibly break existing tests that never expected the property to exist in the json output).

So the goal should be to expose a model that makes sense to humans and is easy to use. I think that is unrelated to YAGNI, which is about building features that _might_ be needed in the future. Or not, who knows?

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
bart-degreed: What other forms of inheritance are you thinking of that make the term ambiguous here?

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/Animal.cs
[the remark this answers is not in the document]
bart-degreed: Defining a model where its hard to imagine sensible properties does not sound like a realistic model. Having a model that humans can relate to improves readability of tests, now and in the future (when adding filters or fields). Do you think it would be good practice to throw-away existing models and tests, so we can replace them with something appropriate at the time we need to add tests for filters? This is so basic, I think we can (and should) get it right initially.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
maurei: See https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/833#discussion_r498676811: no longer need to check if posting to the `males` endpoint will return a resource of type `males`, that is basic functionality that is tested elsewhere.

For the `type` of the pet, this is not possible because that would require an `include`.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/ContentPerson.cs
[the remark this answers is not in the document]
bart-degreed: The example you're referring to doesn't apply, because there the properties and tables are named the same. This is similar to `Person Owner` vs `Person Person`. The many-to-many table you're using here links favorites, not just any many-to-many relationships that may exist between persons and content. If that were the case, you'd need to add an extra column that indicates the relationship type. But I don't think JADNC even supports that.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
maurei: See https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/833#discussion_r498676811: no longer need to check if posting to the `males` endpoint will return a resource of type `males`, that is basic functionality that is tested elsewhere.

For the `type` of the pet, this is not possible because that would require an `include`.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
maurei: I would argue there is no need for a very strict convention here. Different approaches could be more valuable in different scenarios. But if I would have to choose, I think 2. makes sense

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
maurei: Man and Woman: agreed

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
maurei: Inheritance of any JADNC component

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
maurei: I think so! I put a Console.Writeline statement in the constructor which occurred as many times as there are tests in this class.

test/UnitTests/UnitTests.csproj
maurei: What would be your preferred way of getting rid of the dependency?

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/ContentPerson.cs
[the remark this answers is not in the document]
maurei: It's true that it doesn't work with JADNC but it can still be used like that within the context of EF Core. Probably not a very common use-case in combination with JADNC, but it is just to point out that I don't think it is convention to embed the relationship name in the table/class name. Similarly you wouldn't necessarily rename the `Person` table to `Author` just because there is an author to-one relationship between `Article` and `Person`.

One could change the property name to reflect that:
```c#
// in Person.cs
public List<PersonContentItem> PersonFavoriteContentItems { get; set; }
```

But given that there already is a `FavoriteContent` property, I don't think this adds much value to put `Favorite` in this one too.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/InheritanceTests.cs
```diff
@@ -0,0 +1,310 @@
...
+            {
+                var favoriteContent = (await dbContext.Males
+                        .Include(h => h.HumanFavoriteContentItems)
+                            .ThenInclude(hp => hp.ContentItem)
+                        .SingleAsync(h => h.Id.Equals(man.Id)))
+                    .HumanFavoriteContentItems.Select(hp => hp.ContentItem).ToList();
+        
+                favoriteContent.Should().HaveCount(2);
+                favoriteContent.Should().ContainSingle(h => h is Book);
+                favoriteContent.Should().ContainSingle(h => h is Video);
```
bart-degreed, line 306: More readable:

```c#
var contentItems = await dbContext.HumanFavoriteContentItems
    .Where(x => x.Human.Id == man.Id)
    .Select(x => x.ContentItem)
    .ToListAsync();

contentItems.Should().HaveCount(2);
contentItems.Should().ContainSingle(h => h is Book);
contentItems.Should().ContainSingle(h => h is Video);
```

bart-degreed (changes requested):

src/JsonApiDotNetCore/Serialization/BaseDeserializer.cs
bart-degreed: You're right, I missed that.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
bart-degreed: I think this is okay for now.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/ContentPerson.cs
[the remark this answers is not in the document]
bart-degreed: We've discussed this offline. The many-to-many table is effectively a relationship and not a data store.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
bart-degreed: My intent was to rename the _folder_ to Inheritance, but the way you've changed it now is probably okay too.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
bart-degreed: Yes, you're right. We decided offline to do the clearing per test, but have a method that accepts multiple for brevity.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
bart-degreed: I'm still seeing usage of `AddRangeAsync`.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
bart-degreed: I'm still seeing unneeded `.Equals` usage in the tests.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
bart-degreed: Sounds good.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
bart-degreed: And I'm inferring we don't want to add the include, because then we'd make assumptions on how things work that we don't want to make at the moment, correct?

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
bart-degreed: Okay, let's pick 2 then. Please update tests to match that convention.

test/UnitTests/UnitTests.csproj
bart-degreed: We've discussed this offline and decided to have unit tests define their own models to run against. Since they are usually very specific and intent to hit all kinds of combinations, having humanly understandable models there is less of a concern.

test/UnitTests/Serialization/Client/ResponseDeserializerTests.cs
bart-degreed: I assumed this was testing basic serialization behavior (used by both client and server), but it turns out to be testing client-specific serialization logic. So it makes no sense to write an integration test.

test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/DocumentTests/LinksWithoutNamespaceTests.cs
bart-degreed: This problem no longer applies and so these changes can be reverted.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
maurei: That wouldn't decrease the length of this particular line, unless you wanted to rename that too

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
maurei: correct

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/InheritanceTests.cs
maurei: Applied

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/InheritanceTests.cs
```diff
@@ -0,0 +1,310 @@
...
+            // Assert
+            httpResponse.Should().HaveStatusCode(HttpStatusCode.Created);
+
+            await _testContext.RunOnDatabaseAsync(async dbContext =>
+            {
+                var assertMan = await dbContext.Males
+                    .Include(h => h.HealthInsurance)
+                    .SingleAsync(h => h.Id == int.Parse(responseDocument.SingleData.Id));
+                
+                assertMan.HealthInsurance.GetType().Should().Be(insurance.GetType());
```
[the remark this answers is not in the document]
maurei: Applied

test/JsonApiDotNetCoreExampleTests/AppDbContextExtensions.cs
```diff
@@ -9,23 +9,52 @@ public static class AppDbContextExtensions
     {
         public static async Task ClearTableAsync<TEntity>(this DbContext dbContext) where TEntity : class
         {
-            var entityType = dbContext.Model.FindEntityType(typeof(TEntity));
-            if (entityType == null)
+            await ClearTableAsync(dbContext,typeof(TEntity));
+        }
+        
+        public static async Task ClearTableAsync<TEntity1, TEntity2>(this DbContext dbContext) where TEntity1 : class
```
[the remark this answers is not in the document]
maurei: Applied

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/InheritanceDbContext.cs
```diff
@@ -0,0 +1,36 @@
+using JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance.Models;
+using Microsoft.EntityFrameworkCore;
+
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance
+{
+    public sealed class InheritanceDbContext : DbContext
+    {
+        public InheritanceDbContext(DbContextOptions<InheritanceDbContext> options) : base(options) { }
+        
+        public DbSet<Man> Males { get; set; }
```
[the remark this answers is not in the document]
maurei: Applied

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/InheritanceDbContext.cs
```diff
@@ -0,0 +1,36 @@
...
+                .HasValue<FamilyHealthInsurance>(2);
+            
+            modelBuilder.Entity<ContentItem>()
+                .ToTable("Content")
+                .HasDiscriminator<int>("Type")
+                .HasValue<Video>(1)
+                .HasValue<Book>(2);
+            
+            modelBuilder.Entity<HumanFavoriteContentItem>()
+                .HasKey(pp => new { ContentPersonId = pp.ContentItemId, PersonId = pp.HumanId });
```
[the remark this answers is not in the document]
maurei: Hadn't updated it, changed it to abbrevation of HumanFavoriteContentItem

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/CompanyHealthInsurance.cs
```diff
@@ -0,0 +1,10 @@
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance.Models
+{
+    public class CompanyHealthInsurance : HealthInsurance
```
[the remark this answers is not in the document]
maurei: Applied

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/InheritanceDbContext.cs
```diff
@@ -0,0 +1,36 @@
...
+    public sealed class InheritanceDbContext : DbContext
+    {
+        public InheritanceDbContext(DbContextOptions<InheritanceDbContext> options) : base(options) { }
+        
+        public DbSet<Man> Males { get; set; }
+        
+        protected override void OnModelCreating(ModelBuilder modelBuilder)
+        {
+            modelBuilder.Entity<Human>()
+                .ToTable("Persons")
```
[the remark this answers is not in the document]
maurei: Applied

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/FamilyHealthInsurance.cs
```diff
@@ -0,0 +1,10 @@
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance.Models
+{
+    public class FamilyHealthInsurance : HealthInsurance
```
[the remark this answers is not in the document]
maurei: Applied

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/Human.cs
```diff
@@ -0,0 +1,29 @@
...
+        [HasManyThrough(nameof(HumanFavoriteContentItems))]
+        public List<ContentItem> FavoriteContent { get; set; }
+        
+        public List<HumanFavoriteContentItem> HumanFavoriteContentItems { get; set; }
+    }
+}
+    
+
+    
+
```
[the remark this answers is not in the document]
maurei: Applied

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/Human.cs
```diff
@@ -0,0 +1,29 @@
...
+        public HealthInsurance HealthInsurance { get; set; }
+        
+        [HasMany]
+        public List<Human> Parents { get; set; }
+        
+        [NotMapped]
+        [HasManyThrough(nameof(HumanFavoriteContentItems))]
+        public List<ContentItem> FavoriteContent { get; set; }
+        
+        public List<HumanFavoriteContentItem> HumanFavoriteContentItems { get; set; }
```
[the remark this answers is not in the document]
maurei: ICollection suffices for the current tests. Applied.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/Models/Video.cs
```diff
@@ -0,0 +1,10 @@
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.ResourceInheritance.Models
+{
+    public class Video : ContentItem
```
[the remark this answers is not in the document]
maurei: Applied

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
maurei: Applied

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
maurei: Not anymore

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/ResourceInheritanceTests.cs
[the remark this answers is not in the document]
maurei: Not anymore

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/InheritanceDbContext.cs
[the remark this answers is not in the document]
maurei: Also double checked all other parameter names in lambda methods for consistent abbreviations

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceInheritance/InheritanceTests.cs
maurei: This amazes me by the way! EF Core is smart enough to throw in an `Include` under the hood huh?

bart-degreed: Yes, due to the select. This is functionally equivalent to an Include, otherwise the select would never work.

bart-degreed (approved):

PR #836, opened 2020-09-22
Fixed support for using multiple DbContexts

### LINKED ISSUE #739: Injecting multiple DbContexts does not work

bart-degreed:
Fixes #739.

Added example project with tests, as this is tricky to get right (and not accidentally break in the future).

In the past, this was accomplished by calling `ResourceGraphBuilder.AddDbContext()` from `services.AddJsonApi()`. I've changed that to instead accept a collection of `DbContext` types in the non-generic overload of `services.AddJsonApi()`, so that we can also register matching resolvers.

This also fixes `InverseNavigation` to loop over all `DbContexts` instead of a single one.

And for this to work, I found another bug where `ResourceService<TResource>` was depending on injected `IResourceRepository<TResource,TId>` instead of `IResourceRepository<TResource>`, so it would pick the wrong repository.

Finally, I simplified caching of scanned assemblies that was spread over multiple layers doing duplicate work.


maurei (changes requested):
I am aware this PR only fixes support for multiple DbContexts rather than introducing it as a new feature or improving it as a feature. But regardless, perhaps we can consider improving the follow.

I think it is unnecessarily verbose having to both pass along `DbContextA`, `DbContextB` to `AddJsonApi()` call and also having to register a `IDbContextResolver` implementation for each.

We could register a generic resolver service `DbContextResolver<TDbContext>` like this

```c#
public class DbContextResolver<TDbContext> : IDbContextResolver where TDbContext : DbContext
{
    private readonly DbContext _context;

    public DbContextAResolver(TDbContext context)
    {
        _context = context;
    }

    public DbContext GetContext()
    {
        return _context;
    }
}
```

Then a dev can just inject `DbContextResolver<DbContextA>` in his custom repository.

Alternatively, if we do not want to offer this solution out-of-the-box but want to leave the responsibility with the developer, I would propose to change the example using this rather approach than two separate DbContextResolver services. It is cleaner and as such easier to follow.

Other than this I have nothing to comment on the PR. Looking good 👍

bart-degreed:
> I am aware this PR only fixes support for multiple DbContexts rather than introducing it as a new feature or improving it as a feature. But regardless, perhaps we can consider improving the follow.

Great suggestion! I've pushed another commit to address this.

maurei (approved):

PR #845, opened 2020-09-28
Resource-specific meta

### LINKED ISSUE #737: IHasMeta does not work

bart-degreed:
Fixes #737.


src/JsonApiDotNetCore/Serialization/Building/IResourceObjectBuilder.cs
```diff
@@ -16,9 +16,10 @@ public interface IResourceObjectBuilder
         /// Adds the attributes and relationships that are enlisted in <paramref name="attributes"/> and <paramref name="relationships"/>.
         /// </summary>
         /// <param name="resource">Resource to build a <see cref="ResourceObject"/> for.</param>
+        /// <param name="includeResourceMeta">When <c>true</c>, retrieves resource-specific meta from <see cref="IResourceDefinition{TResource}"/>.</param>
         /// <param name="attributes">Attributes to include in the building process.</param>
         /// <param name="relationships">Relationships to include in the building process.</param>
         /// <returns>The resource object that was built.</returns>
-        ResourceObject Build(IIdentifiable resource, IReadOnlyCollection<AttrAttribute> attributes, IReadOnlyCollection<RelationshipAttribute> relationships);
+        ResourceObject Build(IIdentifiable resource, bool includeResourceMeta, IReadOnlyCollection<AttrAttribute> attributes, IReadOnlyCollection<RelationshipAttribute> relationships);
```
maurei, line 23: I think it would be favorable to implement resource-level metadata without having it ripple through to the public API of `ResourceObjectBuilder`. The metadata can be assigned in the `ResponseSerializer` class

maurei (approved):

src/JsonApiDotNetCore/Serialization/Building/IResourceObjectBuilder.cs
bart-degreed: This is not possible, because `ResourceDefinition.GetMeta` requires a resource instance, not a `ResourceObject`. After offline discussion, we found that the best solution is to replace the boolean flag with an overridden Build method in `ResourceObjectBuilder`.

maurei: We discussed this in a meeting and found a way around it.

maurei (approved):

maurei:
👍

PR #851, opened 2020-10-04
Relationship endpoint bugfixes, implement POST and DELETE on to-many relationship link

### LINKED ISSUE #353: Support POST requests to a to-many relationship link
As per the json:api spec (under the section linked at http://jsonapi.org/format/#crud-updating-relationships)

> A server MUST respond to PATCH, POST, and DELETE requests to a URL from a to-many relationship link
    ...
   If a client makes a POST request to a URL from a relationship link, the server MUST add the specified 
   members to the relationship unless they are already present. If a given type and id is already in the 
   relationship, the server MUST NOT add it again

### LINKED ISSUE #578: Response when updating relationship of resource not conform spec
This issue is analogous to #577 but for relationships instead of attributes.

> 200 OK
> **If a server accepts an update but also changes the targeted relationship(s) in other ways than those specified by the request, it MUST return a 200 OK response. The response document MUST include a representation of the updated relationship(s).**
> 
> A server MUST return a 200 OK status code if an update is successful, the client’s current data remain up to date, and the server responds only with top-level meta data. In this case the server MUST NOT include a representation of the updated relationship(s).

the part in **bold** is currently not properly implemented.

The associated repository method does not return anything:

```c#
public interface IEntityWriteRepository<TEntity, in TId>
{
  ....
  Task UpdateRelationshipsAsync(object parent, RelationshipAttribute relationship, IEnumerable<string> relationshipIds);
  ....
}


```

This is fine for the default implementation. But if, as in #577, a developer has business logic that affects the relationships in a way different than as specified by the request, a response **must** include a representation of the updated relationships

### LINKED ISSUE #850: Support DELETE requests to a to-many relationship link
While diving into #353, I realised that DELETE on `/articles/1/toManyRelationship` is also not implemented. See https://jsonapi.org/format/#crud-updating-relationships.

### LINKED ISSUE #852: Patching  resource relationship on primary endpoint with non-existing resource fails
When doing a `PATCH /articles/{id}` and updating the `author` relationship with a reference to a non-existing `person` resource, it fails with a 500. No check is performed if the relationship actually exists.

[Should be a 404 with a descriptive message](https://jsonapi.org/format/#crud-updating-responses-404).

[See this test](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/1b862c980dae42f0564d2028c62fe9d2048dd2c0/test/JsonApiDotNetCoreExampleTests/Acceptance/Spec/UpdatingRelationshipsTests.cs#L220)

### LINKED ISSUE #853: Implicit removes and FK constraint violations on PATCH relationship endpoint
The mechanisms that have been introduced in #502 to fix various FK constraint violations for the `PATCH /articles` pipeline do not exist for `PATCH /articles/1/relationships/author`

### LINKED ISSUE #870: PATCH Successfully but Return Data is Null
## Description
I tried to PATCH the _department_ resource. Request body as below:

```
PATCH /departments/10

{
    "data": {
        "type": "departments",
        "id": "10",
        "attributes": {
            "name": "Department 10",
            "description": ""
        }
    }
}
```

The data has been changed successfully, but the response data received is: null.

```
{
    "links": {
        "self": "http://localhost:5000/departments/10"
    },
    "data": null
}
```

The response should have received changed data right?
...

## Environment

- JsonApiDotNetCore Version: 4.0.0-beta1

maurei:
This PR concerns write operations and addresses several topics:

- Add POST/DELETE relationship endpoints, which enable to add/remove resources to/from a to-many relationship.
- Redesign of handling updates, where we try to fetch the minimum data required to perform the operation. If the operation fails, we run additional queries to determine why and produce a descriptive error according to json:api requirements.
- Rely more on the EF Core navigation metadata model, instead of dealing with foreign keys directly.
- Apply the whole operation in a single `SaveChanges()` call, removing the need for transaction management.
- Add resource change tracking for POST, similar to PATCH. Note this only has an effect when using client-generated IDs.
- Refactoring of tests to hit all combinations of create/update/delete for the various endpoints and relationship kinds.
- Additional tests for corner cases like malformed request bodies, foreign key constraint violations, implicit removal in one-to-one relationships etc.

Fixes #353
Fixes #850 
Fixes #852
Fixes #853
Fixes #578
Fixes #870


src/JsonApiDotNetCore/Errors/QueryExecutionException.cs
```diff
@@ -0,0 +1,16 @@
...
+using JsonApiDotNetCore.Queries;
+using JsonApiDotNetCore.Serialization.Objects;
+using Newtonsoft.Json;
+
+namespace JsonApiDotNetCore.Errors
+{
+    /// <summary>
+    /// The error that is thrown when Entity Framework Core fails executing a query.
+    /// </summary>
+    public sealed class QueryExecutionException : Exception
```
bart-degreed, line 12: - This should not be in the Errors namespace (see https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/808)
- The point of having our own exception (instead of reusing `DbUpdateException`) is to separate resource services from EF Core implementation, thus enabling alternative data stores. So the documentation should reflect that.
- The name `QueryExecutionException` is too broad, it is not just any query (read or write) that fails. I'm thinking `DataStoreUpdateException` is more appropriate.

src/JsonApiDotNetCore/Errors/QueryExecutionException.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        public QueryExecutionException(Exception exception) : base(exception.Message, exception) { }
```
bart-degreed, line 14: The point of wrapping exceptions is to create an abstraction, so passing the inner message unmodified does not feel right. Assuming we have no need for callers to pass a specific message explaining what was going on, then we should use a standard text here, such as "Failed to apply changes in the underlying data store.". For those authorized, the call stack contains details on what happened.

src/JsonApiDotNetCore/Repositories/EntityFrameworkCoreRepository.cs
```diff
@@ -135,73 +131,44 @@ public virtual async Task CreateAsync(TResource resource)
...
-                else
-                    relationEntry.Collection(hasOneAttr.InverseNavigation).Load();
+                await _dbContext.SaveChangesAsync();
             }
-            else if (relationshipAttr is HasManyAttribute hasManyAttr && !(relationshipAttr is HasManyThroughAttribute))
+            catch (DbUpdateException exception)
             {
-                foreach (IIdentifiable relationshipValue in (IEnumerable)trackedRelationshipValue)
-                    _dbContext.Entry(relationshipValue).Reference(hasManyAttr.InverseNavigation).Load();
+                throw new QueryExecutionException(exception);
```
bart-degreed, line 150: All except one usage of `_dbContext.SaveChangesAsync()` is wrapped in the same try/catch. Assuming that's a bug, better extract this into a private method.

src/JsonApiDotNetCore/Errors/RelationshipUpdateForbiddenException.cs
```diff
@@ -0,0 +1,18 @@
+using System.Net;
+using JsonApiDotNetCore.Serialization.Objects;
+
+namespace JsonApiDotNetCore.Errors
+{
+    /// <summary>
+    /// The error that is thrown when a request is received that contains an unsupported HTTP verb.
+    /// </summary>
```
bart-degreed, line 8: Documentation is incorrect.

The json::api spec reads:
```
A server MUST return 403 Forbidden in response to an unsupported request to update a relationship.
```

So a better title would be: "Unsupported request to update relationship."

But looking at the usage, this exception is always thrown for a more specific scenario, which makes me think the next title is better: "Only to-many relationships can be updated through this endpoint."

And the `{toOneRelationship}` in Detail should be enclosed in single quotes, similar to other exceptions.

And the parameter `toOneRelationship` should be renamed to `relationshipName`. The exception is about to-manies and does not need to know what other kinds of relationships do exist and how they relate to each other.

src/JsonApiDotNetCore/Services/JsonApiResourceService.cs
```diff
@@ -307,45 +324,166 @@ public virtual async Task<TResource> UpdateAsync(TId id, TResource requestResour
...
+                var resource = await GetProjectedPrimaryResourceById(id);
+                AssertPrimaryResourceExists(resource);
+                
+                throw;
+            }
+        }
 
-            string[] relationshipIds = null;
-            if (relationships != null)
+        private readonly struct RelationshipAssignment
```
bart-degreed, line 427: Don't use structs, they have pitfalls and are tricky to get right. Only for extreme performance optimizations should they be considered. And don't declare nested types in the middle, but always at the end of the containing type.

src/JsonApiDotNetCore/Services/JsonApiResourceService.cs
```diff
@@ -365,7 +503,51 @@ private void AssertRelationshipExists(string relationshipName)
...
+            foreach (var assignment in nonNullRelationshipAssignments)
+            {
+                IReadOnlyCollection<string> identifiers;
+                if (assignment.Value is IIdentifiable identifiable)
+                {
+                    identifiers = new [] { TypeHelper.GetResourceStringId(identifiable) };
+                }
+                else
+                {
+                    identifiers = ((IReadOnlyCollection<IIdentifiable>) assignment.Value).Select(TypeHelper.GetResourceStringId).ToArray();
```
bart-degreed, line 527: This cast fails on any other collection type, which is completely unneeded because you're calling `.Select` on it. Instead cast to `IEnumerable<IIdentifiable>`, if you must cast.

src/JsonApiDotNetCore/TypeHelper.cs
```diff
@@ -290,12 +291,37 @@ public static object ConvertStringIdToTypedId(Type resourceType, string stringId
...
         }
+        
+        /// <summary>
+        /// Gets the string value of the id of an identifiable through the typed value. This can be useful when dealing with
+        /// obfuscated IDs.
+        /// </summary>
+        public static string GetResourceStringId(IIdentifiable identifiable)
+        {
+            return GetResourceTypedId(identifiable).ToString();
+        }
```
bart-degreed, line 302: This is a confusing method that should not exist. It does not return the StringId of a resource, but calls .ToString() on the typed ID of a resource. Please inline all usages and remove.

src/JsonApiDotNetCore/TypeHelper.cs
```diff
@@ -290,12 +291,37 @@ public static object ConvertStringIdToTypedId(Type resourceType, string stringId
...
 
+        /// <summary>
+        /// Gets the typed value of the id of an identifiable.
+        /// </summary>
         public static object GetResourceTypedId(IIdentifiable resource)
         {
+            if (resource == null) throw new ArgumentNullException(nameof(resource));
```
bart-degreed, line 309: This method is internal and should not have a null check. Also the added doc-comment does not add any helpful info so should not be there.

src/JsonApiDotNetCore/TypeHelper.cs
```diff
@@ -290,12 +291,37 @@ public static object ConvertStringIdToTypedId(Type resourceType, string stringId
...
+        /// <summary>
+        /// Sets the typed value of the id of an identifiable.
+        /// </summary>
+        public static void SetResourceTypedId(IIdentifiable identifiable, object id)
+        {
+            if (identifiable == null) throw new ArgumentNullException(nameof(identifiable));
+            if (id == null) throw new ArgumentNullException(nameof(id));
+            
+            identifiable.GetType().GetProperty(nameof(Identifiable.Id)).SetValue(identifiable, id);
+        }
```
bart-degreed, line 324: This method is used only once, so should be inlined. See earlier remark on how Helpers are an anti-pattern that we should try to avoid.

src/JsonApiDotNetCore/Repositories/DbContextExtensions.cs
```diff
@@ -16,18 +16,35 @@ public static class DbContextExtensions
...
                     ((IIdentifiable) entry.Entity).StringId == entity.StringId);
 
             return (TEntity) entityEntry?.Entity;
         }
 
+        internal static object GetTrackedOrAttachCurrent(this DbContext context, IIdentifiable entity)
+        {
+           var trackedEntity = context.GetTrackedEntity(entity);
+            if (trackedEntity == null)
+            {
```
bart-degreed, line 35: strange indents

bart-degreed (changes requested):
Holding off further review until clean work has been done.

src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs
```diff
@@ -211,6 +236,20 @@ public virtual async Task<IActionResult> DeleteAsync(TId id)
 
             return NoContent();
         }
+
+        /// <summary>
+        /// Removes resources from a to-many relationship.
+        /// </summary>
+        public virtual async Task<IActionResult> DeleteRelationshipAsync(TId id, string relationshipName, [FromBody] IReadOnlyCollection<IIdentifiable> removalValues)
```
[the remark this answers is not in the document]
maurei: I don't think this is an improvement. We're not removing resource identifiers from a relationship. We're removing values from a relationship.

bart-degreed:
Continued in #874.

PR #874, opened 2020-11-06
Write enhancements

### LINKED ISSUE #353: Support POST requests to a to-many relationship link

### LINKED ISSUE #578: Response when updating relationship of resource not conform spec

### LINKED ISSUE #790: Support for Self-Referencing ManyToMany Relationships
## Description
- When trying to use HasManyThrough on a self-referencing ManyToMany relationship, an error is thrown. This is because SingleOrDefault is used to find the left and right joining properties: https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/9d7614f1679a1dced5806ad6736ac5b9a5c3bd5f/src/JsonApiDotNetCore/Builders/ResourceGraphBuilder.cs#L160. A special case should be added if the property annotated by HasManyThrough is an IEnumerable of the same type as the current class.

## Environment

- JsonApiDotNetCore Version: v4.0.0-alpha5
- Other Relevant Package Versions: Asp Net Core 3.1

### LINKED ISSUE #850: Support DELETE requests to a to-many relationship link

### LINKED ISSUE #852: Patching  resource relationship on primary endpoint with non-existing resource fails

### LINKED ISSUE #853: Implicit removes and FK constraint violations on PATCH relationship endpoint

### LINKED ISSUE #870: PATCH Successfully but Return Data is Null

bart-degreed:
Continuation of #851.

This PR concerns write operations and addresses several topics:

- Add POST/DELETE relationship endpoints, which enable to add/remove resources to/from a to-many relationship.
- Redesign of handling updates, where we try to fetch the minimum data required to perform the operation. If the operation fails, we run additional queries to determine why and produce a descriptive error according to json:api requirements.
- Rely more on the EF Core navigation metadata model, instead of dealing with foreign keys directly.
- Apply the whole operation in a single `SaveChanges()` call, removing the need for internal transaction management.
- Add resource change tracking for POST, similar to PATCH. Note that in practice, you'll only notice if using client-generated IDs.
- Refactoring of tests to hit all combinations of create/update/delete for the various endpoints and relationship kinds.
- Additional tests for corner cases like malformed request bodies, foreign key constraint violations, implicit removal in one-to-one relationships, required/optional relationships, composite foreign keys etc.

Fixes #353
Fixes #850 
Fixes #852
Fixes #853
Fixes #578
Fixes #870
Fixes #790


bart-degreed:
Replaced by #875.

Issue #732, opened 2020-05-04
Feature Request: Use C# Source Generators to reduce boilerplate controllers


jeff-becker-projekt202:
## Description
One of my current annoyances is )that I have a lot of boilerplate controllers. This could be addressed with a [Source Generator](https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/) which inspects a data context and emits controllers for the configured resources.

I'll implement & PR this, I'd just like feedback on what it needs to look like.
...

## Environment

- JsonApiDotNetCore Version: 4.x
- Other Relevant Package Versions:

bart-degreed:
This sounds interesting. I'm wondering how you would know which controllers to generate, based on the syntax trees that represent the project source code. In the past, all resources had to be exposed as DbSet properties on DbContext, but we've recently switched to using EF Core's DbContext.Model instead, which is populated at application runtime.

jeff-becker-projekt202:
So my thinking was to determine the set of resources based on classes marked with the resource attribute. Then only emit controllers for those if there is not already a controller defined for that resource.

bart-degreed:
I think that could work. `ResourceAttribute` currently requires to specify a resource name. but we can make that optional.

wayne-o:
Does this need to be actual real code generated - could the framework
CreateInstance of the type if it senses there isn't one? Kind of like what
ember does.

bart-degreed:
Probably that is possible too, according to https://stackoverflow.com/questions/848904/in-asp-net-mvc-is-it-possible-to-make-a-generic-controller.

KirovAir:
It should be fairly easily doable.
https://www.strathweb.com/2018/04/generic-and-dynamically-generated-controllers-in-asp-net-core-mvc/

jeff-becker-projekt202:
[...]

Issue #365, opened 2018-08-01
Make controller boilerplate optional


jaredcnance:
We can probably make the controller boilerplate optional a few different ways:
- Define a non-generic controller with parametrized route definitions and then route to the correct services. This would be similar to the approach used for operations. The problem here is you cannot define your own controllers. This would be a breaking change
- Codegen controllers that haven't been defined. Or require an attribute on the resource/entity class. See https://github.com/AArnott/CodeGeneration.Roslyn and https://jeremydmiller.com/2015/11/11/using-roslyn-for-runtime-code-generation-in-marten/ ... need to make sure we can generate types within the executing assembly.

bart-degreed:
Related to #732.

PR #900, opened 2020-12-02
Dynamic controller generation

maurei:
Introduces dynamically generated controllers to reduce boilerplate code as well as improvements in `JsonApiRoutingConvention` to support that.

Context of this PR is the orientation on how OpenAPI integration will look like. An element there that will probably will be introduced is controller action annotations. A challenge here will be to see how we can still reduce controller boilerplate code while allowing for easy documentation of the individual endpoints

Related to #732 and #365. It doesn't address the issue using source generators as proposed in that issue, but we're reflectively registering controllers at runtime. I suspect this will be more flexible.


bart-degreed:
Closing in favor of #1117.

PR #903, opened 2020-12-08
Fixed: write content type on invalid Accept headers

bart-degreed:
Fixed: write content type on invalid Accept headers, to enable FireFox to render the response as json.

Before/after:
image


wayne-o:
could we accept extensions to this accept header? so something like

application/vnd.api+json;domain-command=DoSomethingDomainey


Currently this doesn't work

____________________

*Wayne Douglas*
Co-founder & Architect

*highstreet.ly <https://highstreet.ly/>*
Mobile: 07508 215459

------------------------------------


____________________

bart-degreed:
@wayne-o The syntax you're proposing is not allowed by the json:api spec, see [here](https://jsonapi.org/format/1.1/#content-negotiation).

However, using "profiles" and "extensions" are allowed (though they have some strict rules). Could you open a separate issue to describe what you need and how JADNC could support your use case?

maurei (approved):

PR #909, opened 2020-12-14
Naming consistency

bart-degreed:
Naming consistency: our framework is called JsonApiDotNetCore and the spec it implements is called JSON:API (in this casing)
Fixed invalid doc-comment reference

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 <img src ="https://raw.githubusercontent.com/json-api-dotnet/JsonApiDotnetCore/master/logo.png" />
 </p>
 
-# JSON:API Framework for ASP.NET Core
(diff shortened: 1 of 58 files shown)


maurei (approved):
Note: we are anticipating a rename of the project in the future as soon as the `Core` suffix disappears in the namespaces of new releases of `ASP.NET`. When this happens we will consider to drop `Core` from the project and repository name too.  We will then also adopt a more esthetic spelling like `JSON:API.NET` in the docs because it is customary to use normal interpunction in project titles (e.g. [.ASP.NET Core](https://github.com/dotnet/aspnetcore) in the docs/readme vs `aspnetcore` as the repository name).

Issue #350, opened 2018-07-23
DbUpdateConcurrencyException when creating resources with relationships with ConcurrencyStamp


jaredcnance:
Currently, if you try to create a resource with a HasOne relationship to an object that has a ConcurrencyStamp property via the EF `.IsConcurrencyToken()` method you will likely get a `DbUpdateConcurrencyException` thrown. Many of the ASP.Net Core Identity classes have this concurrency check.

See this for details: https://github.com/aspnet/EntityFrameworkCore/issues/9166#issuecomment-318204397

As a workaround, you need to identify the resources that enforce the concurrency check and manually fetch them from the database. You also need to prevent JADNC from attaching the relationship pointer. So, change this:

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/e22e101739238b4d7ca1d8b078c491c23b417f72/src/JsonApiDotNetCore/Data/DefaultEntityRepository.cs#L85-L92

To something along the lines of this:

```c#
public override async Task<Relationship> CreateAsync(Relationship relationship)
{
    // this would need to implement something similar to the default
    // implementation, excluding any relationships that include the 
    // concurrency stamp
    AttachOtherRelationships();

    if (relationship.Role == null)
        throw new JsonApiException(400, "Cannot create relationship without a Role");

    var role = await _context.Roles.SingleOrDefaultAsync(r => r.Id == relationship.Role.Id);
    if (role == null)
        throw new JsonApiException(400, $"Role '{relationship.Role?.Id}' does not exist");

    // manually assign the Role
    relationship.Role = role;

    _relationships.Add(relationship);
    await _context.SaveChangesAsync();
    return relationship;
}
```

bart-degreed:
Superseded by #1004.

PR #917, opened 2021-01-10
Explores support for concurrency tokens using PostgreSQL

bart-degreed:
Related to #350.

*Update (2021-11-08): Rebased on latest changes in master branch*

Because we fetch the row before update and apply changes on that, a concurrency violation is only reported when two concurrent requests update the same row in parallel. Instead, we want to produce an error if the token sent by the user does not match the stored token. To do that, we need to convince EF Core to use that as original version. That's not too hard.

Now the problem is that there is no way to send the token for relationships or deleting a resource. Skipped tests have been added to demonstrate this.

We could fetch such related rows upfront to work around that, but that kinda defeats the purpose of using concurrency tokens in the first place. It may be more correct to fail when a user is trying to add a related resource that has changed since it was fetched. This reasoning may be a bit too puristic and impractical, but at least that's how EF Core seems to handle it.

Solutions considerations:
- Add 'version' to resource identifier object, so the client can send it. The spec does not explicitly forbid adding custom fields, however 'meta' would probably be the recommended approach. Instead of extending the definition, we could encode it in the StringId.
- Once we have access to that token value, we need to somehow map that to 'the' resource property. What if there are multiple concurrency token properties on a resource? And depending on the database used, this could be typed as numeric, guid, timestamp, binary or something else.
- Given that PostgreSQL uses a number (uint xmin), should we obfuscate or even encrypt that? If the latter, we need to add an option for api developers to set the encryption key.

See also:
https://github.com/json-api/json-api/issues/600
https://github.com/json-api/json-api/pull/824

### DIFF
--- a/JsonApiDotNetCore.sln.DotSettings
+++ b/JsonApiDotNetCore.sln.DotSettings
@@ -637,5 +637,6 @@ $left$ = $right$;</s:String>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=subdirectory/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=unarchive/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Workflows/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=xmin/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=xunit/@EntryIndexedValue">True</s:Boolean>
 </wpf:ResourceDictionary>
--- a/src/JsonApiDotNetCore/Errors/DataConcurrencyException.cs
+++ b/src/JsonApiDotNetCore/Errors/DataConcurrencyException.cs
@@ -1,0 +1,23 @@
+using System;
+using System.Net;
+using JetBrains.Annotations;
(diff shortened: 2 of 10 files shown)


bart-degreed:
To solve the second and third bullet, we can leave it up to the resource class itself to choose the matching property and apply the conversion. Something along the lines of:

```c#
// JADNC
public interface IVersionedIdentifiable : IIdentifiable
{
    string Version { get; set; }
}

public abstract class VersionedIdentifiable<TId> : Identifiable<TId>, IVersionedIdentifiable
{
    public abstract string Version { get; set; }
}

// API Project
public sealed class Disk : VersionedIdentifiable<long>
{
    [Attr]
    public string Manufacturer { get; set; }

    [Attr]
    public string SerialCode { get; set; }

    [ConcurrencyCheck]
    [Timestamp]
    [Attr(PublicName = "concurrencyToken")]
    public uint xmin { get; set; }

    public override string Version
    {
        get => xmin.ToString();
        set => xmin = uint.Parse(value);
    }
}
```

bart-degreed:
> Add 'version' to resource identifier object

This won't work for DELETE resource requests, because no body is being sent. The same applies for updating a relationship via relationship endpoints. We could decide on a separator that is not allowed to occur in ID or version. For example, choosing '@' would allow:

```http
DELETE /photos/123@789 HTTP/1.1
```
which we can then interpret as: resource with ID 123 at version 789. Perhaps '@' is not the best choice, in case someone uses email addresses for IDs.

bart-degreed:
Superseded by #1119.

PR #930, opened 2021-01-22
Added support for bulk/batch requests (atomic:operations)

### LINKED ISSUE #247: Operations: Should be ORM agnostic

### LINKED ISSUE #936: Add support for bulk/batch requests (atomic:operations)
JsonApiDotNetCore v3.x provided support for bulk/batch by implementing the experimental [JSON:API Bulk Extension](https://jsonapi.org/extensions/bulk/). This extension was deprecated some time ago, so we removed the code in v4.

An official (non-experimental) specification is now available in the form of [Atomic Operations](https://jsonapi.org/ext/atomic/), which we intend to implement.

bart-degreed:
## Introduction

This PR implements the [Atomic Operations](https://jsonapi.org/ext/atomic/) JSON:API extension, which defines how to perform multiple write operations in a linear and atomic manner.

Clients can send an array of operations in a single request. JsonApiDotNetCore guarantees that those operations will be processed in order and will either completely succeed or fail together.

For more details, see the updated documentation in this PR.

## How it works

The request execution pipeline for operations is the following:
`RequestDeserializer` -> `OperationsController` (***new***) -> `OperationsProcessor` (***new***) -> `ResourceService` -> `ResourceRepository` -> `ResponseSerializer`

The deserializer and serializer have been updated to understand the new document structure. A new controller type performs ASP.NET ModelState validation on all operations upfront and then delegates to the processor. The processor starts a database transaction and loops over the operations in the list. For each operation, the associated service is called. From there, it follows the usual flow for single-resource requests. After all operations have been processed, the processor commits the transaction.

Fixes #936.
Fixes #247.


src/JsonApiDotNetCore/Serialization/AtomicOperationsResponseSerializer.cs
```diff
@@ -0,0 +1,97 @@
...
+            return SerializeObject(document, _options.SerializerSettings);
+        }
+
+        private AtomicResultObject SerializeOperation(OperationContainer operation)
+        {
+            ResourceObject resourceObject = null;
+
+            if (operation != null)
+            {
+                _request.CopyFrom(operation.Request);
```
maurei, line 71: This feels a bit like trickery: copying values from the operation container to the `IJsonApiRequest` and calling `Reset` below to preserve the flow of the pipeline which was built for one operation per request.
To not have to do tricky with copying new values and resetting caches, we would need to have multiple `IJsonApiRequest` instances floating around. Let's discuss an approach in this direction?
In fact, `IJsonApiRequest` probably hasn't been a good naming choice. At the time, a client request was always associated to one JSON:API operation and now we're supporting multiple operations (still just one request).
So perhaps an approach could be: `IJsonApiRequest` should have minimally one `JsonApiOperation` in `IJsonApiRequest.Operations`. Then it would be possible to have dedicated instances of eg `TargetedFields`, `FieldToSerialize` associated to each operation.

src/JsonApiDotNetCore/Serialization/FieldsToSerialize.cs
```diff
@@ -63,5 +63,11 @@ public IReadOnlyCollection<RelationshipAttribute> GetRelationships(Type resource
             var resourceContext = _resourceContextProvider.GetResourceContext(resourceType);
             return resourceContext.Relationships;
         }
+
```
maurei, line 66: In operations, coped services are being reused for multiple requests. Originally, the resource pipeline is designed for scoped usage and not to be reused. Would it be cleaner to create a new instance for every operation
so that we don't break with that pattern? The introduction of methods like Reset on objects which are designed for a context where resetting makes no sense feels weird to me.
That would mean we would have a list of instances of this services, each of them associated to an operations result.

Alternatively, can we let this service manage the state of the cache internally eg. in the beginning of every attributes call (I didn't think this through) to obfuscate that the service is being reused, so to say?

src/JsonApiDotNetCore/Middleware/IJsonApiRequest.cs
```diff
@@ -57,5 +58,20 @@ public interface IJsonApiRequest
...
+        /// In case of an atomic:operations request, this indicates the kind of operation currently being processed.
+        /// </summary>
+        OperationKind? OperationKind { get; }
+
+        /// <summary>
+        /// In case of an atomic:operations request, identifies the overarching transaction.
+        /// </summary>
+        Guid? TransactionId { get; }
+
+        /// <summary>
```
maurei, line 72: Suggested name: CopyValuesFrom, or UseValuesFrom. I got confused because I was expecting a new JsonApiRequest instance when a copy method is called on a JsonApiRequest.

src/JsonApiDotNetCore/Resources/OperationContainer.cs
```diff
@@ -0,0 +1,54 @@
+using System;
+using System.Collections.Generic;
+using JsonApiDotNetCore.Middleware;
+
+namespace JsonApiDotNetCore.Resources
+{
+    /// <summary>
```
maurei, line 7: The name OperationContainer doesn't imply it is for writing, can we come up with a more descriptive name?I s OperationContext perhaps a name that sets clearer expectations?

src/JsonApiDotNetCore/Resources/OperationContainer.cs
```diff
@@ -0,0 +1,54 @@
...
+            Kind = kind;
+            Resource = resource ?? throw new ArgumentNullException(nameof(resource));
+            TargetedFields = targetedFields ?? throw new ArgumentNullException(nameof(targetedFields));
+            Request = request ?? throw new ArgumentNullException(nameof(request));
+        }
+
+        public void SetTransactionId(Guid transactionId)
+        {
+            ((JsonApiRequest) Request).TransactionId = transactionId;
+        }
```
maurei, line 29: I'm confused. What does it mean to call WithResource on an operations write container?

src/JsonApiDotNetCore/Resources/OperationContainer.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public OperationContainer WithResource(IIdentifiable resource)
+        {
+            if (resource == null) throw new ArgumentNullException(nameof(resource));
+
+            return new OperationContainer(Kind, resource, TargetedFields, Request);
+        }
+
+        public ISet<IIdentifiable> GetSecondaryResources()
```
maurei, line 38: What does it mean to get secondary resources from an object that represents a write operation?

src/JsonApiDotNetCore/Repositories/DbContextExtensions.cs
```diff
@@ -42,5 +44,20 @@ public static object GetTrackedIdentifiable(this DbContext dbContext, IIdentifia
...
         }
+
+        /// <summary>
+        /// Detaches all entities from the change tracker.
+        /// </summary>
+        public static void ResetChangeTracker(this DbContext dbContext)
+        {
+            if (dbContext == null) throw new ArgumentNullException(nameof(dbContext));
+
+            List<EntityEntry> entriesWithChanges = dbContext.ChangeTracker.Entries().ToList();
```
maurei, line 55: No need to call ToList() here?

src/JsonApiDotNetCore/AtomicOperations/Processors/CreateProcessor.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            if (operation.Resource.LocalId != null)
+            {
+                var serverId = newResource != null ? newResource.StringId : operation.Resource.StringId;
+                var resourceContext = _resourceContextProvider.GetResourceContext<TResource>();
+
+                _localIdTracker.Assign(operation.Resource.LocalId, resourceContext.PublicName, serverId);
+            }
+
+            return newResource == null ? null : operation.WithResource(newResource);
```
maurei, line 42: I'm confused with the flow of OperationContainer here. This method accepts an operation container which "represents a write operation",
and returns a new instance of such a write operation representing object?  Should we consider separating a OperationsResult from the OperationsContainer which currently seems to represent both a result from a write operation as well as the write operation itself?

maurei (changes requested):

src/JsonApiDotNetCore/Serialization/FieldsToSerialize.cs
```diff
@@ -63,5 +63,11 @@ public IReadOnlyCollection<RelationshipAttribute> GetRelationships(Type resource
...
+        /// <inheritdoc />
+        public void ResetCache()
+        {
+            _sparseFieldSetCache.Reset();
```
maurei, line 70: Is it feasible to internally track a dedicated `SparseFieldSetCache` per operation, so that this object becomes less of an object that is reused in multiple operation contexts, but is actually designed to function over multiple operations?

src/JsonApiDotNetCore/Middleware/IJsonApiRequest.cs
bart-degreed: `void CopyFrom()` is a commonly used pattern, see https://www.google.com/search?q=CopyFrom&rlz=1C1CHBF_enUS859US859&oq=CopyFrom&aqs=chrome..69i57&sourceid=chrome&ie=UTF-8.
We're not copying just some values, we're copying _everything_, that's the pattern. So I think its clearest to keep it this way.

src/JsonApiDotNetCore/Serialization/AtomicOperationsResponseSerializer.cs
bart-degreed: We've discussed this offline. These are all valid points. The approach I've chosen is to keep it backwards compatible. We should consider to refactor this for the next major release.

Making `IJsonApiRequest` a rich object is like returning to what `IJsonApiContext` used to be and is not the right direction in my opinion. It is built in the middleware, where we don't have access to the request body. Instead, we should simplify `IJsonApiRequest` and have another (non-injected) object like `OperationContainer` that contains all info (targeted fields etc) that is built during deserialization and contains the data we need to enter the service layer. We could then pass along this container object for operations-requests and single-resource requests, which enables unification.

src/JsonApiDotNetCore/Resources/OperationContainer.cs
bart-degreed: I'm trying to avoid the word 'context' here, which often means 'throw in everything you may need'. OperationContainer is currently only used for atomic:operations, which involves writes (create/update/delete) only. We may be able to reuse this structure in the next major version to also wrap input data for read requests.

src/JsonApiDotNetCore/Resources/OperationContainer.cs
bart-degreed: It refers to [withers](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/with-expression), a pattern to change a single property of an immutable object that produces a new immutable object with the changed value.

src/JsonApiDotNetCore/Resources/OperationContainer.cs
bart-degreed: This is a calculation method that solely acts on information within the object itself. It is used for relationship changes.

src/JsonApiDotNetCore/Repositories/DbContextExtensions.cs
bart-degreed: First reason is this makes debugging a little easier. The second is I don't want to depend on how setting state to detached influences the change tracker internals. If setting that property removes it from the internal list, we could get an exception 'the underlying collection has changed' while enumerating, so it feels a bit safer to me to take a snapshot first.

src/JsonApiDotNetCore/AtomicOperations/Processors/CreateProcessor.cs
bart-degreed: It used to return a resource instead of a container, but then I found I needed more data from the container downwards, so I decided to reuse the structure. I would define a separate structure for the result, but that may bite us in the future when we need more of the incoming data and would require to introduce a breaking change. Given that, I think it is not worth it.

src/JsonApiDotNetCore/Serialization/FieldsToSerialize.cs
bart-degreed: We could, although I'm not sure if we should do that. We don't know what logic executes in `IResourceDefinition.OnApplySparseFieldSet` so it may take a dependency on things that change. If we were to cache this over all operations, the order of operations in the request could lead to different outcomes. The downside of re-executing it per operation is that the custom logic is expensive. In that case, the developer implementing the callback can do its own caching. So I think we should leave it as-is.

src/JsonApiDotNetCore/AtomicOperations/Processors/ICreateProcessor.cs
```diff
@@ -0,0 +1,14 @@
+using JsonApiDotNetCore.Resources;
+
+namespace JsonApiDotNetCore.AtomicOperations.Processors
+{
+    /// <summary>
+    /// Processes a single operation to create a new resource with attributes, relationships or both.
```
maurei, line 6: "with attributes, relationships or both" is a bit redundant

src/JsonApiDotNetCore/AtomicOperations/Processors/IUpdateProcessor.cs
```diff
@@ -0,0 +1,15 @@
+using JsonApiDotNetCore.Resources;
+
+namespace JsonApiDotNetCore.AtomicOperations.Processors
+{
+    /// <summary>
+    /// Processes a single operation to update the attributes and/or relationships of an existing resource.
+    /// Only the values of sent attributes are replaced. And only the values of sent relationships are replaced.
```
maurei, line 7: I think the idea of partial update doesn't have to be explained here, its the default approach which is also central to the spec.

src/JsonApiDotNetCore/AtomicOperations/Processors/ICreateProcessor.cs
bart-degreed: This was copied from the service layer. I think it clarifies what you can do, which is not directly obvious to everyone. We have tests for only attributes, only relationships or neither. Aside from that, we have relationship endpoint which always target a single relationship instead of multiple.

src/JsonApiDotNetCore/AtomicOperations/Processors/IUpdateProcessor.cs
bart-degreed: See my other remark, it was taken from service layer too. We have these comments on controllers, services and repositories already, I think having it here too complements that.

maurei (approved):

Issue #924, opened 2021-01-18
After enabling options.EnableResourceHooks = true; I get a NullReference on GET Articles/GetSecondary [id/relationshipName]


woofi7:
#### DESCRIPTION
The request is crashing on this line because the type is `IIdentifiable` instead of the implemented type `Image`.
https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/src/JsonApiDotNetCore/Configuration/ResourceGraph.cs#L85

The implementation seems to be lost at this place: 
https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/src/JsonApiDotNetCore/Hooks/Internal/ResourceHookExecutorFacade.cs#L131
```
System.NullReferenceException: Object reference not set to an instance of an object.
  at JsonApiDotNetCore.Configuration.ResourceGraph.GetRelationships
    C:\projects\jsonapidotnetcore\src\JsonApiDotNetCore\Configuration\ResourceGraph.cs:85
  at JsonApiDotNetCore.Hooks.Internal.Traversal.TraversalHelper.RegisterRelationshipProxies
    C:\projects\jsonapidotnetcore\src\JsonApiDotNetCore\Hooks\Internal\Traversal\TraversalHelper.cs:200
  at JsonApiDotNetCore.Hooks.Internal.Traversal.TraversalHelper.CreateRootNode
    C:\projects\jsonapidotnetcore\src\JsonApiDotNetCore\Hooks\Internal\Traversal\TraversalHelper.cs:56
  at JsonApiDotNetCore.Hooks.Internal.ResourceHookExecutor.GetHook
    C:\projects\jsonapidotnetcore\src\JsonApiDotNetCore\Hooks\Internal\ResourceHookExecutor.cs:199
  at JsonApiDotNetCore.Hooks.Internal.ResourceHookExecutor.OnReturn
    C:\projects\jsonapidotnetcore\src\JsonApiDotNetCore\Hooks\Internal\ResourceHookExecutor.cs:121
  at JsonApiDotNetCore.Hooks.Internal.ResourceHookExecutorFacade.OnReturnRelationship
    C:\projects\jsonapidotnetcore\src\JsonApiDotNetCore\Hooks\Internal\ResourceHookExecutorFacade.cs:131
  at JsonApiDotNetCore.Services.JsonApiResourceService`2+<GetSecondaryAsync>d__11.MoveNext
[...]

PR #932, opened 2021-01-27
Check method GetResourceContext(type) returns null

farturi:
Check if method GetResourceContext(type) returns null before returning Relationships

Closes #{ISSUE_NUMBER}

#### QUALITY CHECKLIST
- [ ] Adapted tests
- [ ] Documentation updated
- [ ] Created issue to update [Templates](https://github.com/json-api-dotnet/Templates/issues/new): {ISSUE_NUMBER}

### DIFF
--- a/src/JsonApiDotNetCore/Configuration/ResourceGraph.cs
+++ b/src/JsonApiDotNetCore/Configuration/ResourceGraph.cs
@@ -82,7 +82,14 @@ namespace JsonApiDotNetCore.Configuration
         {
             if (type == null) throw new ArgumentNullException(nameof(type));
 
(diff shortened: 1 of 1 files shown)


bart-degreed:
This is a breaking change. Where the call would have failed before, it now silently hides the error. What's the reason for changing this?

farturi:
Hi Bart, this is to fix this problem:
https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/924

bart-degreed:
I appreciate that you're trying to help, but I think this is not the right solution. Instead, the caller should be changed to not send an invalid type.

PR #947, opened 2021-02-09
Added auto-expansion of inheritdoc comments

### LINKED ISSUE #940: Missing help in IntelliSense
We found that not all documentation-comments from JADNC are displayed in Visual Studio when JADNC is consumed as a package reference. This happens on `<inheritdoc />` usage, which is poorly supported by IDEs.

We should try if https://github.com/saucecontrol/InheritDoc solves that for us. It can expand inheritdoc markers in doc-comment xml files as a build step.

bart-degreed:
Added auto-expansion of inheritdoc comments using https://github.com/saucecontrol/InheritDoc, so that VS Intellisense picks it up when consumed as package reference.

Fixes #940.

### DIFF
--- a/src/JsonApiDotNetCore/Controllers/JsonApiCommandController.cs
+++ b/src/JsonApiDotNetCore/Controllers/JsonApiCommandController.cs
@@ -18,7 +18,9 @@ namespace JsonApiDotNetCore.Controllers
     /// <typeparam name="TId">The resource identifier type.</typeparam>
     public abstract class JsonApiCommandController<TResource, TId> : BaseJsonApiController<TResource, TId> where TResource : class, IIdentifiable<TId>
     {
-        /// <inheritdoc />
+        /// <summary>
+        /// Creates an instance from a write-only service.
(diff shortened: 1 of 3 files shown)


bart-degreed:
This does not remove warnings from DocFX, will look at those later.

bart-degreed:
Looked into why warnings are still being produced by DocFX. The reason is that DocFX uses roslyn to re-parse the sources and discards the compiler-generated JsonApiDotNetCore.xml that was produced during solution build.

This can be changed in docfx.json by replacing:
```json
"files": [ "**/JsonApiDotNetCore.csproj" ],
```
with:
```json
"files": [ "src/JsonApiDotNetCore/bin/Release/netcoreapp3.1/JsonApiDotNetCore.dll" ],
```

The upside of that is documentation generation runs a lot faster and handles inheritdoc properly, but the downside is we lose integration with sources on GitHub and some other minor details, so I think it is not worth the change.

Before:
image

After:
image

maurei (approved):

PR #967, opened 2021-03-11
Make InputFormatters and OutputFormatters implement IApiRequestFormatMetadataProvider

maurei:
- `JsonApiInputFormatter` implements `IApiRequestFormatMetadataProvider`. 
- `JsonApiOutputFormatter` implements `IApiResponseFormatMetadataProvider`.
- Example test project has been adjusted so that for each test collection only controllers are registered that originate from the same namespace, rather than loading all the controllers from the entire test assembly.


src/JsonApiDotNetCore/Middleware/JsonApiFormatter.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            return new MediaTypeCollection
+            {
+                new MediaTypeHeaderValue(mediaType)
+            };
+        }
+
+        private bool IsAtomicOperationsType(Type objectType)
+        {
+            return objectType.GetInterface(nameof(IEnumerable)) != null && objectType.GetGenericArguments().First() == _operationContainerType;
```
bart-degreed, line 35: I wanted to try why such complex logic is needed, instead of the obvious `typeof(IList<OperationContainer>)`, but there are no tests. Add some tests first, please.

src/JsonApiDotNetCore/Middleware/JsonApiInputFormatter.cs
```diff
@@ -5,8 +5,8 @@
 
 namespace JsonApiDotNetCore.Middleware
 {
-    /// <inheritdoc />
-    public sealed class JsonApiInputFormatter : IJsonApiInputFormatter
+    /// <inheritdoc cref="JsonApiDotNetCore.Middleware.IJsonApiInputFormatter" />
```
bart-degreed, line 8: I assume the cref is to disambiguate between the docs on JsonApiFormatter and IJsonApiInputFormatter. We haven't used this before, but perhaps we should start doing that. Does this work correctly in:
- the generated documentation website?
- IntelliSense in IDE when consuming our NuGet?

bart-degreed (changes requested):

src/JsonApiDotNetCore/Middleware/JsonApiFormatter.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.Net.Http.Headers;
+
+namespace JsonApiDotNetCore.Middleware
+{
+    /// <summary>
+    /// Application-wide entry point for writing JSON:API response bodies.
+    /// </summary>
+    public abstract class JsonApiFormatter : IApiRequestFormatMetadataProvider
+    {
+        private readonly Type _operationContainerType = typeof(OperationContainer);
```
bart-degreed, line 17: static

maurei:
superseded by #969 and #972

PR #969, opened 2021-03-17
Test controller provider

maurei:
Introduces a mechanism to explicitly register only the controllers required for a particular test suite, rather than always registering all controllers from the entire test assembly.

This mechanism is explicitly available in the constructor of every test suite. Example:

```c#
public class TestCollection : IClassFixture<Fixture<TestStartup, TestDbContext>>
{
    public TestCollection(Fixture<TestStartup, TestDbContext> testContext)
    {
        testContext.AddController<OperationsController>();
    }	
}
```


test/TestBuildingBlocks/BaseIntegrationTestContext.cs
```diff
@@ -41,6 +43,12 @@ protected BaseIntegrationTestContext()
             _lazyFactory = new Lazy<WebApplicationFactory<TRemoteStartup>>(CreateFactory);
         }
 
+        public void AddController<TController>()
```
bart-degreed, line 46: I'd like this renamed to `RegisterController` or `UseController`, because it is a method on the test context. If it was an (extension) method on `IServiceCollection`, `AddController` would be appropriate.

test/TestBuildingBlocks/BaseIntegrationTestContext.cs
```diff
@@ -57,6 +65,8 @@ private WebApplicationFactory<TRemoteStartup> CreateFactory()
 
             factory.ConfigureServicesBeforeStartup(services =>
             {
+                services.UseControllers(_testControllerProvider);
```
bart-degreed, line 68: This should be called after `_beforeServicesConfiguration?.Invoke(services);`.

test/TestBuildingBlocks/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+using JsonApiDotNetCore;
+using Microsoft.AspNetCore.Mvc.ApplicationParts;
+using Microsoft.AspNetCore.Mvc.Controllers;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace TestBuildingBlocks
+{
+    internal static class ServiceCollectionExtensions
+    {
+        public static void UseControllers(this IServiceCollection services, TestControllerProvider provider)
```
bart-degreed, line 12: This should be split into `ClearControllers` and `AddControllers` methods, which are sequentially called from test context. Alternatively, rename to `ReplaceControllers`.

test/TestBuildingBlocks/TestControllerProvider.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Microsoft.AspNetCore.Mvc.Controllers;
+
+namespace TestBuildingBlocks
+{
+    internal sealed class TestControllerProvider : ControllerFeatureProvider
+    {
+        private readonly IList<Type> _allowedControllerTypes = new List<Type>();
```
bart-degreed, line 10: I think this should be a `HashSet` too. And please add a line break, to separate fields from properties.

bart-degreed (changes requested):
Looking good overall, few remarks.

src/JsonApiDotNetCore/Middleware/HeaderConstants.cs
```diff
@@ -1,7 +1,10 @@
+using JetBrains.Annotations;
+
 #pragma warning disable AV1008 // Class should not be static
 
 namespace JsonApiDotNetCore.Middleware
 {
+    [PublicAPI]
```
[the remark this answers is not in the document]
maurei: The CI build failed on it earlier, see https://ci.appveyor.com/project/jaredcnance/jsonapidotnetcore/builds/38266703

bart-degreed (approved):

PR #977, opened 2021-04-12
Write callbacks

### LINKED ISSUE #934: Provide write callbacks in IResourceDefinition

bart-degreed:
Closes #934


src/JsonApiDotNetCore/Serialization/Building/IncludedResourceObjectBuilder.cs
```diff
@@ -188,23 +194,25 @@ protected override RelationshipEntry GetRelationshipData(RelationshipAttribute r
             };
         }
 
-        /// <summary>
-        /// Gets the resource object for <paramref name="parent" /> by searching the included list. If it was not already built, it is constructed and added to
-        /// the inclusion list.
-        /// </summary>
-        private ResourceObject GetOrBuildResourceObject(IIdentifiable parent)
+        private ResourceObject TryGetBuiltResourceObjectFor(IIdentifiable resource)
```
maurei, line 197: Why not leave it as a single method and just rename it to `GetBuiltResourceObjectFor`?

src/JsonApiDotNetCore/Serialization/IFieldsToSerialize.cs
```diff
@@ -10,6 +10,11 @@ namespace JsonApiDotNetCore.Serialization
     /// </summary>
     public interface IFieldsToSerialize
     {
+        /// <summary>
+        /// Indicates whether attributes and relationships should be serialized, based on the current endpoint.
+        /// </summary>
```
maurei, line 15: I see why we need this information in the serializer and how introducing it this way is non-breaking, but since the depencency on the endpoint type is in response serializer and not in this service, it feels wrong to add this property here. Should we add it as an obsolete field and inject `JsonApiRequest` in the response serializer in the next breaking release?

maurei (commented):
First review pass

src/JsonApiDotNetCore/Repositories/EntityFrameworkCoreRepository.cs
```diff
@@ -167,18 +172,48 @@ public virtual async Task CreateAsync(TResource resourceFromRequest, TResource r
...
+
             DbSet<TResource> dbSet = _dbContext.Set<TResource>();
             await dbSet.AddAsync(resourceForDatabase, cancellationToken);
 
             await SaveChangesAsync(cancellationToken);
+
+            await _resourceDefinitionAccessor.OnWriteSucceededAsync(resourceForDatabase, OperationKind.CreateResource, cancellationToken);
+        }
+
+        private async Task<object> EditRelationshipAsync(TResource leftResource, RelationshipAttribute relationship, object rightResourceIds,
```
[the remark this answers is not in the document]
bart-degreed: Renamed to `VisitSetRelationshipAsync`, I'm open for other suggestions. Though this is a private method, so I wouldn't worry to much about it.

src/JsonApiDotNetCore/Repositories/EntityFrameworkCoreRepository.cs
```diff
@@ -449,6 +522,8 @@ private bool IsOneToOneRelationship(RelationshipAttribute relationship)
 
         protected virtual async Task SaveChangesAsync(CancellationToken cancellationToken)
         {
+            cancellationToken.ThrowIfCancellationRequested();
```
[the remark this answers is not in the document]
bart-degreed: This prevents saving changes if the HTTP client has disconnected. In general it is good practice to call this before doing expensive work or in expensive loops.

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
```diff
@@ -1,13 +1,16 @@
...
+using JsonApiDotNetCore.Middleware;
 using JsonApiDotNetCore.Queries.Expressions;
+using JsonApiDotNetCore.Resources.Annotations;
 
 namespace JsonApiDotNetCore.Resources
 {
     /// <summary>
-    /// Provides a resource-centric extensibility point for executing custom code when something happens with a resource. The goal here is to reduce the need
-    /// for overriding the service and repository layers.
+    /// Provides an extensibility point to add business logic that is resource-oriented instead of endpoint-oriented.
```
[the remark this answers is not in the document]
bart-degreed: The other key extensibility points (controller, service, repository) are endpoint-oriented, so I'd like to make this distinction explicit.

> Also we don't explicitly exclude any other ways it could be oriented

I don't get your point here, what do you mean with this?

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
```diff
@@ -129,5 +132,205 @@ public interface IResourceDefinition<TResource, TId>
...
         IDictionary<string, object> GetMeta(TResource resource);
+
+        /// <summary>
+        /// Executes after the original version of the resource has been retrieved from the underlying data store, as part of a write request.
+        /// <para>
+        /// Implementing this method enables to perform validations and make changes to <paramref name="resource" />, before the fields from the request are
+        /// copied into it.
+        /// </para>
+        /// <para>
+        /// For POST resource requests, this method is typically used to assign property default values or to side-load-and-attach required relationships.
```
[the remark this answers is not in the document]
bart-degreed: Side load means it gets loaded independently. That's insufficient. Attaching what was side-loaded to the resource is required for correct functioning (in the case of EF Core and other ORMs). Even when implementing a custom repository, it needs to be attached so that a custom repository (using Dapper, for example) can handle it properly.

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
```diff
@@ -129,5 +132,205 @@ public interface IResourceDefinition<TResource, TId>
...
+        /// Note this intentionally excludes <see cref="OperationKind.DeleteResource" /> and <see cref="OperationKind.AddToRelationship" />, because for those
+        /// endpoints no resource is retrieved upfront.
+        /// </param>
+        /// <param name="cancellationToken">
+        /// Propagates notification that request handling should be canceled.
+        /// </param>
+        Task OnPrepareWriteAsync(TResource resource, OperationKind operationKind, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Executes before replacing (overwriting) a to-one relationship.
```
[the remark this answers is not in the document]
bart-degreed: I've reworded this a bit, but still tried to be precise about the effect this has.

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
```diff
@@ -129,5 +132,205 @@ public interface IResourceDefinition<TResource, TId>
...
+        /// <para>
+        /// Implementing this method enables to perform validations and change <paramref name="rightResourceId" />, before the relationship is updated.
```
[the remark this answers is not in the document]
bart-degreed: X would be: "you reading this, the developer that is building an API." I've tried to stick with existing wording used by the read-only callbacks.

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
```diff
@@ -129,5 +132,205 @@ public interface IResourceDefinition<TResource, TId>
...
+        /// </para>
+        /// </summary>
+        /// <param name="leftResource">
+        /// The original resource retrieved from the underlying data store, that declares <paramref name="hasOneRelationship" />.
```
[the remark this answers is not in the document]
bart-degreed: In the next example:
```c#
public class Article
{
    [HasOne]
    public Person Author { get; set; }
}
```
the class Article (which represents the 'articles' resource) declares the relationship. The Person class does not. That's what I'm trying to express.

`RelationshipAttribute.LeftType` is documented as:
> The parent resource type. This is the type of the class in which this attribute was used.

I don't understand the mismatch here.

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
```diff
@@ -129,5 +132,205 @@ public interface IResourceDefinition<TResource, TId>
...
+        /// </param>
+        /// <param name="hasOneRelationship">
+        /// The to-one relationship being replaced or cleared.
```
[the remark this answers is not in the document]
bart-degreed: See my earlier remark on rewordings.

src/JsonApiDotNetCore/Serialization/Building/IncludedResourceObjectBuilder.cs
bart-degreed: Because I'm preventing that `OnSerialize` is called multiple times on the same resource instance. Which may be an expensive call, or just unexpected. For example, if you're prefixing an existing attribute, it gets prefixed multiple times.

src/JsonApiDotNetCore/Serialization/IFieldsToSerialize.cs
bart-degreed: I agree that the separation of responsibilities got polluted a bit here. That already happened with making `FieldsToSerialize` depend on `SparseFieldSetCache`, requiring the `ResetCache` method in the interface (my fault too). What used to be a data-only object is now aware of ambient state. The idea of making `ResponseSerializer` context-aware does not sound too appealing to me either.

So I'd like to propose to keep it like this for now and reconsider if/when we're going to redesign the serialization layer into flexibly reusable building blocks. Today `IFieldsToSerialize` is only used internally in response serialization, which makes it unlikely we'd break existing API projects.

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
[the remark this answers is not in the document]
maurei: I understand that it is pointed to the reader and X would be "you", but leaving it out in the sentence makes it grammatically incorrect

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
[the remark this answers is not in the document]
maurei: Indeed, it is the *class* `Article` that declares the relationship, not a particular *instance* of `Article`. But the `leftResource` parameter description here is about an instance, not the class/resource as a whole.

src/JsonApiDotNetCore/Serialization/Building/IncludedResourceObjectBuilder.cs
maurei: That can also be achieved by leaving it a single method and just calling it [before this line](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/d03316609b5325a325404c3f6c7a56f0a9f2c083/src/JsonApiDotNetCore/Serialization/Building/IncludedResourceObjectBuilder.cs#L204)

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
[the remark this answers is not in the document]
maurei: Given some resource as a starting point, and choosing a relationship and side-loading that relationship, I don't think it makes sense to interpret *"side loading a relationship"* as anything else than loading that relationship value and associating it to the starting point (the initial resource). Introducing an additional word (attach) for just the latter part  just confuses me because it suggests something else needs to be done to apart from what is usually trivially done with it. Also, to underline the unclarity that follows from introducing word: I'm still not sure whether attaching is actually referring to creating that association (assignment?), or some EF Core-esque attaching to something like a change tracker.

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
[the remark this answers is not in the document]
maurei: > I don't get your point here, what do you mean with this?
I mean that it makes more sense to say what it IS instead of naming all the things that it isn't (and that list probably now is incomplete).  

> I'd like to make this distinction explicit
I think stating it is resource-centric is explicit enough. This paragraph is not the place to lay out the distinction between different orientations in extensibility we have throughout the framework. That could be placed in the overarching document about extensibility.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/SoftDeletion/SoftDeletionAwareResourceService.cs
```diff
@@ -0,0 +1,129 @@
...
+using JsonApiDotNetCore.Repositories;
+using JsonApiDotNetCore.Resources;
+using JsonApiDotNetCore.Services;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Logging;
+
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.SoftDeletion
+{
+    [UsedImplicitly(ImplicitUseKindFlags.InstantiatedNoFixedConstructorSignature)]
+    public class SoftDeletionAwareResourceService<TResource, TId> : JsonApiResourceService<TResource, TId>
```
maurei, line 20: Name is a bit awkward. I prefer "SoftDeletionResourceService" because leaving out "aware" will clearly not suggest it is a service that is unaware of soft deletion

test/JsonApiDotNetCoreExampleTests/IntegrationTests/SoftDeletion/SoftDeletionTests.cs
```diff
@@ -175,32 +220,24 @@ public async Task Can_get_secondary_resources()
             httpResponse.Should().HaveStatusCode(HttpStatusCode.OK);
 
             responseDocument.ManyData.Should().HaveCount(1);
-            responseDocument.ManyData[0].Id.Should().Be(company.Departments.Skip(1).Single().StringId);
+            responseDocument.ManyData[0].Id.Should().Be(company.Departments.ElementAt(1).StringId);
         }
 
         [Fact]
-        public async Task Cannot_get_secondary_resources_for_deleted_parent()
+        public async Task Cannot_get_secondary_resource_for_soft_deleted_parent()
```
maurei, line 227: Given that we're differentiating between HasMany and HasOne relationships for relationship endpoint tests, should we do the same for secondary endpoint tests?

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceDefinitions/Serialization/SerializationTests.cs
```diff
@@ -0,0 +1,735 @@
...
+
+            string socialSecurityNumber2 = encryptionService.Decrypt((string)responseDocument.Included[1].Attributes["socialSecurityNumber"]);
+            socialSecurityNumber2.Should().Be(existingScholarship.Participants[2].SocialSecurityNumber);
+
+            hitCounter.DeserializeCount.Should().Be(0);
+            hitCounter.SerializeCount.Should().Be(2);
+        }
+
+        [Fact]
+        public async Task Skips_on_get_ToOne_relationship()
```
maurei, line 497: Perhaps we should consider firing this hook for resource identifiers too: it would seem like a valid use-case to me to handle obfuscated IDs using these hooks, i.e. one would want to be able to serialize and deserialize IDs at all times

test/JsonApiDotNetCoreExampleTests/IntegrationTests/SoftDeletion/SoftDeletionAwareResourceService.cs
bart-degreed: As discussed: aware means it can be chained in an inheritance tree, adding behavior if needed or passing on to the next layer otherwise (see decorator design pattern).

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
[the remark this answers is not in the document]
bart-degreed: As discussed, I intend to change "attach" to "link" to disambiguate with change tracker.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceDefinitions/Serialization/SerializationTests.cs
bart-degreed: We've discussed this and I am against that. Resource definition write + serialization callbacks should be hit for resources, not _references to_ resources. They are intended to act on table data (attributes), not pointers. Reference callbacks might be a feature that is added at a later stage, if we conclude we need it. In any case it should be a separate extensibility point and not intermixed with the existing ones. Because its confusing to sometimes get a fully filled resource instance and an empty instance with only ID filled on other occasions. Users would need to deeply understand how the pipelines work internally to be able to reason about when to get what and revert back to injecting `IJsonApiRequest`, `ITargetedFields` and similar, while the whole point of this PR is to reduce the need for such complexity and provide a simple and straight-forward way to manipulate data.

You then pointed out a suspected inconsistency in the example for archiving: a filter is applied on a secondary to-many resource request, but not on a to-many relationship request. There are currently no tests for relationships. So I've tried it today and it turns out to 'just work':

```c#
[Fact]
public async Task Get_ToMany_relationship_excludes_archived()
{
    // Arrange
    TelevisionStation station = _fakers.TelevisionStation.Generate();
    station.Broadcasts = _fakers.TelevisionBroadcast.Generate(2).ToHashSet();
    station.Broadcasts.ElementAt(1).ArchivedAt = null;

    await _testContext.RunOnDatabaseAsync(async dbContext =>
    {
        dbContext.Stations.Add(station);
        await dbContext.SaveChangesAsync();
    });

    string route = $"/televisionStations/{station.StringId}/relationships/broadcasts";

    // Act
    (HttpResponseMessage httpResponse, Document responseDocument) =
        await _testContext.ExecuteGetAsync<Document>(route);

    // Assert
    httpResponse.Should().HaveStatusCode(HttpStatusCode.OK);

    responseDocument.ManyData.Should().HaveCount(1);
    responseDocument.ManyData[0].Id.Should().Be(station.Broadcasts.ElementAt(1).StringId);
}
```

I would prefer not to try to improve/fix `OnApplyFilter` logic as part of this PR. First reason is because this PR is already quite big, second reason is because it raises additional questions when exactly the read callbacks should execute and we'd need a lot more tests, third is we might potentially break existing users in the process. So let's revisit that topic another time, perhaps when we know better how we want to change it or we have concrete bug reports.

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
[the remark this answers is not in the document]
bart-degreed: Valid point. I'd still like to explain how things are connected, so I've adapted the text to refer to the _type_ instead of the _instance_.

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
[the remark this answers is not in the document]
maurei: I thought we agreed on changing this.  [Enable is a transitive verb](https://www.merriam-webster.com/dictionary/enable), and transitive verbs [cannot be used without an object](https://en.wikipedia.org/wiki/Transitive_verb). Consider replacing with an intransitive verb like "allow for" or rephrasing the sentence as a whole.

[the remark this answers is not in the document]
maurei: I thought we agreed on changing this.  [Enable is a transitive verb](https://www.merriam-webster.com/dictionary/enable), and transitive verbs [cannot be used without an object](https://en.wikipedia.org/wiki/Transitive_verb). Consider replacing with an intransitive verb like "allow for" or rephrasing the sentence as a whole.

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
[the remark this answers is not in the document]
maurei: >> The original resource retrieved from the underlying data store. Its type, <typeparamref name="TResource" />, declares <paramref name="hasOneRelationship" />.

The second sentence now feels misplaced because it comes across more as a general statement that relationships are declared on the level of types (which doens't belong in this particular section) than as a description about what "left" refers to in the parameter name `leftResource`, which I believe is what you're trying to illuminate.

Suggestion:

>> The resource on the left side of <paramref name="hasOneRelationship" /> as retrieved from the underlying data store, indicating that the relationship is declared by <typeparamref name="TResource" name="hasOneRelationship" />.

Or

>> The resource as retrieved from the underlying data store. The indication "left" specifies that <paramref name="hasOneRelationship" /> is declared by <typeparamref name="TResource" />.


Also: 
- `The ~original~ resource`; a tautology as this is implied by retrieving it being retrieved from the data store.
- `The resource ***as*** retrieved`; else it reads as "the resource" being the subject that retrieves something, which is not the case.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceDefinitions/Serialization/SerializationTests.cs
maurei: Good to see it works in this case, I think it adds to the completeness of this example (I guess it is no surprise that it does given the same query layer wrapping that occurs as with the get secondary endpoint) I suggest adding this test to the test collection for the same reason. 

Similarly the equivalent of `Get_primary_resource_by_ID_with_include_and_filter_includes_archived` for the same endpoint of the test in your snippet above should also be included for completeness

How is archiving intended to work with adding to and removing from relationships?

- `Can(not)_add_archived_resource_to_HasMany_relationship`
- `Can(not)_remove_archived_resource_from_HasMany_relationship`



> Resource definition write + serialization callbacks should be hit for resources, not references to resources.
As discussed, I don't feel that having a hard distinction between "resource references" (pointers) and resources is very helpful. I think people will be more inclined to think of them both being "resources", where one is just a more minimal representation of the other, and this would ultimately mean that we should call more hooks for resource identifiers too. I think realising that the current archiving implementation should also work for the relationship endpoint above shows that this way of looking at it is intuitive. The flipside is also true: if we would decide not to, then I suspect the `Get_ToMany_relationship_excludes_archived` case should also not be supported either since no archived resources are involved in that endpoint (just resource references), but to me that clearly wouldn't be favourable.

But I also agree leave it at this for now. Let's revisit this at a later time.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/Archiving/ArchiveTests.cs
```diff
@@ -0,0 +1,612 @@
...
+                services.AddResourceDefinition<TelevisionBroadcastDefinition>();
+            });
+        }
+
+        [Fact]
+        public async Task Can_get_archived_resource_by_ID()
+        {
+            // Arrange
+            TelevisionBroadcast broadcast = _fakers.TelevisionBroadcast.Generate();
+
```
maurei, line 41: Initially I was puzzled about how the archiving took placed, suspected it was perhaps a wrong name in the test and went into the resource definition to figure out how archiving worked. It took me a little bit to figure out that broadcoasts are archived by default. Could that be reversed? I think it will lead to less surprises.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/MultiTenancy/MultiTenancyTests.cs
```diff
@@ -0,0 +1,989 @@
...
+
+            testContext.ConfigureServicesAfterStartup(services =>
+            {
+                services.AddResourceService<MultiTenantResourceService<WebShop>>();
+                services.AddResourceService<MultiTenantResourceService<WebProduct>>();
+            });
+        }
+
+        [Fact]
+        public async Task Get_primary_resources_excludes_other_tenants()
```
maurei, line 44: Get_primary_resources_excludes_resources_from_other_tenants

test/JsonApiDotNetCoreExampleTests/IntegrationTests/MultiTenancy/MultiTenancyTests.cs
```diff
@@ -0,0 +1,989 @@
...
+
+            // Assert
+            httpResponse.Should().HaveStatusCode(HttpStatusCode.OK);
+
+            responseDocument.ManyData.Should().HaveCount(1);
+            responseDocument.ManyData[0].Id.Should().Be(shops[1].StringId);
+        }
+
+        [Fact]
+        public async Task Get_primary_resources_with_include_excludes_other_tenants()
```
maurei, line 101: Get_primary_resources_with_include_excludes_resources_from_other_tenants

maurei: There is a few others to which a similar correction applies

test/JsonApiDotNetCoreExampleTests/IntegrationTests/MultiTenancy/MultiTenancyTests.cs
```diff
@@ -0,0 +1,989 @@
...
+            responseDocument.Errors.Should().HaveCount(1);
+
+            Error error = responseDocument.Errors[0];
+            error.StatusCode.Should().Be(HttpStatusCode.NotFound);
+            error.Title.Should().Be("The requested resource does not exist.");
+            error.Detail.Should().Be($"Resource of type 'webShops' with ID '{shop.StringId}' does not exist.");
+        }
+
+        [Fact]
+        public async Task Cannot_get_secondary_resources_from_other_parent_tenant()
```
maurei, line 165: For the relationship endpoint the test name is very explicit about has many vs has one, whereas for the secondary endpoint this is indicated in a subtle manner: resource**s** vs resource, which led me to think initially that a test was missing. Could we equalise these test names?

Eg
- Cannot_get_resources_from_other_parent_tenant_on_secondary_endpoint_for_has_many_relationship
- Cannot_get_resources_from_other_parent_tenant_on_secondary_endpoint_for_has_one_relationship
- Cannot_get_resources_from_other_parent_tenant_on_relationship_endpoint_for_has_many_relationship
- Cannot_get_resources_from_other_parent_tenant_on_relationship_endpoint_for_has_one_relationship

test/JsonApiDotNetCoreExampleTests/IntegrationTests/Microservices/FireAndForgetDelivery/MessageBroker.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        internal bool SimulateFailure { get; set; }
+
+        internal void Reset()
+        {
+            SimulateFailure = false;
+            SentMessages.Clear();
+        }
+
+        internal Task PostMessageAsync(OutgoingMessage message, CancellationToken cancellationToken)
```
maurei, line 23: I think in the context of a message bus, `PublishMessageAsync` or `PostMessageAsync` would be more appropriate

test/JsonApiDotNetCoreExampleTests/IntegrationTests/Microservices/TransactionalOutboxPattern/OutboxTests.Group.cs
```diff
@@ -0,0 +1,547 @@
...
+using Microsoft.EntityFrameworkCore;
+using TestBuildingBlocks;
+using Xunit;
+
+namespace JsonApiDotNetCoreExampleTests.IntegrationTests.Microservices.TransactionalOutboxPattern
+{
+    public sealed partial class OutboxTests
+    {
+        [Fact]
+        public async Task Create_group_adds_to_outbox()
```
maurei, line 19: Create_group_adds_**message**_to_outbox

Additionally, the test names could be more explicit about which message is expected

bart-degreed: I intentionally did not, because it requires a rename of the test method when the kind or number of messages changes, which is usually forgotten and then the name becomes wrong. I intend the test name to say it adds to the outbox, not whether it adds one or multiple messages not which messages exactly.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/Microservices/FireAndForgetDelivery/MessageBroker.cs
bart-degreed: It already is PostMessageAsync. Based on Send vs Post, a widely known concept of message passing styles (see https://www.codeproject.com/Tips/1029254/SendMessage-and-PostMessage)

test/JsonApiDotNetCoreExampleTests/IntegrationTests/MultiTenancy/MultiTenancyTests.cs
bart-degreed: The proposed names are wrong, because no relationship is involved. It is about resources only. This matches the terminology used in the [json:api spec](https://jsonapi.org/format/1.1/#fetching-resources):
> For example, the following request fetches a collection of articles:
> The following request fetches an article:
> And the following request fetches an article’s author:

test/JsonApiDotNetCoreExampleTests/IntegrationTests/MultiTenancy/MultiTenancyTests.cs
bart-degreed: No, the point is that it excludes _everything_ from other tenants, be it resources, IDs, relationships, meta etc.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/Archiving/ArchiveTests.cs
bart-degreed: It could, but I don't consider it worth the effort. Nor do I think it is better, this is just arbitrary. The tests are about dealing with archiving, so it makes the tests shorter to assume resources are archived. Like you said: if you want to understand what is going on, you need to look at the code. It is the other way around for soft-deletion, neither is right nor wrong.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/ResourceDefinitions/Serialization/SerializationTests.cs
bart-degreed: I'll add the two tests for Get Relationship.

> How is archiving intended to work with adding to and removing from relationships?

This does not apply to filtering logic, because no response is returned. Write operations call `GetForUpdate` which ignores filters, paging, sorting, sparse fieldsets and includes. Therefore adding tests for add-to/remove-from adds nothing. It is irrelevant a resource (left or right) is archived for those operations.

> I don't feel that having a hard distinction between "resource references" (pointers) and resources is very helpful

If you're unwilling to understand that _data_ and _references to data_ are two separate concepts, I cannot help you. And you're mixing up what data is _selected_ vs what _filters_ are applied. Filtering on something that is not selected is quite common.

src/JsonApiDotNetCore/Resources/IResourceDefinition.cs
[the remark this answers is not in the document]
bart-degreed: I don't like both suggestions. The first one is a long, complicated sentence. In the second, my intention is not to explain what 'left' means, just say where it came from.

> The original resource...

This refers to the original version, as opposed to the current version or proposed version in row versioning. See https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.changetracking.entityentry.originalvalues?view=efcore-5.0. It not only matters it was retrieved, but it matters too at which point in time to understand the version. So I don't think original is redundant, it clarifies.

> The resource as retrieved

In full, this would be "The original resource [that was] retrieved from the underlying data store.", but it can be omitted. You're just misreading, this is not wrong.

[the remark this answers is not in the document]
bart-degreed: I'm going to change this to what you suggested:
> The resource as retrieved from the underlying data store. The indication "left" specifies that <paramref name="hasOneRelationship" /> is declared by <typeparamref name="TResource" />.

test/JsonApiDotNetCoreExampleTests/IntegrationTests/MultiTenancy/MultiTenancyTests.cs
bart-degreed: I'm going to replace "_excludes_other_tenants" with "_hides_other_tenants".

test/JsonApiDotNetCoreExampleTests/IntegrationTests/MultiTenancy/MultiTenancyTests.cs
bart-degreed: I'm going to change "secondary_resources" to "secondary_resource_collection".

test/JsonApiDotNetCoreExampleTests/IntegrationTests/Microservices/TransactionalOutboxPattern/OutboxTests.Group.cs
bart-degreed: I'm going to replace `_adds_` with `_writes_`.

ThomasBarnekow:
@bart-degreed, I've looked at this again and it seems it would work for my role-based access control use case. However, I would have one question. Before I can ask that, I need to provide some background.

**Background**: I've now implemented role-based access control, using two extension points. Firstly, **read access** is controlled with resource definitions (but I decided to use those a while after you proposed to look at this PR, because I first tried resource services, which did not work out). I created a base class `AccessControlledResourceDefinition<TResource, TId>` that derives from `JsonApiResourceDefinition<TResource, TId>`. It overrides the `OnApplyFilter(FilterExpfression existingFilter):FilterExpression` method to provide a general read access control framework and, if resource-specific permissions are required, calls an abstract method `AddPermissions():void` to be implemented by concrete resource definitions. Thus, to control read access, subclasses will have to construct collections of `FilterExpression`s. Let's call that the "**First Mechanism**".

Secondly, I then also needed to control write access but, to be honest, did not consider those write callbacks carefully enough. The reason is that, at first, I did not consider resource definitions for read/query access control (see above). Therefore, I first played with resource services (derived from `JsonApiResourceService<TResource, TId>`) but found that those would not provide what I needed because two important base class methods could not be overridden. Therefore, I thought that controllers would be a better extension point because I would have to implement one for each resource anyhow. Therefore, I created a base class called `AccessControlledJsonApiController<TResource, TId>` that derives from `JsonApiController<TResource, TId>`. It requires an `ICommandAccessController` that provides methods like `MayCreate(IAccessControlledResource):bool`. The interface `IAccessControlledResource`  defines properties like `CreatedBy:Guid` (e.g., to identify the authors of a resource). My base class for all resources implements both `IAccessControlledResource` and `IIdentifiable<TId>`.

`MayCreate(IAccessControlledResource):bool`, for example, has access to the `ApplicationUser` resource that represents the authenticated user of the current request. The `ApplicationUser` is injected into the `HttpContext` by middleware that queries the database (using the `DbContext` directly) based on information contained in the access token. The query makes sure all other resources required to make role-based access control decisions are included and accessible via navigation properties. Thus, `MayCreate(IAccessControlledResource):bool` and other methods use Linq expressions to determine whether an `ApplicationUser` has the required `ApplicationUserRole` to perform the respective operation. Let's call that the "**Second Mechanism**".

**Question**: At last, the question. Using resource definitions with write callbacks, is it possible to avoid having to implement two mechanisms, i.e., the First Mechanism that is based on `FilterExpression`s (also a new trick to learn for almost everybody) and the Second Mechanism that is based on Linq expressions?

Based on the access permissions, you might want to return the following results for a GET request:

1. Where the expected result is a resource collection (e.g., `GET /resources`), you would return:
    - 200 OK with zero or more resources to which the user has _resource-specific_ read access, if the user has read access to those resources _in general_ (but there might still not be any resource that is _specifically_ accessible); or
    - 403 Forbidden, if the user does not have the general permission to read resources of that type, for example.
2. Where the expected result is a single resource (e.g., `GET /resources/{id}`), you would return:
    - 200 OK with that single resource, if the user has _resource-specific_ read access; or
    - 403 Forbidden, if the user does not have the _resource-specific_ read access permission.

For the first point, the API would have to include something like:

```csharp
Task BeforeSerializeAsync(ISet<TResource> resources);
```

to let the implementer remove resources from the set, using Linq expressions rather than `FilterExpression`s and, therefore, use a single mechanism instead of two. But I don't know whether or not that is possible. And whether or not that needs to be async could be discussed.

maurei (approved):

maurei:
@ThomasBarnekow  see #1001

PR #986, opened 2021-05-02
Filters: allow has() with a condition

### LINKED ISSUE #985: Filters: allow condition on has() function
## Proposal

Currently, the `has` function in query string filters takes a single parameter: the target collection. For example:

```http
GET /users?filter=has(roles) HTTP/1.1
```

This returns the subset of users that are associated with at least one role. Combining this with an include, we can filter the included set of roles to a specific subset:

```http
GET /users?filter=has(roles)&include=roles&filter[roles]=equals(name,'Administrator') HTTP/1.1
```

This returns the subset of users that are associated with at least one role, then filters the returned roles with name Administrator. The problem is that the resulting subset of customers now may contain an empty set of roles (in case a customer does have a role, but it is not named Administrator).

What we'd like to get instead, is the subset of customers that have at least one role with name Administrator. This proposal enables that by adding an optional condition to the `has` function, which allows to specify a filter. Then we can write:

```http
GET /users?filter=has(roles,equals(name,'Administrator'))&include=roles&filter[roles]=equals(name,'Administrator') HTTP/1.1
```

And this returns the subset of users that are associated with at least one role named Administrator, along with only those roles. So no more customers with an empty included set of roles.

## Implementation

Currently, the `has` function translates to the LINQ `Any()` extension method. An overload of `Any()` exists that takes a predicate. So if the `has` function contains the optional condition, we'll call the overload with predicate instead. EF Core translates that to a `WHERE EXISTS(...)` clause.

bart-degreed:
Adds support for an optional condition in the `has` function within filter query strings.
Also added EBNF grammar for filters to documentation and narrowed some types in the filter parse tree.

Closes #985.

### DIFF
--- a/docs/usage/reading/filtering.md
+++ b/docs/usage/reading/filtering.md
@@ -125,7 +125,7 @@ If you want to use the new filter notation in that case, prefix the parameter va
 GET /articles?filter[caption]=tech&filter=expr:equals(caption,'cooking')) HTTP/1.1
 ```
 
-## Custom Filters
+# Custom Filters
 
 There are multiple ways you can add custom filters:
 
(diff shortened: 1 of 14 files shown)


bart-degreed:
/CC @wayne-o because you asked about this in the Gitter chat [some time ago](https://gitter.im/json-api-dotnet-core/Lobby?at=5fff2ca4787d8f79c8fd1831).

bart-degreed:
@maurei Reverted to Draft, because my code contains a TODO that I still need to investigate.

maurei (approved):

PR #987, opened 2021-05-05
Rewrite of LinkBuilder to use ASP.NET Core routing to render links

### LINKED ISSUE #956: Incorrect links when using custom route
When only some endpoints have a custom route, rendering of the links for included resources is incorrect. See [here](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/02496277f072ccecd574449126700bbfa17ed34d/test/JsonApiDotNetCoreExampleTests/IntegrationTests/HostingInIIS/HostingTests.cs#L60) and [here](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/02496277f072ccecd574449126700bbfa17ed34d/test/JsonApiDotNetCoreExampleTests/IntegrationTests/HostingInIIS/HostingTests.cs#L106).

This likely can be solved by using the route dictionary for rendering links, as proposed in #953.

bart-degreed:
The limitation that custom routes must end in the public resource name no longer applies.

- Fixed: Resource-level Self links in atomic:operations responses are now hidden when no controller exists for the resource type.
- Fixed: For determining which links to render, settings from primary resource were used on secondary endpoints. For example, if you configure Customer to show all links, but Orders no show none, then /customer/1/orders would show all links.
- Optimization: Compound `page[size]` parameter value (example: `10,articles:5`) is calculated once, instead of again for each pagination link.
- Deprecated: `IJsonApiRequest.BasePath`. This information is no longer needed, but we still set it for back-compat.
- Added support for non-standard route parameters in links, for example: `[DisableRoutingConvention, Route("{shopName}/products")]`

Fixes #956.

#### QUALITY CHECKLIST
- [x] Adapted tests
- [x] Documentation updated
- [ ] N/A: Created issue to update [Templates](https://github.com/json-api-dotnet/Templates/issues/new): {ISSUE_NUMBER}


bart-degreed:
@alastairtree can you give this a try?

maurei (approved):

PR #992, opened 2021-05-07
Adds option to emit jsonapi version in response documents

### LINKED ISSUE #980: Set json-api version in responses
In responses, we should include the json:api [version](https://jsonapi.org/format/1.1/#document-jsonapi-object) it adheres to, including extensions where applicable.

> version - whose value is a string indicating the highest JSON:API version supported.
If the version member is not present, clients should assume the server implements at least version 1.0 of the specification.
Note: Because JSON:API is committed to making additive changes only, the version string primarily indicates which new features a server may support.

bart-degreed:
Fixes #980.

### DIFF
--- a/src/JsonApiDotNetCore/Configuration/IJsonApiOptions.cs
+++ b/src/JsonApiDotNetCore/Configuration/IJsonApiOptions.cs
@@ -35,6 +35,11 @@ namespace JsonApiDotNetCore.Configuration
         AttrCapabilities DefaultAttrCapabilities { get; }
 
         /// <summary>
+        /// Indicates whether responses should contain a jsonapi object that contains the highest JSON:API version supported. False by default.
+        /// </summary>
+        bool IncludeJsonApiVersion { get; }
+
+        /// <summary>
         /// Whether or not <see cref="Exception" /> stack traces should be serialized in <see cref="ErrorMeta" /> objects. False by default.
         /// </summary>
         bool IncludeExceptionStackTraceInErrors { get; }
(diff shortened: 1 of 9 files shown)


src/JsonApiDotNetCore/Configuration/IJsonApiOptions.cs
```diff
@@ -34,6 +34,11 @@ internal NamingStrategy SerializerNamingStrategy
         /// </summary>
         AttrCapabilities DefaultAttrCapabilities { get; }
 
+        /// <summary>
+        /// Indicates whether responses should contain a jsonapi object that contains the highest JSON:API version supported. False by default.
```
maurei, line 38: We don't support serving different versions of the JSON:API spec and I think it is unlikely we will in the near future, but just in case we do, it wouldn't hurt to keep this a little more generic and use an enum that has a single value

```c#
public enum JsonApiVersion 
{
   Latest
}
```

bart-degreed: I tend to think that the json:api version is _not_ negotiable between server and client. The spec describes that top-level members can occur in requests and responses, but then describes Jsonapi (which contains the version) as _"an object describing the server’s implementation"_.

So even if JADNC would implement support for serving multiple json:api versions in the future, an API developer would need to guess what a client needs. In practice, I think the following would happen:
1. A new json:api version is published (JADNC still returns the old version)
2. JADNC is updated to return the new version (indicating compliance with new features)
3. API developers update to the latest JADNC and instruct their API clients on how to adapt

So I don't think there's a point in ever implementing support for multiple json:api versions in a single JADNC release. If that changes, we can add an extra property to select the version, aside from the boolean flag. That allows to choose version compliance, apart for _emitting_ that info to clients.

maurei (approved):

PR #998, opened 2021-05-14
Adds support for ETag header in GET and HEAD requests

### LINKED ISSUE #933: Caching: Add ETag support to save bandwidth
One of the said advantages of using REST is it uses the existing rich feature set of the HTTP standard, which is widely supported by various clients, servers and intermediate proxies. One of those features is cache control to reduce network traffic.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag for an introduction to ETags.

Note the confusing terminology here: a 'resource' in HTTP refers to the full response body, while a 'resource' in json:api refers to a *section* within the response body. For example, a HTTP resource could contain a collection of json:api resources, followed by a collection of related json:api resources. Or even partial json:api resources, when sparse fieldsets are used.

The use-case here concerns reducing network traffic. Not optimistic concurrency (due to the above mismatch). And not reducing database pressure (see below).

For example, consider a mobile client that wants to refresh its locally cached version of a json:api resource. The server sends back "Not Modified" if it determines the resource is still fresh. If stale, it sends back the full json:api resource. Likewise, the mobile client can check periodically whether a subset of the json:api resource attributes has changed, by repeatedly requesting it with a sparse fieldset. The same applies for a collection of resources: if the collection is unchanged since last time, the server does not need to send it again.

This proposal optimizes network traffic, not database pressure. For each request, the server still needs to fetch the data, in order to determine if it has changed. When caching database results is desired, developers can already implement their own `IResourceRepository` and apply various caching strategies there. It is not the scope of this proposal.

To implement this, JADNC sends back a strong ETag header on GET responses, which contains an MD5 hash of the binary response body. For an incoming GET request with an ETag, the server renders a response internally and calculates its MD5 hash. When both hashes are identical, the server returns Not Modified with an empty body.

If calculating the hash adds substantial overhead, we should put it behind an option that is disabled by default.

Example:
```http
GET /api/blogs/1/articles?include=author&fields[articles]=title HTTP/1.1

HTTP/1.1 200 OK
ETag: "51142bc1-7449-479b075b2891b"

{
  "data": [{ ... }],
  "included": [{ ... }]
}
```

Then, after some time....

```http
GET /api/blogs/1/articles?include=author&fields[articles]=title HTTP/1.1
ETag: "51142bc1-7449-479b075b2891b"

HTTP/1.1 304 Not Modified
```

bart-degreed:
See documentation updates for details.

Closes #933.

### DIFF
--- a/docs/usage/caching.md
+++ b/docs/usage/caching.md
@@ -1,0 +1,75 @@
+# Caching with ETags
+
+_since v4.2_
+
+GET requests return an [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) HTTP header, which can be used by the client in subsequent requests to save network bandwidth.
+
+Be aware that the returned ETag represents the entire response body (a 'resource' in HTTP terminology) for a request URL that includes the query string.
+This is unrelated to JSON:API resources. Therefore, we do not use ETags for optimistic concurrency.
+
+Getting a list of resources returns an ETag:
+
+```http
+GET /articles?sort=-lastModifiedAt HTTP/1.1
+Host: localhost:5000
+```
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json
+Server: Kestrel
+Transfer-Encoding: chunked
+ETag: "7FFF010786E2CE8FC901896E83870E00"
+
+{
+  "data": [ ... ]
+}
+```
+
+The request is later resent using the received ETag. The server data has not changed at this point.
+
+```http
+GET /articles?sort=-lastModifiedAt HTTP/1.1
+Host: localhost:5000
+If-None-Match: "7FFF010786E2CE8FC901896E83870E00"
+```
+
+```http
+HTTP/1.1 304 Not Modified
+Server: Kestrel
+ETag: "7FFF010786E2CE8FC901896E83870E00"
+```
+
+After some time, the server data has changed.
+
+```http
+GET /articles?sort=-lastModifiedAt HTTP/1.1
+Host: localhost:5000
+If-None-Match: "7FFF010786E2CE8FC901896E83870E00"
+```
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json
+Server: Kestrel
+Transfer-Encoding: chunked
+ETag: "356075D903B8FE8D9921201A7E7CD3F9"
+
+{
+  "data": [ ... ]
+}
+```
+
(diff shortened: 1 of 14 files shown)


docs/usage/caching.md
```diff
@@ -0,0 +1,75 @@
+# Caching with ETags
+
+_since v4.2_
+
+GET requests return an [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) HTTP header, which can be used by the client in subsequent requests to save network bandwidth.
+
+Be aware that the returned ETag represents the entire response body (a 'resource' in HTTP terminology) for a request URL that includes the query string.
```
maurei, line 7: Have you given weak validations a thought?  Are there situation in which we would benefit from having weak ETags? Eg, could these be considered semantically equivalent?

```json
{
  "meta": {
    "totalResources": 3
  },
  "links": {
    "self": "/api/books?page[size]=1&page[numer]=1",
    "first": "/api/books?page[size]=1&page[numer]=1",
    "last": "/api/books?page[size]=1&page[numer]=3"
  },
  "data": [
    {
      "type": "books",
      "id": "1",
      "attributes": {
        "title": "Frankenstein",
        "publishYear": 1818
      },
      "relationships": {
        "author": {
          "links": {
            "self": "/api/books/1/relationships/author",
            "related": "/api/books/1/author"
          }
        }
      },
      "links": {
        "self": "/api/books/1"
      }
    }
  ]
}
```
versus (mind the difference: totalRecords)
```json
{
  "meta": {
    "totalResources": 10
  },
  "links": {
    "self": "/api/books?page[size]=1&page[numer]=1",
    "first": "/api/books?page[size]=1&page[numer]=1",
    "last": "/api/books?page[size]=1&page[numer]=10"
  },
  "data": [
    {
      "type": "books",
      "id": "1",
      "attributes": {
        "title": "Frankenstein",
        "publishYear": 1818
      },
      "relationships": {
        "author": {
          "links": {
            "self": "/api/books/1/relationships/author",
            "related": "/api/books/1/author"
          }
        }
      },
      "links": {
        "self": "/api/books/1"
      }
    }
  ]
}
```

bart-degreed: I have looked at it, but I don't think they are useful in our case. The number of total records is actual information, so if that changes then caches should be invalidated. A byte-range difference with the same semantic meaning would be tabs vs spaces in the JSON, for example. I can't think of cases where we'd want to do that.

However the signature of `IETagGenerator` allows them to be produced.

maurei (approved):

PR #1021, opened 2021-06-22
Fix exception propagation

### LINKED ISSUE #1017: Consecutive exception hides original error
For example, see [here](https://ci.appveyor.com/project/json-api-dotnet/jsonapidotnetcore/branch/master/job/4a6c29p4n9wdhted#L194). The initial `DataStoreUpdateException` disappears because a new exception is thrown when disposing `PlaceholderResourceCollector` before the first exception can be caught. This results in a 500 error response plus a log entry at error level. Instead the follow-up exception should be swallowed so that the original `DataStoreUpdateException` can propagate.

bart-degreed:
Fixes #1017.


bart-degreed:
I considered adding a note to our documentation on HasOne, but it's not entirely clear to me when EF Core chooses to create an identifying foreign key. It does not always do that (see commit details).

PR #1041, opened 2021-08-18
Feature/allow leading space

### LINKED ISSUE #1040: Leading spaces in literals
_Please read our [Contributing Guides](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/.github/CONTRIBUTING.md) before submitting a bug._

#### DESCRIPTION
_A clear and concise description of what the bug is._

The parser for literals does not allow leading spaces (but it does allow trailing spaces). Generally a literal should allow whatever the literal value is.

For example, you might want to search a name field that ends with a particular surname, starting with a space delimiter, e.g. endsWith(name,' Surname'), or a field that contains a space.

**Note:** There is an existing test, in `LegacyFilterParseTests` that check a single space in a literal throws an Unexpected whitespace exception, so perhaps this check was intentional. Maybe for an 'equals()' test it is a bit strange to check for a single space (or if the literal is a numeric field), but something like contains(name,' ') to look for something containing a space might make more sense.

#### STEPS TO REPRODUCE

1. Test a search filter with a trailing space works, e.g. /webAccounts?filter=startsWith(name,'Given ')
  * The encoded value would be /webAccounts?filter=startsWith(name,%27Given%20%27)
2. Test a search filter with only a space, e.g. /webAccounts?filter=contains(name,' ')
  * The encoded value would be /webAccounts?filter=contains(name,%27%20%27)

Tests are provided in an example branch:

        [InlineData("filter", "startsWith(name,'GivenName ')", null, "startsWith(name,'GivenName ')")]
        [InlineData("filter", "endsWith(name,' Surname')", null, "endsWith(name,' Surname')")]
        [InlineData("filter", "contains(name,' ')", null, "contains(name,' ')")]

#### EXPECTED BEHAVIOR
_A clear and concise description of what you expected to happen._

Literal values should be treated as is, even if they contain leading whitespace.

#### ACTUAL BEHAVIOR
_A clear and concise description of what happens instead._

Leading whitespace, including a literal of whitespace only, returns the error "Unexpected whitespace.".

#### VERSIONS USED
- JsonApiDotNetCore version: v4.2.0 +10 (current head)
- ASP.NET Core version:
- Entity Framework Core version:
- Database provider:

sgryphon:
Closes #1040

#### QUALITY CHECKLIST
- [X] Adapted tests
- [ ] Documentation updated
- [ ] Created issue to update [Templates](https://github.com/json-api-dotnet/Templates/issues/new): {ISSUE_NUMBER}


test/JsonApiDotNetCoreExampleTests/UnitTests/QueryStringParameters/FilterParseTests.cs
```diff
@@ -115,6 +115,9 @@ public void Reader_Read_Fails(string parameterName, string parameterValue, strin
         [InlineData("filter", "equals(title,'Brian O''Quote')", null, "equals(title,'Brian O''Quote')")]
         [InlineData("filter", "equals(title,'!@#$%^&*()-_=+\"''[]{}<>()/|\\:;.,`~')", null, "equals(title,'!@#$%^&*()-_=+\"''[]{}<>()/|\\:;.,`~')")]
         [InlineData("filter", "equals(title,'')", null, "equals(title,'')")]
+        [InlineData("filter", "startsWith(name,'GivenName ')", null, "startsWith(name,'GivenName ')")]
+        [InlineData("filter", "endsWith(name,' Surname')", null, "endsWith(name,' Surname')")]
+        [InlineData("filter", "contains(name,' ')", null, "contains(name,' ')")]
```
maurei, line 120: These tests don't pass because `name` is not does exist on the blog resource. I suggest to use the following:

```c#
[InlineData("filter", "startsWith(title,'Brian ')", null, "startsWith(title,'Brian ')")]
[InlineData("filter", "endsWith(title,' O''Quote')", null, "endsWith(title,' O''Quote')")]
[InlineData("filter", "contains(title,' ')", null, "contains(title,' ')")]
```

test/JsonApiDotNetCoreExampleTests/UnitTests/QueryStringParameters/LegacyFilterParseTests.cs
```diff
@@ -77,6 +76,7 @@ public void Reader_Read_Fails(string parameterName, string parameterValue, strin
         [InlineData("filter[caption]", "unknown:some", "equals(caption,'unknown:some')")]
         [InlineData("filter[author.userName]", "Jack", "equals(author.userName,'Jack')")]
         [InlineData("filter", "expr:equals(caption,'some')", "equals(caption,'some')")]
+        [InlineData("filter", "expr:equals(caption,' ')", "equals(caption,' ')")]
```
maurei, line 79: Instead of this test, I think the following would be more useful:

```c#
[InlineData("filter[caption]", " ", "equals(caption,' ')")]
```

test/JsonApiDotNetCoreExampleTests/UnitTests/QueryStringParameters/FilterParseTests.cs
maurei: Additionally, to hit the `unexpected whitespace` error in the adjusted code path, I would suggest to add the following test:

```c#
[InlineData("filter", "equals(title, )", "Unexpected whitespace.")]
```

maurei (changes requested):
Thanks for your input!  @bart-degreed is on vacation currently, I'd like to have him have a look at it too once he returns. But for the meanwhile I've left some comments for you to look into.

test/JsonApiDotNetCoreExampleTests/UnitTests/QueryStringParameters/FilterParseTests.cs
bart-degreed: > These tests don't pass because `name` is not does exist on the blog resource. I suggest to use the following:
> 
> ```cs
> [InlineData("filter", "startsWith(title,'Brian ')", null, "startsWith(title,'Brian ')")]
> [InlineData("filter", "endsWith(title,' O''Quote')", null, "endsWith(title,' O''Quote')")]
> [InlineData("filter", "contains(title,' ')", null, "contains(title,' ')")]
> ```

Yes, this needs to use the `title` field. However, I'd keep the values as simple as possible, as in: " Some", "Some " and not intermix this with escaped quotes.

> Additionally, to hit the `unexpected whitespace` error in the adjusted code path, I would suggest to add the following test:
> 
> ```cs
> [InlineData("filter", "equals(title, )", "Unexpected whitespace.")]
> ```

Unexpected whitespace is already covered by an existing case:
```c#
[InlineData("filter", " ", "Unexpected whitespace.")]
```

test/JsonApiDotNetCoreExampleTests/UnitTests/QueryStringParameters/LegacyFilterParseTests.cs
bart-degreed: Agreed. The "expr:" prefix executes the exact same code which is covered by FilterParseTests, so it can be left out here. The test that was removed in this PR:
```c#
[InlineData("filter[author]", " ", "Unexpected whitespace.")]
```
was an invalid test, because it is perfectly fine to use unescaped whitespace in legacy syntax.

test/JsonApiDotNetCoreExampleTests/UnitTests/QueryStringParameters/FilterParseTests.cs
sgryphon: I ended up using owner.displayName, so it is a concrete example of why someone might want to search for a leading space (e.g. you want a particular last name). Although I guess "containing a space" would work for titles as well. Yeah, don't want to mix up with quotes. 

Happy to still change to Some, but I think it is less meaningful.

I ended up adding the extra unexpected whitespace test, as it is not at the start of the whole expression, but the start of one token.

test/JsonApiDotNetCoreExampleTests/UnitTests/QueryStringParameters/LegacyFilterParseTests.cs
sgryphon: Legacy test changed.

bart-degreed (approved):
Looks good to me. Thanks for your contribution!

maurei (approved):

Issue #1079, opened 2021-09-10
Cannot perform `any` filter operation on enum collection attribute


Bezz242:
#### DESCRIPTION
Cannot filter on enum attribute. Syntax not quite clear, error misses information.

#### STEPS TO REPRODUCE
I have an entity Users with an attribute Role, which is an enum of [Admin, Doctor...].
In Postman, a simple GET `/users` will return a collection of all users.

I want to filter on Role.

I use the GET with url `/users?filter=any(roles,'Admin')`

In C# the roles are

```
        public enum Role
        {
            Admin,
            Doctor,
            Nurse,
            Assistant
        }
```

The user:

```    
    public class User : Identifiable<Guid>, IManagedEntity
    {
        [Key]
        public override Guid Id { get; set; }

        [Attr]
        [Required]
        public string Email { get; set; }

        [Attr]
        public bool EmailConfirmed { get; set; }
        
        [Attr(Capabilities = AttrCapabilities.AllowChange), NotMapped]
        public string Password { get; set; }
        
        [JsonIgnore]
        public string? PasswordHash { get; set; }

        [Attr]
        public Role[]? Roles { get; set; }
    }  
```

#### EXPECTED BEHAVIOR
When I do a GET with the correct syntax for a filter on enum Role, I want to get a collection of users that has the specific Role. 

#### ACTUAL BEHAVIOR
Postman receives an error message:

```
{
    "errors": [
        {
            "id": "ea9d21c3-d86f-4228-aaca-9afb124ee0da",
            "status": "400",
            "title": "The specified filter is invalid.",
            "detail": ", expected.",
            "source": {
                "parameter": "filter"
            }
        }
    ]
}
```
_The detail field appears to be missing information._

Logging in IDE (Rider) shows:

```
info: JsonApiDotNetCore.Middleware.ExceptionHandler[0]
      Errors = [
        {
          "Id": "ea9d21c3-d86f-4228-aaca-9afb124ee0da",
          "Status": "400",
          "Title": "The specified filter is invalid.",
          "Detail": ", expected.",
          "Source": {
            "Parameter": "filter"
          }
        }
      ]
      JsonApiDotNetCore.Errors.InvalidQueryStringParameterException: Errors = [
        {
          "Id": "ea9d21c3-d86f-4228-aaca-9afb124ee0da",
          "Status": "400",
          "Title": "The specified filter is invalid.",
          "Detail": ", expected.",
          "Source": {
[...]

PR #1081, opened 2021-09-11
Repro #1079

maurei:
Reproduces the issue reported in #1079.


bart-degreed:
This is by design, see https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/1079#issuecomment-917636387.

PR #1122, opened 2021-12-02
Update to .NET 6 with EF Core 6

### LINKED ISSUE #1109: Version 5 for .NET 6 Support
I can see in the README that this package offers .NET 6 support, however upon looking through nuget and the released on GitHub there is no version 5, I also checked branches just to be sure and can't see anything on there.

### LINKED ISSUE #1121: Breaking changes in PostgreSQL provider for EF Core 6
The release notes for Npgsql 6 list [breaking changes](https://www.npgsql.org/efcore/release-notes/6.0.html?tabs=annotations).

> Npgsql 6.0 brings some major breaking changes and is not a simple in-place upgrade. Carefully read the breaking change notes below and upgrade with care.

The design change in handling `DateTime` and `DateTimeOffset` requires little changes in JADNC itself, but it mostly affects API clients. And because of that, our suite of integration tests.

As explained in https://www.roji.org/postgresql-dotnet-timestamp-mapping, the PostgreSQL column types "timestamp without time zone" (timestamp) and "timestamp with time zone" (timestamptz) both do NOT store time zone information and both do NOT store a UTC offset. They store solely a timestamp, which is always in UTC. This means that a `DateTimeOffset` with a non-zero UTC offset cannot be stored in these columns.

The old provider would auto-convert offsets to UTC, as well as convert between local dates and UTC dates (based on the time zone that the code is executing in), which led to a lot of confusion. The new provider fails when using any `DateTime`/`DateTimeOffset` that is not explicitly in UTC when mapped to column type timestamptz (the default). Alternatively, when the column type is timestamp, it only accepts local values.

As explained in the blog post, using UTC offsets without a time zone is a flawed design, resulting in subtle bugs. For details, read https://blog.nodatime.org/2011/08/what-wrong-with-datetime-anyway.html. And given that these columns cannot store an offset, it makes sense to optimize for the "UTC everywhere" pattern.

For JSON:API clients, this means they must now send a local time (example: "2001-02-03T04:05:06") for columns of type timestamp. And a time in UTC with offset 0 (examples: "2001-02-03T04:05:06Z", "2001-02-03T04:05:06+0:00") for columns of type timestamptz. Anything else will be rejected. This puts more burden on clients, who now have to account for these restrictions. It's tempting to adapt JADNC to do the conversion that the provider used to do. But I think the new behavior makes more sense for JADNC too. It greatly simplifies the behavior of JSON:API servers: dates are either in "the" local time zone (whatever that may be), or dates are UTC timestamps. The API developer chooses per column between these, and the client must adapt to that. Ultimately the client is in the best position to do conversions, if needed, based on the time zone where it resides.

Since the JSON:API server does not know in which time zone a client lives and the offset cannot be stored, it cannot roundtrip an incoming date with offset +5 in the response to the client. Also, the time zone of the JSON:API server and the time zone of the database server should not influence how the API handles time. The old behavior would get it wrong half of the year when DST is in effect.

bart-degreed:
Overview:
- Updates JADNC to .NET 6, utilizing [Minimal Hosting APIs](https://docs.microsoft.com/en-us/aspnet/core/tutorials/min-web-api?view=aspnetcore-6.0&tabs=visual-studio) (using controllers) in examples and documentation.
- Updates JADNC to EF Core 6, adapting to [breaking changes in the PostgreSQL provider](https://www.npgsql.org/efcore/release-notes/6.0.html?tabs=annotations).
- Added `[NoResource]` to suppress startup warning that entity does not implement `IIdentifiable`.

The cibuild fails because it times out, caused by the large diff. I've verified locally and all looks good. Green build (with cleanupcode/inspectcode disabled) [here](https://ci.appveyor.com/project/json-api-dotnet/jsonapidotnetcore/builds/41814018).

Closes #1109.
Closes #1121.

#### QUALITY CHECKLIST
- [x] Adapted tests
- [x] Documentation updated
- [ ] N/A: Created issue to update [Templates](https://github.com/json-api-dotnet/Templates/issues/new): {ISSUE_NUMBER}


src/JsonApiDotNetCore/Configuration/JsonApiApplicationBuilder.cs
```diff
@@ -182,29 +182,29 @@ private void AddMiddlewareLayer()
 
     private void AddResourceLayer()
     {
-        RegisterImplementationForOpenInterfaces(ServiceDiscoveryFacade.ResourceDefinitionInterfaces, typeof(JsonApiResourceDefinition<,>));
+        RegisterImplementationForOpenInterfaces(ServiceDiscoveryFacade.ResourceDefinitionOpenInterfaces, typeof(JsonApiResourceDefinition<,>));
```
maurei, line 185: Should be `Unbound`  instead of `Open`

src/JsonApiDotNetCore/Configuration/JsonApiApplicationBuilder.cs
```diff
@@ -182,29 +182,29 @@ private void AddMiddlewareLayer()
...
 
         _services.AddScoped<IResourceDefinitionAccessor, ResourceDefinitionAccessor>();
         _services.AddScoped<IResourceFactory, ResourceFactory>();
     }
 
     private void AddRepositoryLayer()
     {
-        RegisterImplementationForOpenInterfaces(ServiceDiscoveryFacade.RepositoryInterfaces, typeof(EntityFrameworkCoreRepository<,>));
+        RegisterImplementationForOpenInterfaces(ServiceDiscoveryFacade.RepositoryOpenInterfaces, typeof(EntityFrameworkCoreRepository<,>));
```
maurei, line 193: Should be `Unbound`  instead of `Open`

src/JsonApiDotNetCore/Configuration/JsonApiApplicationBuilder.cs
```diff
@@ -182,29 +182,29 @@ private void AddMiddlewareLayer()
...
 
         _services.AddScoped<IResourceRepositoryAccessor, ResourceRepositoryAccessor>();
     }
 
     private void AddServiceLayer()
     {
-        RegisterImplementationForOpenInterfaces(ServiceDiscoveryFacade.ServiceInterfaces, typeof(JsonApiResourceService<,>));
+        RegisterImplementationForOpenInterfaces(ServiceDiscoveryFacade.ServiceOpenInterfaces, typeof(JsonApiResourceService<,>));
```
maurei, line 200: Should be `Unbound`  instead of `Open`

src/JsonApiDotNetCore/Configuration/JsonApiApplicationBuilder.cs
```diff
@@ -182,29 +182,29 @@ private void AddMiddlewareLayer()
...
     }
 
-    private void RegisterImplementationForOpenInterfaces(HashSet<Type> openGenericInterfaces, Type implementationType)
+    private void RegisterImplementationForOpenInterfaces(HashSet<Type> openInterfaces, Type implementationType)
```
maurei, line 203: Should be `Unbound`  instead of `Open`

src/JsonApiDotNetCore/Configuration/ServiceCollectionExtensions.cs
```diff
@@ -85,25 +85,25 @@ public static IServiceCollection AddResourceDefinition<TResourceDefinition>(this
...
         ArgumentGuard.NotNull(services, nameof(services));
 
-        RegisterForConstructedType(services, typeof(TResourceDefinition), ServiceDiscoveryFacade.ResourceDefinitionInterfaces);
+        RegisterTypeForOpenInterfaces(services, typeof(TResourceDefinition), ServiceDiscoveryFacade.ResourceDefinitionOpenInterfaces);
 
         return services;
     }
 
-    private static void RegisterForConstructedType(IServiceCollection services, Type implementationType, IEnumerable<Type> openGenericInterfaces)
+    private static void RegisterTypeForOpenInterfaces(IServiceCollection serviceCollection, Type implementationType, IEnumerable<Type> openInterfaces)
```
maurei, line 93: Should be `Unbound`  instead of `Open`

src/JsonApiDotNetCore/Configuration/ServiceDiscoveryFacade.cs
```diff
@@ -15,7 +15,7 @@ namespace JsonApiDotNetCore.Configuration;
 [PublicAPI]
 public sealed class ServiceDiscoveryFacade
 {
-    internal static readonly HashSet<Type> ServiceInterfaces = new()
+    internal static readonly HashSet<Type> ServiceOpenInterfaces = new()
```
maurei, line 18: Should be `Unbound`  instead of `Open`

src/JsonApiDotNetCore/Configuration/ServiceDiscoveryFacade.cs
```diff
@@ -32,14 +32,14 @@ public sealed class ServiceDiscoveryFacade
         typeof(IRemoveFromRelationshipService<,>)
     };
 
-    internal static readonly HashSet<Type> RepositoryInterfaces = new()
+    internal static readonly HashSet<Type> RepositoryOpenInterfaces = new()
```
maurei, line 35: Should be `Unbound`  instead of `Open`

src/JsonApiDotNetCore/Configuration/ServiceDiscoveryFacade.cs
```diff
@@ -32,14 +32,14 @@ public sealed class ServiceDiscoveryFacade
...
     {
         typeof(IResourceRepository<,>),
         typeof(IResourceWriteRepository<,>),
         typeof(IResourceReadRepository<,>)
     };
 
-    internal static readonly HashSet<Type> ResourceDefinitionInterfaces = new()
+    internal static readonly HashSet<Type> ResourceDefinitionOpenInterfaces = new()
```
maurei, line 42: Should be `Unbound`  instead of `Open`

src/JsonApiDotNetCore/Configuration/TypeLocator.cs
```diff
@@ -38,62 +46,62 @@ internal sealed class TypeLocator
...
+    /// Gets the implementation type with service interface (to be registered in the IoC container) for the specified open interface and its type arguments,
+    /// by scanning for types in the specified assembly that match the signature.
     /// </summary>
     /// <param name="assembly">
-    /// The assembly to search in.
+    /// The assembly to search for matching types.
     /// </param>
-    /// <param name="openGenericInterface">
-    /// The open generic interface.
+    /// <param name="openInterface">
```
maurei, line 55: Should be `Unbound`  instead of `Open`

src/JsonApiDotNetCore/Configuration/TypeLocator.cs
```diff
@@ -103,30 +111,30 @@ internal sealed class TypeLocator
...
-    /// Gets all derivatives of the concrete, generic type.
+    /// Scans for types in the specified assembly that derive from the specified open type.
     /// </summary>
     /// <param name="assembly">
-    /// The assembly to search.
+    /// The assembly to search for derived types.
     /// </param>
-    /// <param name="openGenericType">
-    /// The open generic type, e.g. `typeof(ResourceDefinition&lt;&gt;)`.
+    /// <param name="openType">
```
maurei, line 119: Should be `Unbound`  instead of `Open`

src/JsonApiDotNetCore/Middleware/JsonApiRoutingConvention.cs
```diff
@@ -146,14 +146,14 @@ private string TemplateFromController(ControllerModel model)
...
     /// <summary>
-    /// Determines the resource associated to a controller by inspecting generic arguments in its inheritance tree.
+    /// Determines the resource type associated to a controller by inspecting generic type arguments in its inheritance tree.
     /// </summary>
-    private Type? ExtractResourceClrTypeFromController(Type type)
+    private Type? ExtractResourceClrTypeFromController(Type controllerType)
     {
         Type aspNetControllerType = typeof(ControllerBase);
         Type coreControllerType = typeof(CoreJsonApiController);
         Type baseControllerType = typeof(BaseJsonApiController<,>);
```
maurei, line 155: Should be `baseControllerUnboundType`

src/JsonApiDotNetCore/Queries/Internal/SystemExpressionExtensions.cs
```diff
@@ -20,14 +20,14 @@ public static Expression CreateTupleAccessExpressionForConstant(this object? val
         // Which represents the next C# code:
         //   Tuple.Create<T>(value).Item1;
 
-        MethodInfo tupleCreateMethod = typeof(Tuple).GetMethods()
+        MethodInfo tupleCreateOpenMethod = typeof(Tuple).GetMethods()
```
maurei, line 23: Should be `Unbound`  instead of `Open`

test/UnitTests/Graph/TypeLocatorTests.cs
```diff
@@ -9,45 +9,40 @@ namespace UnitTests.Graph;
...
+    public void GetContainerRegistrationFromAssembly_Gets_Implementation()
     {
         // Arrange
         Assembly assembly = GetType().Assembly;
-        Type openGeneric = typeof(IGenericInterface<>);
-        Type genericArg = typeof(int);
-
-        Type expectedImplementation = typeof(Implementation);
-        Type expectedInterface = typeof(IGenericInterface<int>);
+        Type openInterface = typeof(IGenericInterface<>);
```
maurei, line 16: Should be `Unbound`  instead of `Open`

test/UnitTests/Graph/TypeLocatorTests.cs
```diff
@@ -9,45 +9,40 @@ namespace UnitTests.Graph;
...
-    public void GetDerivedGenericTypes_Gets_Implementation()
+    public void GetDerivedTypesForOpenType_Gets_Implementation()
     {
         // Arrange
         Assembly assembly = GetType().Assembly;
-        Type openGeneric = typeof(BaseType<>);
-        Type genericArg = typeof(int);
-
-        Type expectedImplementation = typeof(DerivedType);
+        Type openType = typeof(BaseType<>);
```
maurei, line 35: Should be `Unbound`  instead of `Open`

maurei (approved):

PR #1123, opened 2021-12-06
OpenAPI: Generalization of naming conventions

### LINKED ISSUE #1080: Generalized support for naming strategies in schema ids
In the private repository we used a kebabcase naming convention, and the `JsonApiSchemaIdSelector` is configured to support only that. This should be generalized by using the configured naming strategy on `JsonApiOptions`.

maurei:
Closes #1080


src/JsonApiDotNetCore.OpenApi/StringExtensions.cs
```diff
@@ -0,0 +1,11 @@
+using System.Text.RegularExpressions;
+
+namespace JsonApiDotNetCore.OpenApi;
+
+internal static class StringExtensions
+{
+    public static string Pascalize(this string source)
```
bart-degreed, line 7: It's _very_ confusing to add a method with an identical signature as in Humanizer, but with slightly different behavior. There's no way to see at the call side which one is being used.

src/JsonApiDotNetCore.OpenApi/StringExtensions.cs
```diff
@@ -0,0 +1,11 @@
...
+    {
+        return Regex.Replace(source, "(?:^|-|_| +)(.)", match => match.Groups[1].Value.ToUpper());
```
bart-degreed, line 9: Perf: Create a static readonly `Regex` instance, passing in `RegexOptions.Compiled`.

test/OpenApiClientTests/LegacyClient/ClientAttributeRegistrationLifeTimeTests.cs
```diff
@@ -22,7 +22,7 @@ public async Task Disposed_attribute_registration_for_document_does_not_affect_r
             Data = new AirplaneDataInPatchRequest
             {
                 Id = airplaneId,
-                Type = AirplanesResourceType.Airplanes,
+                Type = AirplaneResourceType.Airplanes,
```
bart-degreed, line 25: Why was this changed to singular? From an API consumer perspective, there is no such thing as an "airplane" resource type. There's just the "airplanes" resource type. This "type" field should be unrelated to generated CLR type names and endpoint method names.

test/OpenApiClientTests/OpenApiClientTests.csproj
```diff
@@ -42,5 +42,25 @@
...
+      <OutputPath>PascalCaseClient.cs</OutputPath>
+      <CodeGenerator>NSwagCSharp</CodeGenerator>
+      <Options>/UseBaseUrl:false /ClientClassAccessModifier:internal /GenerateExceptionClasses:false /AdditionalNamespaceUsages:JsonApiDotNetCore.OpenApi.Client.Exceptions</Options>
+    </OpenApiReference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="xunit.runner.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
```
bart-degreed, line 64: Please put this above external dependencies, like in the other test project files.

test/OpenApiTests/LegacyOpenApiIntegration/swagger.json
```diff
@@ -2064,7 +2064,7 @@
         "type": "object",
         "properties": {
           "type": {
-            "$ref": "#/components/schemas/airplanes-resource-type"
+            "$ref": "#/components/schemas/airplane-resource-type"
```
bart-degreed, line 2067: I think this should remain plural as well.

test/OpenApiClientTests/OpenApiClientTests.csproj
```diff
@@ -31,19 +31,36 @@
...
+      <CodeGenerator>NSwagCSharp</CodeGenerator>
+      <Options>/UseBaseUrl:false /ClientClassAccessModifier:internal /GenerateExceptionClasses:false /AdditionalNamespaceUsages:JsonApiDotNetCore.OpenApi.Client.Exceptions</Options>
     </OpenApiReference>
   </ItemGroup>
 
   <ItemGroup>
-    <EmbeddedResource Include="..\OpenApiTests\LegacyOpenApiIntegration\swagger.json" />
+    <None Update="xunit.runner.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
```
bart-degreed, line 64: Please put this above external dependencies, like in the other test project files.

test/OpenApiTests/OpenApiTests.csproj
```diff
@@ -14,9 +13,16 @@
...
 
   <ItemGroup>
     <EmbeddedResource Include="LegacyOpenApiIntegration\swagger.json" />
   </ItemGroup>
+  
+  <ItemGroup>
+    <None Update="xunit.runner.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
```
bart-degreed, line 27: Please put this above external dependencies, like in the other test project files.

src/JsonApiDotNetCore.OpenApi/ServiceCollectionExtensions.cs
```diff
@@ -41,6 +41,7 @@ private static void AddCustomApiExplorer(IServiceCollection services, IMvcCoreBu
             var apiDescriptionProviders = provider.GetRequiredService<IEnumerable<IApiDescriptionProvider>>();
 
             JsonApiActionDescriptorCollectionProvider descriptorCollectionProviderWrapper = new(controllerResourceMapping, actionDescriptorCollectionProvider);
+
```
bart-degreed, line 44: Why the added line break?

bart-degreed (changes requested):
Done with review pass.

bart-degreed:
It's annoying that running the tests checks out swagger.json files, which results in failure of cleanupcode later. Can we commit them as-is and exclude them from cleanupcode?

test/OpenApiTests/OpenApiTestContext.cs
```diff
@@ -0,0 +1,58 @@
...
+[UsedImplicitly(ImplicitUseKindFlags.InstantiatedNoFixedConstructorSignature)]
+public sealed class OpenApiTestContext<TStartup, TDbContext> : IntegrationTestContext<TStartup, TDbContext>, IAsyncLifetime
+    where TStartup : class
+    where TDbContext : DbContext
+{
+    internal JsonElement Document { get; private set; }
+
+    public async Task InitializeAsync()
+    {
+        UseController<SupermarketsController>();
```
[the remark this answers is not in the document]
maurei: This indeed doesn't make sense. Setting the controllers should be done in the constructor of a test suite. I would still prefer to store the loaded swagger document in the class fixture rather than using a static property on the test suite.

test/OpenApiTests/OpenApiTestContext.cs
```diff
@@ -0,0 +1,58 @@
...
+    private async Task WriteSwaggerDocumentToFileAsync(string document)
+    {
+        string testSuitePath = GetTestSuitePath();
+        string documentPath = Path.Join(testSuitePath, "swagger.json");
+        await File.WriteAllTextAsync(documentPath, document);
+    }
+
+    private string GetTestSuitePath()
+    {
+        string solutionTestDirectoryPath = Directory.GetParent(Environment.CurrentDirectory)!.Parent!.Parent!.Parent!.FullName;
```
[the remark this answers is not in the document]
maurei: > It's either a file, a directory, or a path. There's no such thing as a directory path.

With `directory path` I mean the path to a directory. [I think there is nothing wrong with that](https://unix.stackexchange.com/a/131585). Will clarify by writing `pathToDirectory`.

test/OpenApiTests/OpenApiTestContext.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        UseController<SupermarketsController>();
+
+        string content = await GetAsync("swagger/v1/swagger.json");
+
+        await WriteSwaggerDocumentToFileAsync(content);
+
+        JsonDocument parsedContent = JsonDocument.Parse(content);
+
+        Document = parsedContent.ToJsonElement();
```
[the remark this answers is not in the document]
maurei: I did not do this because the first line is belongs to the third line as much as it belongs to the second line. The with the separators you've introduced this is less evident.

I also don't care that much because in both cases it is perfectly readable to me, so I will use your suggestion

test/OpenApiTests/OpenApiTests.csproj
```diff
@@ -14,6 +13,7 @@
     <PackageReference Include="coverlet.collector" Version="$(CoverletVersion)" PrivateAssets="All" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="$(AspNetVersion)" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(TestSdkVersion)" />
+    <PackageReference Include="BlushingPenguin.JsonPath" Version="1.0.*" />
```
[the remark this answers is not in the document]
maurei: Good point, thanks

test/OpenApiTests/NamingConvention/PascalCase/PascalCaseTests.cs
```diff
@@ -0,0 +1,437 @@
...
+{
+    private readonly OpenApiTestContext<PascalCaseNamingConventionStartup<NamingConventionDbContext>, NamingConventionDbContext> _testContext;
+
+    public PascalCaseTests(OpenApiTestContext<PascalCaseNamingConventionStartup<NamingConventionDbContext>, NamingConventionDbContext> testContext)
+    {
+        _testContext = testContext;
+    }
+
+    [Fact]
+    public void Pascal_casing_convention_is_applied_to_GetCollection_endpoint()
```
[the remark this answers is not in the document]
maurei: I deliberately avoided using the exact same test names over different suites because it makes it impossible to choose the right one when jumping to a file. That's why I decided to have all the names equal expect for the first two terms. This This way it is still easy to compare the method names while still being able to jump to the right one.

test/TestBuildingBlocks/JsonElementExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+        protected JsonElementAssertions(JsonElement subject)
+        {
+            _subject = subject;
+        }
+
+        public void ContainProperty(string propertyName)
+        {
+            string json = _subject.ToString();
+
+            json.ShouldNotBeNull();
```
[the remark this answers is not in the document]
maurei: Looking at `JsonElement.ToString`, we can be sure it's not going to be null.

src/JsonApiDotNetCore.OpenApi/StringExtensions.cs
maurei: Why is this a problem? The job is being done i.e., the string is being pascalized. If for whatever reason you need to know where this method is defined (if at all the reader was aware of Humanizer having such method, which I do not think we should assume), it is easily figured out.

test/OpenApiClientTests/LegacyClient/ClientAttributeRegistrationLifeTimeTests.cs
maurei: I don't agree. In all CLR types we've chosen it to be singular. Choosing plural here would be weird. Perhaps you're confused with the _value_ of this enum that _should_ be plural: `SupermarketResourceType.Supermarkets`

test/OpenApiTests/LegacyOpenApiIntegration/swagger.json
maurei: This here and https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/1123#discussion_r769609681 go hand in hand. What we pick here results in the CLR type name. As argued there, I think this should be singular, as long as the value is plural.

maurei:
> It's annoying that running the tests checks out swagger.json files, which results in diffs from cleanupcode later. Can we commit them as-is and exclude them from cleanupcode?

I don't know why cleanupcode is still hitting these files, we already have a `swagger.json` file mask in the code analysis. How  do we turn this off?

test/OpenApiTests/NamingConvention/Supermarket.cs
```diff
@@ -0,0 +1,25 @@
+using JetBrains.Annotations;
+using JsonApiDotNetCore.Resources;
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace OpenApiTests.NamingConvention;
+
+[UsedImplicitly(ImplicitUseTargetFlags.Members)]
+[Resource(ControllerNamespace = "OpenApiTests")]
```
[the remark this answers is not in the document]
maurei: It wasn't the default, `OpenApiTests.Controllers` was the namespace value. The problem this solved is that the controllers are placed in the same namespace as where it is referenced from. Putting it in a different namespace and adding a using directive to that namespace, will trigger cleanup code to remove that using directive, which then in turn leads to a failing CI.

src/JsonApiDotNetCore.OpenApi/StringExtensions.cs
bart-degreed: Because it puts people on the wrong track, seeing the package reference to Humanizer and then this call. https://csharpcodingguidelines.com/naming-guidelines/#AV1711. Sure you could define your own class named `Int32` whose `Parse` behaves slightly different, which would be similarly confusing.

test/OpenApiClientTests/LegacyClient/ClientAttributeRegistrationLifeTimeTests.cs
bart-degreed: I'm thinking of an imaginary enum that would list all the resource types in the project:
```
public enum PlaceholderName
{
    Airplanes,
    Supermarkets
}
```
Then the convention would be to pluralize `PlaceholderName` if it had `[Flags]`, which is not the case. So I'm fine with making this singular.

test/OpenApiTests/LegacyOpenApiIntegration/swagger.json
bart-degreed: Agreed.

test/OpenApiTests/OpenApiTestContext.cs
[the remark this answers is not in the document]
bart-degreed: I'm making a distinction to what invokes the logic, vs where that logic is defined.

test/OpenApiTests/OpenApiTestContext.cs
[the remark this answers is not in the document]
bart-degreed: The terms are overloaded. Sometimes 'directory' means one node in the tree, at other times it implies the full path to a location in a subtree. See https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.move?view=net-6.0. I'm fine with the proposed clarification.

test/OpenApiTests/NamingConvention/PascalCase/PascalCaseTests.cs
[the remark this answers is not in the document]
bart-degreed: There are only 3 files, so it's not that hard. And the IDE shows the class/file name as well, IIRC. We're doing this in other places already, such as https://github.com/json-api-dotnet/JsonApiDotNetCore/tree/master/test/JsonApiDotNetCoreTests/IntegrationTests/Links.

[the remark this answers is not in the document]
bart-degreed: Likewise, jump to Can_get_primary_resources and you'll find a bunch of matches in different namespaces. That's what namespaces are for. You wouldn't encode the namespace in regular class names, would you?

src/JsonApiDotNetCore.OpenApi/JsonApiObjects/Documents/NullableResourceIdentifierResponseDocument.cs
```diff
@@ -1,19 +1,25 @@
...
 
 namespace JsonApiDotNetCore.OpenApi.JsonApiObjects.Documents;
 
+// Types in the current namespace are never touched by ASP.NET ModelState validation, therefore using a non-nullable reference type for a property does not
+// imply this property is required. Instead, we use [Required] explicitly, because this is how Swashbuckle is instructed to mark properties as required.
 [UsedImplicitly(ImplicitUseTargetFlags.Members)]
 internal sealed class NullableResourceIdentifierResponseDocument<TResource> : NullableSingleData<ResourceIdentifierObject<TResource>>
     where TResource : IIdentifiable
 {
+    [JsonPropertyName("meta")]
```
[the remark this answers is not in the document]
maurei: It does; in the case of a pascal casing convention, swashbuckle will also apply it to the keys in the data structure. I noticed I missed a few, added them there too.

test/OpenApiTests/OpenApiTests.csproj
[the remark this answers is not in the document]
bart-degreed: I'm still seeing it.

test/TestBuildingBlocks/JsonElementExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+        string referenceSchemaId = jsonElementValue.Split('/').Last();
+        referenceSchemaId.Should().Be(value);
+
+        return new ReferenceSchemaIdAssertion
+        {
+            SchemaReferenceId = value
+        };
+    }
+
+    public sealed class ReferenceSchemaIdAssertion
```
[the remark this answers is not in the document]
bart-degreed: You changed it here, but variables of this type are still named as `documentSchemaRefId`.

JsonApiDotNetCore.sln.DotSettings
```diff
@@ -15,6 +15,10 @@ JsonApiDotNetCore.ArgumentGuard.NotNull($EXPR$, $NAME$);</s:String>
 	<s:Int64 x:Key="/Default/CodeEditing/NullCheckPatterns/PatternTypeNamesToPriority/=JetBrains_002EReSharper_002EFeature_002EServices_002ECSharp_002ENullChecking_002ETraceAssertPattern/@EntryIndexedValue">50</s:Int64>
 	<s:Boolean x:Key="/Default/CodeInspection/CodeAnnotations/PropagateAnnotations/@EntryValue">False</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeInspection/ExcludedFiles/FileMasksToSkip/=swagger_002Ejson/@EntryIndexedValue">True</s:Boolean>
+	
+	
+
```
bart-degreed, line 20: Why the extra line breaks? Did you review changes before commit?

test/OpenApiTests/JsonElementExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+            Execute.Assertion.ForCondition(_subject.TryGetProperty(propertyName, out _))
+                .FailWith($"Expected JSON element '{escapedJson}' to contain a property named '{propertyName}'.");
+        }
+
+        public void BeJson(string expectedDocument)
+        {
+            string json = _subject.ToString();
+
+            json.Should().BeJson(expectedDocument);
+        }
```
bart-degreed, line 82: Why add another version of `BeJson` when we already have one? Reuse the existing one.

test/OpenApiTests/JsonElementExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        return new ReferenceSchemaIdContainer
+        {
+            ReferenceSchemaId = value
+        };
+    }
+
+    public sealed class ReferenceSchemaIdContainer
+    {
+        internal string ReferenceSchemaId { get; init; } = null!;
```
bart-degreed, line 46: Why this clumsiness? You're using the wrong construct so that you need to suppress a compiler warning.

src/JsonApiDotNetCore.OpenApi.Client/ApiException.cs
```diff
@@ -0,0 +1,44 @@
+using JetBrains.Annotations;
+
+// We cannot rely on generating ApiException as soon as we are generating multiple clients, see https://github.com/RicoSuter/NSwag/issues/2839#issuecomment-776647377.
+// Instead, we take the generated code as is and use it for the various clients.
+#nullable disable
+// @formatter:off
+// ReSharper disable All
```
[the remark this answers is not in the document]
bart-degreed: I'm not seeing the expected improvements.

bart-degreed (changes requested):
Will look at other parts later

test/OpenApiTests/NamingConventions/CamelCase/CamelCaseTests.cs
```diff
@@ -0,0 +1,550 @@
+using System.Text.Json;
+using TestBuildingBlocks;
+using Xunit;
+
+namespace OpenApiTests.NamingConventions.CamelCase;
+
+public sealed class PascalCaseTests
```
bart-degreed: PascalCaseTests in CamelCaseTests.cs

test/OpenApiTests/NamingConventions/KebabCase/KebabCaseTests.cs
```diff
@@ -0,0 +1,551 @@
+using System.Text.Json;
+using TestBuildingBlocks;
+using Xunit;
+
+namespace OpenApiTests.NamingConventions.KebabCase;
+
+public sealed class PascalCaseTests
```
bart-degreed: PascalCaseTests in KebabCaseTests.cs

test/OpenApiTests/OpenApiTestContext.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        return swaggerDocument;
+    }
+
+    private static string GetSwaggerDocumentAbsoluteOutputPath(string? relativePath)
+    {
+        AssertHasSwaggerDocumentOutputPath(relativePath);
+
+        string solutionRoot = Path.Combine(Assembly.GetExecutingAssembly().Location, "../../../../../../");
+        string outputPath = Path.Join(solutionRoot, relativePath, "swagger.g.json");
```
bart-degreed, line 46: See my earlier remark on `Path.Combine` vs `Path.Join`.

test/OpenApiTests/OpenApiTests.csproj
```diff
@@ -14,9 +19,10 @@
...
   </ItemGroup>
 
   <ItemGroup>
     <EmbeddedResource Include="LegacyOpenApiIntegration\swagger.json" />
   </ItemGroup>
```
bart-degreed, line 27: Based on the existing convention to place project items above dependencies, this should be moved up.

bart-degreed (changes requested):

test/OpenApiTests/NamingConventions/CamelCase/CamelCaseTests.cs
bart-degreed: If I diff CamelCaseTests.cs with PascalCaseTests.cs in case-insensitive mode, I'm still seeing changes, for example in variable names. Also, the test names are back to Camel_casing_convention... vs Pascal_casing_convention..., which we discussed earlier and decided to avoid such cosmetic differences.

test/OpenApiTests/NamingConventions/KebabCase/KebabCaseTests.cs
bart-degreed: If I diff CamelCaseTests.cs with KebabCaseTests.cs (ignoring all '-' chars), I'm still seeing changes, for example in variable names.

bart-degreed (approved):

PR #1154, opened 2022-05-04
Update to latest Resharper version

bkoelman:


bkoelman:
The failing cibuild is tracked at https://youtrack.jetbrains.com/issue/RSRP-488628/Breaking-InspectCode-fails-with-Roslyn-Worker-process-exited-une.

PR #1169, opened 2022-07-10
Add support for DateOnly/TimeOnly

### LINKED ISSUE #1168: Add support for DateOnly and TimeOnly model properties

bkoelman:
This PR adds support for using [DateOnly](https://docs.microsoft.com/en-us/dotnet/api/system.dateonly) and [TimeOnly](https://docs.microsoft.com/en-us/dotnet/api/system.timeonly) in resource models, which were [introduced](https://devblogs.microsoft.com/dotnet/date-time-and-time-zone-enhancements-in-net-6/#introducing-the-dateonly-and-timeonly-types) in .NET 6.

PostgreSQL fully supports these types. But unfortunately, System.Text.Json support and ModelState binding support for these types are missing in .NET 6. They were added in .NET 7. To use these types in .NET 6:
1. Add a NuGet reference to [System.Text.Json](https://www.nuget.org/packages/System.Text.Json#versions-body-tab) v7 or higher.
2. Add a NuGet reference to [DateOnlyTimeOnly.AspNet](https://www.nuget.org/packages/DateOnlyTimeOnly.AspNet/) and call `builder.Services.AddDateOnlyTimeOnlyStringConverters();` at startup.

While implementing this, I found that query string parsing uses the OS-level culture settings. For example, if you're running your API on an English language version of Windows, but have customized regional settings to display dates in Dutch format, then you'll need to use Dutch notation for date values in query strings (for example `31-12-2022` instead of `12/31/2022`). This doesn't apply to JSON request/response bodies, so the behavior is quite inconsistent. This PR fixes that by always using the [Invariant culture](https://stackoverflow.com/questions/9760237/what-does-cultureinfo-invariantculture-mean). A backward-compatibility switch has been added, so we don't need to wait for the next major release. To revert to the old behavior, add the following at startup:

```c#
AppContext.SetSwitch("JsonApiDotNetCore.ParseQueryStringsUsingCurrentCulture", true);
```

Fixes #1168.

~NOTE: This PR depends on #1249 and #1253, which should be merged first.~

#### QUALITY CHECKLIST
- [x] Adapted tests


bkoelman:
Converted back to Draft because this hasn't been tested with ModelState validation, so may not work.

bkoelman:
Turns out this isn't going to work well with ModelState validation on .NET 6, because `RangeAttribute` internally depends on `System.ComponentModel.TypeConverter`, which doesn't support `DateOnly` and `TimeOnly`. The missing support was [added](https://github.com/dotnet/runtime/issues/59253) in .NET 7. But there does not seem to be a separate NuGet package we can reference for .NET 6 to make this work out of the box.

Therefore I think we should postpone this PR until JsonApiDotNetCore targets .NET 7.

test/JsonApiDotNetCoreTests/IntegrationTests/QueryStrings/Filtering/FilterOperatorTests.cs
```diff
@@ -15,6 +15,26 @@ namespace JsonApiDotNetCoreTests.IntegrationTests.QueryStrings.Filtering;
...
+    private const string IsoDateTimeUpperBound = "2000-11-22T18:47:32";
+
+    private const string InvariantDateTimeLowerBound = "11/22/2000 9:48:17";
+    private const string InvariantDateTimeInTheRange = "11/22/2000 12:34:56";
+    private const string InvariantDateTimeUpperBound = "11/22/2000 18:47:32";
+
+    private const string TimeSpanLowerBound = "2:15:28:54.997";
+    private const string TimeSpanInTheRange = "2:15:51:42.397";
+    private const string TimeSpanUpperBound = "2:16:22:41.736";
+
```
maurei, line 37: I don't see any benefit in adding these. The tests have become harder to read because it now requires more effort to interpret what the actual values are that are being passed on. It is immediately clear that "20" represents a number and that "21"/"19" are higher/lower. The duplication, if any, is not a problem because it serves the purpose of keeping the things close together that are relevant and thus keeps it readable.

For iso/invariant dates there is a slightly better case to be made for the use of constants, but I still think it's unnecessary

maurei (changes requested):

test/JsonApiDotNetCoreTests/IntegrationTests/QueryStrings/Filtering/FilterOperatorTests.cs
bart-vmware: I extracted these constants while experimenting with varying data types, notations, and culture settings. For example, it simplified replacing all `.`s with `,`s, replacing `/` with `-`, or adding a `Z` to all dates. I also found it easier to spot the pattern in:

```c#
    [Theory]
    [InlineData(IsoDateTimeLowerBound, IsoDateTimeUpperBound, ComparisonOperator.LessThan, IsoDateTimeInTheRange)]
    [InlineData(IsoDateTimeLowerBound, IsoDateTimeUpperBound, ComparisonOperator.LessThan, IsoDateTimeUpperBound)]
    [InlineData(IsoDateTimeLowerBound, IsoDateTimeUpperBound, ComparisonOperator.LessOrEqual, IsoDateTimeInTheRange)]
    [InlineData(IsoDateTimeLowerBound, IsoDateTimeUpperBound, ComparisonOperator.LessOrEqual, IsoDateTimeLowerBound)]
    [InlineData(IsoDateTimeUpperBound, IsoDateTimeLowerBound, ComparisonOperator.GreaterThan, IsoDateTimeInTheRange)]
    [InlineData(IsoDateTimeUpperBound, IsoDateTimeLowerBound, ComparisonOperator.GreaterThan, IsoDateTimeLowerBound)]
    [InlineData(IsoDateTimeUpperBound, IsoDateTimeLowerBound, ComparisonOperator.GreaterOrEqual, IsoDateTimeInTheRange)]
    [InlineData(IsoDateTimeUpperBound, IsoDateTimeLowerBound, ComparisonOperator.GreaterOrEqual, IsoDateTimeUpperBound)]
```

as opposed to:
```c#
    [InlineData("2001-01-01", "2001-01-09", ComparisonOperator.LessThan, "2001-01-05")]
    [InlineData("2001-01-01", "2001-01-09", ComparisonOperator.LessThan, "2001-01-09")]
    [InlineData("2001-01-01", "2001-01-09", ComparisonOperator.LessOrEqual, "2001-01-05")]
    [InlineData("2001-01-01", "2001-01-09", ComparisonOperator.LessOrEqual, "2001-01-01")]
    [InlineData("2001-01-09", "2001-01-01", ComparisonOperator.GreaterThan, "2001-01-05")]
    [InlineData("2001-01-09", "2001-01-01", ComparisonOperator.GreaterThan, "2001-01-01")]
    [InlineData("2001-01-09", "2001-01-01", ComparisonOperator.GreaterOrEqual, "2001-01-05")]
    [InlineData("2001-01-09", "2001-01-01", ComparisonOperator.GreaterOrEqual, "2001-01-09")]
```

For example, searching for "LowerBound" in your IDE highlights all of its occurrences, over multiple test sets.

It also makes it easy to duplicate scenarios (and verify during review), for example, by replacing all `Iso` with `Invariant`, then ensuring the new constants are actually correct, in relation to each other.

While it may not substantially improve the readability of the tests in trivial cases, I found it does make them a lot easier to maintain, and it becomes easier to correlate similar patterns.

It is true that it now requires more effort to determine what the actual values are that are being passed on, because of the indirection. As always, the constants themselves reveal intent, which is lost when using values directly. The point of these tests is to verify patterns (yes, by using actual values), but making the patterns obvious in this overwhelming set of combinations is, I believe, more relevant.

For consistency, I decided to apply the pattern for dates to simple numeric types as well here.

bkoelman: I extracted these constants while experimenting with varying data types, notations, and culture settings. For example, it simplified replacing all `.`s with `,`s, replacing `/` with `-`, or adding a `Z` to all dates. I also found it easier to spot the pattern in:

```c#
    [Theory]
    [InlineData(IsoDateTimeLowerBound, IsoDateTimeUpperBound, ComparisonOperator.LessThan, IsoDateTimeInTheRange)]
    [InlineData(IsoDateTimeLowerBound, IsoDateTimeUpperBound, ComparisonOperator.LessThan, IsoDateTimeUpperBound)]
    [InlineData(IsoDateTimeLowerBound, IsoDateTimeUpperBound, ComparisonOperator.LessOrEqual, IsoDateTimeInTheRange)]
    [InlineData(IsoDateTimeLowerBound, IsoDateTimeUpperBound, ComparisonOperator.LessOrEqual, IsoDateTimeLowerBound)]
    [InlineData(IsoDateTimeUpperBound, IsoDateTimeLowerBound, ComparisonOperator.GreaterThan, IsoDateTimeInTheRange)]
    [InlineData(IsoDateTimeUpperBound, IsoDateTimeLowerBound, ComparisonOperator.GreaterThan, IsoDateTimeLowerBound)]
    [InlineData(IsoDateTimeUpperBound, IsoDateTimeLowerBound, ComparisonOperator.GreaterOrEqual, IsoDateTimeInTheRange)]
    [InlineData(IsoDateTimeUpperBound, IsoDateTimeLowerBound, ComparisonOperator.GreaterOrEqual, IsoDateTimeUpperBound)]
```

as opposed to:
```c#
    [InlineData("2001-01-01", "2001-01-09", ComparisonOperator.LessThan, "2001-01-05")]
    [InlineData("2001-01-01", "2001-01-09", ComparisonOperator.LessThan, "2001-01-09")]
    [InlineData("2001-01-01", "2001-01-09", ComparisonOperator.LessOrEqual, "2001-01-05")]
    [InlineData("2001-01-01", "2001-01-09", ComparisonOperator.LessOrEqual, "2001-01-01")]
    [InlineData("2001-01-09", "2001-01-01", ComparisonOperator.GreaterThan, "2001-01-05")]
    [InlineData("2001-01-09", "2001-01-01", ComparisonOperator.GreaterThan, "2001-01-01")]
    [InlineData("2001-01-09", "2001-01-01", ComparisonOperator.GreaterOrEqual, "2001-01-05")]
    [InlineData("2001-01-09", "2001-01-01", ComparisonOperator.GreaterOrEqual, "2001-01-09")]
```

For example, searching for "LowerBound" in your IDE highlights all of its occurrences, over multiple test sets.

It also makes it easy to duplicate scenarios (and verify during review), for example, by replacing all `Iso` with `Invariant`, then ensuring the new constants are actually correct, in relation to each other.

While it may not substantially improve the readability of the tests in trivial cases, I found it does make them a lot easier to maintain, and it becomes easier to correlate similar patterns.

It is true that it now requires more effort to determine what the actual values are that are being passed on, because of the indirection. As always, the constants themselves reveal intent, which is lost when using values directly. The point of these tests is to verify patterns (yes, by using actual values), but making the patterns obvious in this overwhelming set of combinations is, I believe, more relevant.

For consistency, I decided to apply the pattern for dates to simple numeric types as well here.

maurei: > For consistency, I decided to apply the pattern for dates to simple numeric types as well here.

I'd say in a test suite consistency like this is not a virtue if it doesn't serve readability.

>As always, the constants themselves reveal intent, which is lost when using values directly.

I disagree, I think intent for the simple cases intent is obfuscated instead of revealed. Because in the following snippet
```c#
[InlineData(IntUpperBound, IntLowerBound, ComparisonOperator.GreaterOrEqual, IntUpperBound)]
public async Task Can_filter_comparison_on_whole_number(string matchingValue, string nonMatchingValue, ComparisonOperator filterOperator,
        string filterValue)
```
to me the intent is obfuscated more than compared to in
```c#
[InlineData(21, 19, ComparisonOperator.GreaterOrEqual, 21)]
public async Task Can_filter_comparison_on_whole_number(string matchingValue, string nonMatchingValue, ComparisonOperator filterOperator,
        string filterValue)
```
because of the cognitive effort required to see the relationship between the numbers in the former, whereas in the latter the relationship, and therefore the intent too, is immediately clear.

> It also makes it easy to [...] verify during review

As a result of that relative obfuscation of intent, it would actually take me _more_ time to interpret every single test case, specifically the ones with trivial values which most of them are. It didn't matter much now because I mostly needed to review if the replacements were done correctly for these cases. But it would matter if one were to read freshly.

At the same time however, investing in this readability topic more discussion time than the extra time it would ever take a new reader to interpret the tests also makes little sense, so I'll approve and leave it up to you to decide what to do with it 👍🏼

bkoelman: Thanks for the feedback. I appreciate it, and I do get your point. So this is about whether or not to use constants for the `int` test set. I'd still like them to stay, due to the following:

image

Clicking on `DoubleUpperBound` in the IDE immediately highlights all usages, which I can't do with inlined values. Furthermore, the constants make it easy to spot we're using the same set of values (19, 20, and 21) for all combinations.

At some point, I had a duplicated set of tests for the back-compat switch, but removed them because it requires setting global thread state that potentially influences the execution of other unrelated tests. Though I wouldn't be surprised if we need to play with these in the future again, as culture settings often pop up at unexpected times.

So I'd like to leave it like this, for now at least. Let's see how it evolves in the future. Inlining the constants is just a few keystrokes, which we can do at any time.

PR #1192, opened 2022-09-28
Multi-target Annotations against .NET Standard 1.0

### LINKED ISSUE #1190: JsonApiDotNetCore.Annotation for netstandard2.0
Before coming to my request, let me first say how awesome the new stuff is that you have developed since we last talked. I love the source generators.

Now to my feature request . The context is that I want to create a separate project that:
- contains the model classes hat will have to be annotated with the `ResourceAttribute`, among others; and
- targets `netstandard2.0` because it will also be required by one or more projects that are, unfortunately, bound to .NET Framework.

I thought that I could add the `JsonApiDotNetCore.Annotations` package as a dependency to that model-related project. However, that NuGet package depends on `net60`, which I can't use for my model project for the reasons described above.

Doe the Annotations package have to depend on `net60` or could this also target `netstandard2.0`? In other words, is there a way you could provide a NuGet package that is useable from .NET Framework or `netstandard2.0`?

bkoelman:
Adds netstandard1.0 as a TFM to the `JsonApiDotNetCore.Annotations` NuGet package. This enables consumers to share their models project with .NET Framework code.

The types in the NetStandard target don't actually work with JsonApiDotNetCore. This is provided solely for convenience so that consumers don't need to use conditional defines that are based on the target framework.

Internally, we use project-level conditionals to include/exclude source files per target framework, based on their file name:
- `*.shared.cs`: Included for both targets
- `*.netstandard.cs`: Only included for netstandard1.0 (added as `None` for .NET 6 so it shows up in the IDE)
- `*.cs`: Only included for .NET 6

This approach avoids the use of `#if ... #else ... #endif` in our source files, which is problematic for Resharper CleanupCode.

Closes #1190.

#### QUALITY CHECKLIST
- [x] Adapted tests


ThomasBarnekow (approved):
This looks good to me. It covers all the attributes that I have used in my code (`Resource`, `NoResource`, `Attr`, `HasMany`, and `HasOne`).

bkoelman:
@ThomasBarnekow Thanks!

maurei (approved):

PR #1218, opened 2022-11-12
Improve speed of running tests

bkoelman:
This PR activates async throttling of integration tests (instead of reducing parallelism in xUnit, which is what we had before). It uses the approach described at https://github.com/xunit/xunit/issues/2003#issuecomment-943123699.

Throttling is needed to avoid PostgreSQL failing because the maximum number of open connections has been reached (100 by default). As an alternative for throttling, I've tried increasing the max connections limit to 1000 and 500, but without much success.

I've experimented with various throttling thresholds and found a good balance at 64, which leaves some room for abandoned connections (when hard-killing tests during development and debugging). Watching the database server with pgAdmin while running tests, I didn't see more than ~50 temporary databases simultaneously. Measurements confirm that setting the threshold higher doesn't provide measurable gain.

In addition, all models in integration tests now use `DeleteBehavior.SetNull` instead of `DeleteBehavior.ClientSetNull`, so that PostgreSQL performs the related deletes on the database server. That's not only more reliable but also more efficient; it shaves off another ~5 seconds.

Running all tests on my computer with:

```bash
git clean -xdf ; dotnet build
Measure-Command { dotnet test --no-build | Out-Default }
```

This yields the following results:
| system | master (sec) | PR (sec) | % |
|---|---|---|---|
| local | 96 | 45 | 46% |
| AppVeyor Windows | 209 | 221 | 106% |
| AppVeyor Linux | 239 | 195 | 82% |

Conclusion: this makes running our tests twice as fast locally! The duration of cibuilds varies wildly, so measurements are not very reliable to compare. Also, their VMs have fewer cores, possibly resulting in more context switches. 

#### QUALITY CHECKLIST
- [x] Adapted tests

### DIFF
--- a/test/AnnotationTests/AnnotationTests.csproj
+++ b/test/AnnotationTests/AnnotationTests.csproj
@@ -4,12 +4,6 @@
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
-  <ItemGroup>
-    <None Update="xunit.runner.json">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </None>
-  </ItemGroup>
-
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.0' ">
     <Using Remove="System.Net.Http" />
   </ItemGroup>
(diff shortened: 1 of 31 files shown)


bkoelman:
@maurei Can you check on MacOS whether tests still succeed and if/how it improves testrun duration?

maurei (approved):

Issue #1111, opened 2021-11-16
Add support for handling of non-nullable reference types in `required`.


maurei:
For non-nullable reference types, a resource attribute should be included in the OpenAPI `required` property if model state validation is enabled, and should be excluded otherwise.

The following table displays a comprehensive overview of the behavior of `required`, `nullable` in an OpenAPI schema, depending on the property type, usage of `[Required]`, configuration of NRT, and configuration of model state validation.

<img width="1643" alt="177394179-2943249b-f5f1-4a72-adab-1c1d7ae85fc6" src="https://github.com/json-api-dotnet/JsonApiDotNetCore/assets/10324372/16dfaf94-8797-48bb-b724-5e5384d2be69">

maurei:
An interesting case worth elaborating on is the OAS nullability on **B6** and **H6**. One might expect `TRUE` here, after all within the scope an OAS description it is possible for the field here to be `required` (it cannot be omitted) while simultaneously allowing for `null`.

The caveat here, however, is that our implementation will not allow `null` because the model state validation will not allow it. If MSV is disabled, it will would still fail because of a `null constraint` violation in the database.

bkoelman:
I'm not entirely sure about H6. The API may be running against a pre-existing database without column constraints in place, or not use EF Core at all. Is there a use case where someone intentionally declares a property as `[Required] string?`, or is this just a theoretical case?

Also, note that `[Required]` means more than just not-null. Excerpt from https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.requiredattribute?view=net-6.0:
> The [RequiredAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.requiredattribute?view=net-6.0) attribute specifies that when a field on a form is validated, the field must contain a value. A validation exception is raised if the property is null, contains an empty string (""), or contains only white-space characters.

maurei:
Another (theoretical) use case for having a `[Required] string?` property could be to read `null` as a "pristine" state, but disallowing a client to write `null`. Eg. `[Required] int? Grade` with value `null` could mean "no grade (yet)".

I don't know if this is actually configurable with EF Core, [because `RequiredAttribute` means not null, as far as it is concerns mapping to a database column using EF Core.](https://www.learnentityframeworkcore.com/configuration/data-annotation-attributes/required-attribute).

[...]

PR #1231, opened 2022-12-19
Improved JsonApiClient and relationships into `openapi-required-and-nullable`

maurei:
This PR brings merges into #1185:

- For resource attributes: JsonApiClient improvements when dealing with nullable/required/default-value properties when NRT is on/off. Also contains extensive test coverage.
- For resource relationships: adjustment in `JsonApiDotNetCore.OpenApi` and extensive test coverage

For server-side testing (i.e. the generated OAS file), the table in #1111 provides a transparent overview for resource attributes of what is expected for `required`/`nullable` when NRT and/or MSV are on/off. This piece is already covered by #1185. 

However, for client-side testing, that table doesn't clearly show which `JsonApiClient` behaviour during request is expected to fail and succeed. For that I've extended the overview in this PR to include relevant operations for which the client-side behaviour should be captured, see below. The overview is also extended to contain the same for relationships.

Screenshot 2022-12-24 at 20 00 33

For instance, from this overview it reads:

A client _should_ be able to generate a request intentionally in which the json property 
- `data.attributes.name` is set to `null` (G9);
- `data.attributes.name` is omitted from the request body (G9);
- `data.attributes.weight` is set to its default value `0` (G12)

and _should not_ be able to generate a request intentionally where
- `data.attributes.weight` is omitted from the request body (H12)

where these paths correspond to the `Name`/`Weight` properties on the `Chicken` resource as declared server-side, for the scenario: NRT disabled and MSV enabled

Few more details:
- Note that "NRT disabled" means: disabled both server-side and client-side.  Therefore, the test cases in the PR are:  1 (server-on, client-on) and 2 (server-off, client-off).  That implies we could for completeness also look into (server-on, client-off) and (server-off, client-on), but these scenarios don't generate any unique test cases. So I decided to leave it out for conciseness.
- the scenario (NRT off, MSV on) is included but not (NRT off, MSV off), because it is identical. To see this, [compare columns FGH with KLM](https://user-images.githubusercontent.com/9339132/177394179-2943249b-f5f1-4a72-adab-1c1d7ae85fc6.png)
- the scenario (NRT on, MSV on) is included but not (NRT on, MSV off), because it is nearly identical. [Only I3 differs from C3](https://user-images.githubusercontent.com/9339132/177394179-2943249b-f5f1-4a72-adab-1c1d7ae85fc6.png), but that case doesn't deserve a new test suite because there are plenty of other cases already where omitting is not allowed in a similar circumstance.


maurei:
@bkoelman your two cents on the following please.

Compare the current [RequestTests#Can_exclude_relationships_that_are_required_for_POST_when_performing_PATCH](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/affc1876074fc9d6d5503bcf11ff4c63e0fa0614/test/OpenApiClientTests/SchemaProperties/NullableReferenceTypesDisabled/RequestTests.cs#L551) with the newly proposed [AlternativeFormRequestTests#Can_exclude_relationship_that_is_required_for_POST_when_performing_PATCH](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/affc1876074fc9d6d5503bcf11ff4c63e0fa0614/test/OpenApiClientTests/SchemaProperties/NullableReferenceTypesDisabled/AlternativeFormRequestTests.cs#L65).

In the first form (current form), it isn't clear by reading the test how many or which relationships are being omitted. Also it isn't clear that every omitted relationship represents a distinct and unrelated test case. So I realised that this test too should actually be a `Theory`. In the second format it is now immediately clear that `OldestChicken` and `AllChickens` are the relevant omitted properties and that they represent individual test cases.

It comes at the cost that the [json assertion string](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/affc1876074fc9d6d5503bcf11ff4c63e0fa0614/test/OpenApiClientTests/SchemaProperties/NullableReferenceTypesDisabled/AlternativeFormRequestTests.cs#L139) is not as readable anymore. But given that before it didn't illuminate what was relevant either, even when it was readable, I argue that this isn't a problem. And when the test fails, it will still be readable in the log anyway.

<details>
<summary>Same would apply to the following remaining tests:</summary>

`Can_exclude_optional_attributes`
`Can_exclude_attributes_that_are_required_for_POST_when_performing_PATCH`
`Can_clear_nullable_attributes`
`Can_set_default_value_to_ValueType_attributes`
`Can_exclude_optional_relationships`

These tests only have one omitted property so technically don't need a theory, but could stil consider doing so to highlight which property is omitted:
`Cannot_clear_required_attribute_when_performing_POST`
`Can_clear_nullable_relationship`
</details>

PR #1239, opened 2023-01-04
Answer for question at #1238

### LINKED ISSUE #1238: Missing relationship data when using non Entity Framework apprach
#### SUMMARY
I'm trying to setup simple Non Entity Framework API using GetAsync method for collections from JsonApiResourceService<TResource, TId> class, where the resource has "HasMany" relationship with another resource, but the response does not contain details about that relationship.

#### DETAILS
```
// This is the main resource for which I'm creating the API:
[Resource]
public class Booking: Identifiable<string>
{
    [Attr(PublicName = "title")]
    public required string Title { get; set; }
    [HasMany(PublicName = "spaces")]
    public required List<Space> Spaces { get; set; } // space ids
}

// This is another resource which should be connected to Bookings.
[Resource]
public class Space: Identifiable<string>
{
    [Attr(PublicName = "title")]
    public required string Title { get; set; }    
}

// This is the implementation of BookingService.
public class BookingsService : JsonApiResourceService<Booking, string>
{
    public BookingsService(
        IResourceRepositoryAccessor repositoryAccessor,
        IQueryLayerComposer queryLayerComposer,
        IPaginationContext paginationContext,
        IJsonApiOptions options,
        ILoggerFactory loggerFactory,
        IJsonApiRequest request,
        IResourceChangeTracker<Booking> resourceChangeTracker,
        IResourceDefinitionAccessor resourceDefinitionAccessor
    )
        : base(
            repositoryAccessor,
            queryLayerComposer,
            paginationContext,
            options,
            loggerFactory,
            request,
            resourceChangeTracker,
            resourceDefinitionAccessor
        ) { }

    public override async Task<IReadOnlyCollection<Booking>> GetAsync(CancellationToken cancellationToken)
    {
        return new List<Booking>
        {
            new()
            {
                Id = "1",
                Title = "booking 1",
                Spaces = new List<Space>
                {
                    new() { Id = "1", Title = "test space" }
                }
            },
            new()
            {
                Id = "2",
                Title = "booking 2",
                Spaces = new List<Space>
                {
                    new() { Id = "1", Title = "test space" }
                }
            },
        };
    }
```

#### STEPS TO REPRODUCE
Using the code above I managed to create the Get endpoint for Booking resource with the route like https://mycompanyname.com/api/bookings which produces the following response:
```
{
    "data": [
        {
            "type": "bookings",
            "id": "1",
            "attributes": {
                "title": "booking 1"
            },
            "relationships": {
                "spaces": {
                    "links": {
                        "self": "https://mycompanyname.com/api/bookings/1/relationships/spaces",
                        "related": "https://mycompanyname.com/api/bookings/1/spaces"
                    }
                }
            }
        },
        {
            "type": "bookings",
            "id": "2",
            "attributes": {
                "title": "booking 2"
            },
            "relationships": {
                "spaces": {
                    "links": {
                        "self": "https://mycompanyname.com/api/bookings/2/relationships/spaces",
                        "related": "https://mycompanyname.com/api/bookings/2/spaces"
                    }
                }
            }
        }
    ]
}
```
But unfortunately I don't see information about the spaces in the response  (only relationship links but no ids). I also tried usng "include=spaces" query param, but without any success. Could you please help my here, what should I do to add the ids of spaces to the response. Thanks.

#### VERSIONS USED
- JsonApiDotNetCore version: 5.1.1
- ASP.NET Core version: 7.0
- Entity Framework Core version: -
- Database provider: -

bkoelman:
The first commit in this PR adds the repro scenario from #1238 to the NoEntityFrameworkExample project. The second commit fixes the problem. Set that project as the startup project and press F5 to run.

Without the fix, adding a breakpoint at:

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/be992341d236094e0abbf9d0a8e9bafb788356c8/src/JsonApiDotNetCore/Serialization/Response/ResponseModelAdapter.cs#L67

shows that `_evaluatedIncludeCache` is empty. It does not contain the requested include from query string. Searching for usages of its `Set` method, we can see it is called by `QueryLayerComposer.ComposeFromConstraints`:

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/be992341d236094e0abbf9d0a8e9bafb788356c8/src/JsonApiDotNetCore/Queries/Internal/QueryLayerComposer.cs#L141

which in turn is called by the built-in resource service `GetAsync` method at:

https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/be992341d236094e0abbf9d0a8e9bafb788356c8/src/JsonApiDotNetCore/Services/JsonApiResourceService.cs#L76

So the fix is to call this method before returning the data.

Now, why is this needed? When using the built-in resource service and repository, callbacks are made into `IResourceDefinition<,>` methods, which is our extensibility point for API developers to write custom business logic. In this case, `OnApplyIncludes` is called, which enables developers to add/remove includes from the request. The evaluated set of includes is then stored in `IEvaluatedIncludeCache`. Likewise, there's an `ISparseFieldSetCache` that behaves similarly.

`QueryLayerComposer.ComposeFromConstraints` gathers all the parsed query string parameters, combines that with configured options (such as default page size), and turns that into a `QueryLayer` object, which is the internal representation of the current request. While building that, it visits the `IResourceDefinition<,>` methods. Call `QueryLayer.ToString()` in the debugger to see what it contains. It may be of use in your further custom resource service implementation. `QueryLayer` is passed to `IResourceRepository<>`, which is database-agnostic. Instead of creating a custom resource service, it may be easier to create a custom repository instead. That way, you're reusing a lot of the built-in features at the resource service level.

Closes #1238.

### DIFF
--- a/src/Examples/NoEntityFrameworkExample/Models/Booking.cs
+++ b/src/Examples/NoEntityFrameworkExample/Models/Booking.cs
@@ -1,0 +1,15 @@
+using JsonApiDotNetCore.Resources;
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace NoEntityFrameworkExample.Models;
+
+// This is the main resource for which I'm creating the API:
+[Resource]
+public class Booking : Identifiable<string>
+{
+    [Attr(PublicName = "title")]
+    public required string Title { get; set; }
+
+    [HasMany(PublicName = "spaces")]
+    public required List<Space> Spaces { get; set; } // space ids
+}
--- a/src/Examples/NoEntityFrameworkExample/Models/Space.cs
+++ b/src/Examples/NoEntityFrameworkExample/Models/Space.cs
@@ -1,0 +1,12 @@
+using JsonApiDotNetCore.Resources;
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace NoEntityFrameworkExample.Models;
+
+// This is another resource which should be connected to Bookings.
+[Resource]
+public class Space : Identifiable<string>
+{
+    [Attr(PublicName = "title")]
+    public required string Title { get; set; }
+}
--- a/src/Examples/NoEntityFrameworkExample/NoEntityFrameworkExample.csproj
+++ b/src/Examples/NoEntityFrameworkExample/NoEntityFrameworkExample.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
-    <TargetFramework>$(TargetFrameworkName)</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Examples/NoEntityFrameworkExample/Program.cs
+++ b/src/Examples/NoEntityFrameworkExample/Program.cs
@@ -10,9 +10,19 @@ WebApplicationBuilder builder = WebApplication.CreateBuilder(args);
 string connectionString = GetConnectionString(builder.Configuration);
 builder.Services.AddNpgsql<AppDbContext>(connectionString);
 
-builder.Services.AddJsonApi(options => options.Namespace = "api/v1", resources: resourceGraphBuilder => resourceGraphBuilder.Add<WorkItem, int>());
+builder.Services.AddJsonApi(options =>
+{
+    options.Namespace = "api/v1";
+    options.SerializerOptions.WriteIndented = true;
+}, resources: resourceGraphBuilder =>
+{
+    resourceGraphBuilder.Add<WorkItem, int>();
+    resourceGraphBuilder.Add<Booking, string>();
+    resourceGraphBuilder.Add<Space, string>();
+});
 
 builder.Services.AddResourceService<WorkItemService>();
+builder.Services.AddResourceService<BookingsService>();
 
 WebApplication app = builder.Build();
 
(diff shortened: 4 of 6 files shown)


SJShip:
@bkoelman , Thanks for your explaination! 

So, as I understand, when using `_queryLayerComposer.ComposeFromConstraints(_request.PrimaryResourceType!)` we can build the `QueryLayer` object, but for resource service scenario it allows us to support including relationships only (since it sets `_evaluatedIncludeCache` which, in turn, is used in `ResponseModelAdapter`). It will not support pagination, filtering and fields selection. In order to make the rest of the feature workable we need to somehow apply the `QueryLayer` object to the retireved collection of resources. 

I spotted one of the solution from the `EntityFrameworkCoreRepository`, we could try implementing another resource repository and use `QueryableBuilder` class. But the issue is that the class expects `IModel entityModel` in its constructor and we do not have EntityFramework data context in our solution. Instead, we have a `List<Booking>` resource collection. It's already materialized.

Could you, please, guide me is there any alternative solution to apply query layer without presence of EF data context? Or maybe I missed something?

bkoelman:
Normally all operations are translated to SQL; `QueryLayer` is just a representation of what the user asked for. If not using EF Core, you'll need to find your own solution to fetch what was asked for, which means side-loading related resources, filtering, sorting, pagination, and field selection. The serializer just hides unrequested fields as a last resort.

If your data store supports LINQ queries, you could try to reuse parts of `QueryableBuilder`. We do that for MongoDB, see [here](https://github.com/json-api-dotnet/JsonApiDotNetCore.MongoDb/blob/1fddd2eade024ef04427ac28e6eb9fcdf0c5bfe6/src/JsonApiDotNetCore.MongoDb/Repositories/MongoRepository.cs#L118-L119) on how to overcome the `IModel` requirement. You'll also need to block processing of calls to `.Include()` in the query, as that an EF Core-specific extension method. You can override `QueryableBuilder` methods to do so.

Issue #1051, opened 2021-09-01
Support for various data types of resource attributes and id


maurei:
Resources with attributes of various data types should be thoroughly tested. The following data types should be considered:

- DateTime, DateTimeOffset, TimeSpan, DateOnly, TimeOnly
- decimal, double, float
- signed and unsigned: int, long, short, byte 
- guid
- bool
- string, char
- enum
- tuples
- collections
- [complex attributes](https://www.jsonapi.net/usage/resources/attributes.html#complex-attributes)

Also take into consideration: 
- nullability of above types where applicable
- all EF Core supported datatypes for identifying keys should be considered
- see https://swagger.io/docs/specification/data-models/data-types/

~Also address the following: When using `Identifiable<TId>` where `TId` is other than `string`, eg `int`, the generated client will have its `id` parameter typed as `int` in methods, which is incorrect: it should always be `string`.~

bkoelman:
Should include support for DateOnly and TimeOnly, which were added in .NET 6.

PR #1240, opened 2023-01-06
Rely on supported `AllOf` approach for creating nullable reference schemas

maurei:
Uses the Swashbuckle built-in support for `AllOf` to create nullable reference schema types, rather than the approach I had before using `oneOf`, see https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/861.

Apart from yielding a cleaner implementation, it will also be needed to properly support complex attributes as per #1051.

Todo: set base branch to `open-api` once #1185 is merged.


bkoelman:
Superseded by #1372.

PR #1245, opened 2023-01-17
Documentation: FAQ and Common Pitfalls

### LINKED ISSUE #1248: Docs markdown: use docfx alerts
Review the existing documentation at `/docs/*.md` for usage of tips, notes, warnings, etc. And replace them with [doxfc alerts](https://dotnet.github.io/docfx/docs/markdown.html#alerts). They look a lot nicer!

bkoelman:
Aside from minor infrastructure improvements, this PR adds FAQ and Common Pitfalls to the documentation website.

Closes #1248.

#### QUALITY CHECKLIST
- [ ] N/A: Changes implemented in code
- [ ] N/A: Complies with our [contributing guidelines](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/.github/CONTRIBUTING.md)
- [ ] N/A: Adapted tests
- [x] Documentation updated


docs/getting-started/faq.md
```diff
@@ -0,0 +1,169 @@
...
+
+#### How do I expose non-JSON:API endpoints?
+You can add your own controllers that do not derive from `(Base)JsonApiController` or `(Base)JsonApiOperationsController`.
+Whatever you do in those is completely ignored by JsonApiDotNetCore.
+This is useful if you want to add a few RPC-style endpoints or provide binary file uploads/downloads.
+
+A middle-ground approach is to add custom action methods to existing JSON:API controllers.
+While you can route them as you like, they must return JSON:API resources.
+And on error, a JSON:API error response is produced.
+This is useful if you want to stay in the JSON:API-compliant world, but need to expose something on-standard, for example: `GET /users/me`.
```
maurei, line 109: Typo fix: 

```suggestion
This is useful if you want to stay in the JSON:API-compliant world, but need to expose something non-standard, for example: `GET /users/me`.
```

docs/getting-started/faq.md
```diff
@@ -0,0 +1,168 @@
...
+We use this for accessing [MongoDB](https://github.com/json-api-dotnet/JsonApiDotNetCore.MongoDb/blob/674889e037334e3f376550178ce12d0842d7560c/src/JsonApiDotNetCore.MongoDb/Queries/Internal/QueryableBuilding/MongoQueryableBuilder.cs).
+
+> [!TIP]
+> [ExpressionTreeVisualizer](https://github.com/zspitz/ExpressionTreeVisualizer) is very helpful in trying to debug LINQ expression trees!
+
+#### I love JsonApiDotNetCore! How can I support the team?
+The best way to express your gratitude is by starring our repository.
+This increases our leverage when asking for bug fixes in dependent projects, such as the .NET runtime and Entity Framework Core.
+Of course, a simple thank-you message in our [Gitter channel](https://gitter.im/json-api-dotnet-core/Lobby) is appreciated too!
+We don't take monetary contributions at the moment.
```
maurei, line 162: I'm currently paying for www.jsonapi.net. So I'd propose to set this up, although it isn't much. If we ever get to the point of having an excess, we could consider getting a better appveyor plan for faster  builds.

docs/usage/common-pitfalls.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### One-to-one relationships require custom Entity Framework Core mappings
+Entity Framework Core has great conventions and sane mapping defaults. But two of them are problematic for JSON:API: identifying foreign keys and default delete behavior.
+See [here](~/usage/resources/relationships.md#one-to-one-relationships-in-entity-framework-core) for how to get it right.
+
+#### Prefer model attributes over fluent mappings
+Validation attributes such as `[Required]` are detected by ASP.NET ModelState validation, Entity Framework Core, OpenAPI, and JsonApiDotNetCore.
+When using a Fluent API instead, the other frameworks cannot know about it, resulting in a less streamlined experience.
+
+#### Validation of `[Required]` value types doesn't work
```
maurei, line 115: We've discussed this one extensively during my work on the OpenAPI integration. I remember looking for why it behaves this way, but I don't recall finding anything compelling. Do you know why this is by design?

maurei (changes requested):
This summarises pretty neatly all of the things we've run into over the past few years. Well done!.

docs/getting-started/faq.md
bkoelman: I'm fine with that.

docs/usage/common-pitfalls.md
bkoelman: Because the validation runs on a .NET object instance. A non-nullable value type always has a value. See https://stackoverflow.com/questions/50910093/asp-net-core-require-non-nullable-types.

bkoelman: See #1254 for some thoughts on this.

maurei: Yes this is the technical perspective of which I'm aware why it behaves as such. But what I wonder is why it was designed to behave as such, because from a functional perspective it makes little sense to me that `[Required]` behaves different for omitted value types versus omitted non-nullable reference types. That it uses an instantiated .NET object under the hood to me seems an implementation detail that shouldn't be a decisive factor for how it behaves.

As to #1254, I agree with your remark related to that it is beyond the scope/responsibility of our project to override ASP.NETs behaviour here

docs/getting-started/faq.md
maurei: Will do! Let's leave in your comment about not taking monetary contributions until I've set it up.

maurei (approved):

PR #1254, opened 2023-02-05
Fail on missing required non-nullable value type in request body

bkoelman:
This proof-of-concept PR fails deserialization of an incoming request body at a POST resource endpoint when the value is omitted for a non-nullable value type property decorated with `[Attr] [Required]`.

This works around the limitation in ASP.NET ModelState validation described at https://www.jsonapi.net/usage/resources/nullability.html#value-types.

However, I wonder if we should actually take this approach. It feels like a dirty hack that mixes model binding with validation, which are distinct steps in ASP.NET. Also, it prevents any other ModelState violations from being returned in the response. A better approach would be to still detect it during deserialization, but instead of producing an error, somehow make it result in a ModelState validation error via request-scoped shared state. This would only work properly when ModelState validation is turned on. That approach would potentially apply to various other errors that currently cause a deserialization failure, such as "missing ID", "property is read-only" etc.

And this is just one approach to dealing with the situation. Alternatives:
- Document the behavior and advise users not to do this. This is a fundamental limitation in ASP.NET, not our concern.
- Scan for this situation when building the resource graph and produce a hard error that prevents application startup, or log a warning. The former is a breaking change, the latter means it'll still occur and we may want to handle it.

#### QUALITY CHECKLIST
- [ ] TODO: Adapted tests
- [ ] Documentation updated

### DIFF
--- a/src/JsonApiDotNetCore/Serialization/JsonConverters/ResourceObjectConverter.cs
+++ b/src/JsonApiDotNetCore/Serialization/JsonConverters/ResourceObjectConverter.cs
@@ -1,9 +1,11 @@
+using System.ComponentModel.DataAnnotations;
 using System.Reflection;
 using System.Text.Json;
 using JetBrains.Annotations;
 using JsonApiDotNetCore.Configuration;
 using JsonApiDotNetCore.Resources;
 using JsonApiDotNetCore.Resources.Annotations;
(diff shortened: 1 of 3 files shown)


bkoelman:
Closing due to https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/1245#discussion_r1096798270.

PR #1281, opened 2023-06-14
task: bump to net7, EFCore6

wayne-o:
Since JADNC 4.2.0 works fine under NET7 and EFCore 6 i am proposing we create a Nuget release/4.2.1 

The reason this is valuable is because it allows developers to more easily migrate to JADNC 5.2.0 by adding a step for them to more safely upgrade itteratively and test their code on NET7 / EFCore 6 before taking the leap to 5.2.0

Closes #{ISSUE_NUMBER}

#### QUALITY CHECKLIST
- [ ] Complies with our [contributing guidelines](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/.github/CONTRIBUTING.md)
- [ ] Adapted tests
- [ ] Documentation updated

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,9 +1,9 @@
 <Project>
   <PropertyGroup>
-    <NetCoreAppVersion>netcoreapp3.1</NetCoreAppVersion>
-    <AspNetCoreVersion>3.1.*</AspNetCoreVersion>
-    <EFCoreVersion>3.1.*</EFCoreVersion>
-    <NpgsqlPostgreSQLVersion>3.1.*</NpgsqlPostgreSQLVersion>
+    <NetCoreAppVersion>net7.0</NetCoreAppVersion>
+    <AspNetCoreVersion>7.0.0</AspNetCoreVersion>
+    <EFCoreVersion>6.0.0</EFCoreVersion>
+    <NpgsqlPostgreSQLVersion>6.0.0</NpgsqlPostgreSQLVersion>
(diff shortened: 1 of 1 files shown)


bkoelman:
Running the tests on this PR fails most of them, because the JsonApiDotNetCore project still targets EF Core 3.1, resulting in lots of `MissingMethodException`s. Changing the project file to:

```c#
    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="$(EFCoreVersion)" />
    <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" Version="$(EFCoreVersion)" />
```

fixes that, but still results in 367 failed tests. Many of them are caused by breaking changes in the PostgreSQL EF Core provider around date/time handling. I'm not at all convinced that JADNC v4.x "just works" on .NET 7 with EF Core 6.

In theory, we could multi-target against:
- .NET Core 3.1 with EF Core 3.1
- .NET 6 with EF Core 6 (questionable)
- .NET 7 with EF Core 7 (questionable)

This is still not what you need. EF Core 6 and 7 both run on .NET 6 and 7. We can only multi-target based on the target framework, so we can't offer the combination of EF Core 6 on .NET 7, nor EF Core 7 on .NET 6, while preserving the listed entries above.

However, all of this is irrelevant because changing the dependencies to another major version is a massive breaking change. We [cannot do that](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/VERSIONING_POLICY.md) in a minor version increment. It would break all existing users of v4.x once they update.

PR #1284, opened 2023-06-15
Pluggable filter value conversions

### LINKED ISSUE #1275: JsonConverterAttribute for Query Params
**Is your feature request related to a problem? Please describe.**
Currently, in the serialization and deserialization of request bodies, the `JsonConverterAttribute` is considered, allowing for custom conversion logic to be applied. However, when it comes to filters, this attribute is ignored. While this behavior is expected since query parameters are not in JSON format, there are cases where it would be useful to apply certain conversions.

Consider the following scenario where a custom-type `Wrapper` and its corresponding `WrapperJsonConverter` are used:

```csharp
using System;
using System.Text.Json;
using System.Text.Json.Serialization;

[JsonConverter(typeof(WrapperJsonConverter))]
class Wrapper
{
    public string Part1 { get; }
    public string Part2 { get; }

    public Wrapper(string parts)
    {
        string[] split = parts.Split(",");
        Part1 = split[0];
        Part2 = split[1];
    }

    public Wrapper(string part1, string part2)
    {
        Part1 = part1;
        Part2 = part2;
    }
    
    public override string ToString() => Part1 + "," + Part2;
}

public class WrapperJsonConverter : JsonConverter<Wrapper>
{
    public override Wrapper Read(
        ref Utf8JsonReader reader,
        Type typeToConvert,
        JsonSerializerOptions options
    )
    {
        return new Wrapper(reader.GetString());
    }

    public override void Write(
        Utf8JsonWriter writer,
        Wrapper value,
        JsonSerializerOptions options
    )
    {
        writer.WriteStringValue(value.ToString());
    }
}
```

Currently, when `new Wrapper("part1", "part2")` is serialized and included in a request body, it is correctly converted to `"part1,part2"`. However, if the client attempts to use a query parameter like `equals(wrapper, 'part1,part2')`, it will encounter a `QueryParseException` with the message "Failed to convert 'part1,part2' of type 'String' to type 'Wrapper'." This can lead to confusion and frustration, as the only way forward, is the take the fields from Wrapper and adding them wherever needed.

**Describe the solution you'd like**
To address this limitation, I propose enhancing `RuntimeTypeConverter` to support custom type conversions for query parameters as well. This feature would enable developers to define their own conversion logic for custom types when processing query parameters, similar to how it's done for request bodies.

**Describe alternatives you've considered**

An alternative approach to address the serialization and deserialization of custom types in query parameters could involve utilizing the `JsonApiResourceDefinition`. In this scenario, the `Wrapper` class could be treated as a resource, and a corresponding `JsonApiResourceDefinition` could be created. However, since `Wrapper` isn't a resource, this would not be advisable.

### LINKED ISSUE #1277: How to implement a custom filter
#### SUMMARY

I'm trying to allow custom filter strings like this:

`http://localhost/ressource?filter=lessThan(createdAt,'-00:10:00')`

which would mean to return all ressources where the `createdAt` datetime value is less than 10 minutes in the past.

In JADNC before 5.2.0 this worked when using a QueryRewriter like my `TimespanExpressionRewriter` below. Since version 5.2.0, JADNC throws an exception: "Failed to convert '-00:10:00' of type 'String' to type 'DateTime'."

I suppose the `TimespanExpressionRewriter` is not the right way to go. Hence my question: How do we implement a custom filter like that?

I've already dug around in the JADNC sources without success.

#### DETAILS

My `TimespanExpressionRewriter` class overrides `VisitComparison()` and replaces an expression when the right value looks like `-00:00:00`. It returns the rewritten expression like this:

- Input: `lessThan(updatedAt,'-1.06:00')`
- Output: `and(greaterThan(updatedAt,'2000-06-30T10:00:00+02:00'),lessOrEqual(updatedAt,'2000-07-01T16:00:00+02:00'))`

In JADNC <5.2.0 this worked, when adding the expression rewriting in a `ResourceDefinition` like this:

```cs
public override FilterExpression? OnApplyFilter(FilterExpression? existingFilter)
{
    if (existingFilter != null)
    {
        // Timespan expression
        // Example:
        // http://localhost/ressources?filter=lessThan(createdAt%2C'-0.00%3A50%3A00')

        existingFilter = new TimespanExpressionRewriter().Visit(existingFilter, null) as FilterExpression;
    }

    return existingFilter;
}
```

TimespanExpressionRewriter:
```cs
public class TimespanExpressionRewriter : QueryExpressionRewriter<object?>
{
    public override QueryExpression? VisitComparison(ComparisonExpression expression, object? argument)
    {
        // TryCreate() replaces "expression" if it contains a relative timespan
        if (TimespanExpressionFactory.TryCreate(expression, () => DateTime.UtcNow, out var expr))
        {
            return expr;
        }

        return base.VisitComparison(expression, argument);
    }

    public override QueryExpression? Visit(QueryExpression expression, object argument)
    {
        return base.Visit(expression, argument);
    }
}
```

#### VERSIONS USED
- JsonApiDotNetCore version: 5.2.0
- ASP.NET Core version: 7
- Entity Framework Core version: 7.0.5
- Database provider: MSSQL

### LINKED ISSUE #1283: No error when comparing count() with null
#### DESCRIPTION
Filter parsing should produce an error on comparison of `count()` with `null`.

#### STEPS TO REPRODUCE
Using the GettingStarted project, the following request:

```http
GET /api/people?filter=equals(count(books),null) HTTP/1.1
```

succeeds, without matching anything.

#### EXPECTED BEHAVIOR
An error is produced, because the number of rows (int) can never match `null`.

Something along the lines of:

```json
{
  "errors": [
    {
      "id": "1792eb76-56f3-45f5-bea0-c9a318d475e3",
      "status": "400",
      "title": "The specified filter is invalid.",
      "detail": "Failed to convert 'null' of type 'String' to type 'Int32'.",
      "source": {
        "parameter": "filter"
      }
    }
  ]
}
```

#### ACTUAL BEHAVIOR
No error is produced. The request succeeds without matching anything.

bkoelman:
This PR adds an extensibility point for API developers to add custom converters that turn literal text (a value surrounded by single quotes) in filter query string parameters into a .NET object. The resulting object may or may not be compatible with the property type. In the latter case, a custom expression rewriter can be used to manipulate the filter to match up, potentially adding additional conditions. Such a rewriter is typically invoked from `IResourceDefinition<,>.OnApplyFilter()`. An example is added in this PR to demonstrate that.

A converter must implement `IFilterValueConverter`. Its `CanConvert` method takes an `AttrAttribute`, which provides access to the attribute name and type, and its containing type in the resource graph. Multiple converters can be registered in the IoC container, which are tried sequentially. The `Convert` method performs the actual conversion and executes when `CanConvert` returns `true`. It takes the same `AttrAttribute`, the string value to convert from (without the surrounding quotes), and the outer filter function that consumes the value. The return type is `object`, which enables to return something incompatible to be rewritten later. It must never return `null`, because not all expression types support that. As a workaround, return a sentinel value (for example, [`Missing.Value`](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.missing.value)) that is recognized by your rewriter. The consuming filter functions can be _Comparison_ (`equals`, `greaterThan`, `greaterOrEqual`, `lessThan`, `lessOrEqual`), _Any_ (pick from a set of candidates), or _TextMatch_ (`contains`, `startsWith`, `endsWith`). The last one requires that your rewriter produces a `string`. If that's not possible, or the conversion fails for any other invalid input, throw a `QueryParseException`. By doing so, it gets translated into an appropriate JSON:API error response. Throwing anything else results in an HTTP 500 error.

Alternative designs considered:
- Use [`TypeConverter`](https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/ayybcxe5(v=vs.120)). This requires developers to add `[TypeConverter(typeof(YourConverter))]` on resource properties, where `YourConverter` derives from `System.ComponentModel.TypeConverter`, overriding `CanConvertTo/From` and `ConvertTo/From`. This requires fine-grained control per property (which may be liked or not), but the downside is that everything is globally wired up, so IoC dependencies cannot be injected into the converter. It is also tricky to get right because ModelState validation triggers your type converter too. This API doesn't fit our needs well: we always convert _from_ string, and _to_ something the converter itself decides, which is unknown upfront.
- Use [`ValueConverter<TModel, TProvider>`](https://learn.microsoft.com/en-us/ef/core/modeling/value-conversions?tabs=data-annotations) from Entity Framework Core. They convert to/from an entity property type to a database type (for example, to store enums as strings). Aside from taking a dependency on EF Core for parsing query strings doesn't sound appealing, these are configured through EF Core's fluent API, which JsonApiDotNetCore doesn't have. In EF Core, they can be applied per entity property, or convention-based. It is also possible to configure a converter _instance_, whose construction can depend on injected dependencies in the `DbContext`. The static typing works against us because we'd like the converter to flexibly choose different return types, based on the incoming text. Using the non-generic `ValueConverter` base class doesn't solve that. What's also unfortunate is that the conversion must be a `System.Expression`, which doesn't permit to use modern C# syntax such as the `?.` operator. Expressions are required because these converters are weaved into EF Core's projection shapers that turn SQL result sets into .NET objects.

Additionally, the first commit in this PR improves validation of filters and eliminates the multi-pass parsing of `null` comparisons against to-one relationships.

Fixes #1277.
Fixes #1275.
Fixes #1283.

#### QUALITY CHECKLIST
- [x] Adapted tests


bkoelman:
Superseded by #1286.

PR #1297, opened 2023-08-17
Activate JetBrains Qodana scanner

bkoelman:


bkoelman:
Turned off Qodana for now, it's still too immature.

PR #1382, opened 2023-11-08
Merge master (dapper example) into openapi

bkoelman:
TODO: Why the dramatic decrease in coverage? It looks like the `test/OpenApiTests` directory is now considered non-test code, possibly because `test/OpenApiEndToEndTests.csproj` has a dependency on it.


bkoelman:
Superseded by #1388

PR #1459, opened 2024-02-07
openapi: include headers

### LINKED ISSUE #1367: Include HTTP headers
Swagger UI currently does not display HTTP request/response headers.

This should be possible, according to https://github.com/mattfrear/Swashbuckle.AspNetCore.Filters#add-a-response-header.

Headers to describe in OAS:
- `Location` (out) on POST resource endpoint
- `If-None-Match` (in) on GET/HEAD endpoints
- `ETag` (out) on GET/HEAD endpoints
- `Content-Length` (out) on HEAD requests

As a safety mechanism, an incoming `If-Match` produces an error that indicates wrong usage. There's no need to include it in OAS.

HTTP-level caching is one of the key advantages of using JSON:API, compared to competing technologies. So we should strive to offer the experience described at https://www.jsonapi.net/usage/caching.html using generated client libraries.

verdie-g:
Closes #1367

#### QUALITY CHECKLIST
- [X] Complies with our [contributing guidelines](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/.github/CONTRIBUTING.md)
- [X] Adapted tests
- [x] Documentation updated

image


src/JsonApiDotNetCore.OpenApi/SwaggerComponents/JsonApiOperationDocumentationFilter.cs
```diff
@@ -420,6 +430,31 @@ private static void SetRequestBodyDescription(OpenApiRequestBody requestBody, st
...
+    {
+        OpenApiResponse response = GetOrCreateResponse(responses, statusCode);
+        response.Description = XmlCommentsTextHelper.Humanize(description);
+    }
+
+    private static void SetResponseHeaderETag(OpenApiResponses responses, HttpStatusCode statusCode)
+    {
+        OpenApiResponse response = GetOrCreateResponse(responses, statusCode);
+        response.Headers[HeaderNames.ETag] = new OpenApiHeader
+        {
```
verdie-g, line 442: Might need to add Required=true here if required in this context means that it will always be present.

verdie-g: Found my answer here https://github.com/OAI/OpenAPI-Specification/issues/321#issuecomment-284074270

verdie-g:
This change does make the header appear in the OpenAPI document but the swagger generation remains the same. The reason is that the methods of the generated clients only return the deserialized body. I think an option must be enabled to actually return an object with status code and headers (https://github.com/RicoSuter/NSwag/issues/3557#issuecomment-922370812).

verdie-g:
image

An NSwag method looks like that
image
The argument name `if_None_Match` is not great but what's even less great are these lines
```cs
if (status_ == 304)
{
    string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
    throw new ApiException("The resource was not modified.", status_, responseText_, headers_, null);
}
```
It throws on 403. Though, I'm not sure what would be a better behavior.

bkoelman:
> The argument name if_None_Match is not great

Can we use https://github.com/RicoSuter/NSwag/issues/808 somehow?

> It throws on 403

We have https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/openapi/src/JsonApiDotNetCore.OpenApi.Client/ApiResponse.cs for that; probably needs to be adapted to handle 403 as well.

I've clarified the description in the original issue. For NSwag, we may even provide our own ResponseClass (pass `/GenerateResponseClasses:false`) if that smoothens the caching experience. For Kiota, see https://learn.microsoft.com/en-us/openapi/kiota/query-parameters-request-headers#request-headers and https://github.com/microsoft/kiota/issues/2963.

verdie-g:
> Can we use https://github.com/RicoSuter/NSwag/issues/808 somehow?

Probably but I'm not sure we can use by just passing a command line option so maybe that will fall outside of the scope of the PR.

Also I'm so confused about how clients are generated.

This code clones the repo, makes a change and can't find that change in the generate clients. What am I missing?
```bash
git clone git@github.com:json-api-dotnet/JsonApiDotNetCore.git test
cd test
git checkout openapi
dotnet build

sed -i 's/"query"/"hello"/' src/JsonApiDotNetCore.OpenApi/SwaggerComponents/JsonApiOperationDocumentationFilter.cs
dotnet build

grep hello test/OpenApiClientTests/obj/OpenApiClient.cs # nothing

dotnet clean
dotnet build

grep hello test/OpenApiClientTests/obj/OpenApiClient.cs # nothing

git clean -fdx
dotnet clean
dotnet build

grep hello test/OpenApiClientTests/obj/OpenApiClient.cs # nothing
grep hello test/OpenApiEndToEndTests/obj/QueryStringsClient.cs # nothing
```

bkoelman:
> > Can we use [RicoSuter/NSwag#808](https://github.com/RicoSuter/NSwag/issues/808) somehow?
> 
> Probably but I'm not sure we can use by just passing a command line option so maybe that will fall outside of the scope of the PR.

I don't understand, can you clarify?

> Also I'm so confused about how clients are generated.

When building project OpenApiTests, your changes from JsonApiDotNetCore.OpenApi are compiled. When _running the tests_ in OpenApiTests, it writes updated `swagger.json` into the OpenApiClientTests and/or OpenApiEndToEndTests test project (for example, [here](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/openapi/test/OpenApiTests/QueryStrings/QueryStringTests.cs#L19)). When adding Kiota support, I'd like to change that to output into the server-test project itself, so that multiple clients can reference it.

When building JsonApiDotNetCoreExample, it starts the webserver and writes `GeneratedSwagger/JsonApiDotNetCoreExample.json` (which is consumed by JsonApiDotNetCoreExampleClient) because of [this line](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/openapi/src/Examples/JsonApiDotNetCoreExample/JsonApiDotNetCoreExample.csproj#L4). Turn it off when the JsonApiDotNetCore.OpenApi produces something broken, so you can run the example project and debug it.

bkoelman:
Also, NSwag doesn't always detect changes properly. For example, if project options have changed but the source swagger.json has not, it won't regenerate. Delete the `obj` directory in the client project to workaround that.

verdie-g:
> Delete the obj directory in the client project to workaround that.

Oh that's it thanks.

> We have https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/openapi/src/JsonApiDotNetCore.OpenApi.Client/ApiResponse.cs for that; probably needs to be adapted to handle 403 as well.

Though I'm not sure what to return in that case. I've generated a response object to be able to get the response headers (e.g. ETag) so maybe it could return `{ status: 403, headers: /* ... */, data: null }` and the user should know that if status is 403 then data will be null.

> I don't understand, can you clarify?

NSwag clients are generated using a command line from the csproj. How can I pass an implementation of IParameterNameGenerator there 🤔

bkoelman:
@verdie-g Looking at your last commit message, you may have faced the same mystery as I did this week. See #1463.

verdie-g:
Oh actually I think I just didn't know anymore what I changed so I wrote some random message.

src/JsonApiDotNetCore.OpenApi/SwaggerComponents/JsonApiOperationDocumentationFilter.cs
```diff
@@ -420,6 +459,59 @@ private static void SetRequestBodyDescription(OpenApiRequestBody requestBody, st
...
+        };
+    }
+
+    private static void SetResponseHeaderContentLength(OpenApiResponses responses, HttpStatusCode statusCode)
+    {
+        OpenApiResponse response = GetOrAddResponse(responses, statusCode);
+
+        response.Headers[HeaderNames.ContentLength] = new OpenApiHeader
+        {
+            Description = "Size of the response body in bytes",
```
bkoelman, line 489: Suggestion:
> Size of the HTTP response body, in bytes.

src/JsonApiDotNetCore.OpenApi/SwaggerComponents/JsonApiOperationDocumentationFilter.cs
```diff
@@ -462,4 +554,21 @@ private static void AddQueryStringParameters(OpenApiOperation operation, bool is
...
         });
     }
+
+    private static void AddHeaderParameterIfNoneMatch(OpenApiOperation operation)
+    {
+        operation.Parameters.Add(new OpenApiParameter
+        {
+            In = ParameterLocation.Header,
+            Name = "If-None-Match",
+            Description = "ETag identifying the version of the requested resource.",
```
bkoelman, line 564: Suggestion:
> A list of ETags, resulting in HTTP status 304 without a body, if one of them matches the current fingerprint.

test/OpenApiClientTests/OpenApiClientTests.csproj
```diff
@@ -28,56 +29,56 @@
       <ClassName>OpenApiClient</ClassName>
       <OutputPath>OpenApiClient.cs</OutputPath>
       <CodeGenerator>NSwagCSharp</CodeGenerator>
-      <Options>/GenerateClientInterfaces:true /ClientClassAccessModifier:internal /GenerateExceptionClasses:false /AdditionalNamespaceUsages:JsonApiDotNetCore.OpenApi.Client.Exceptions</Options>
+      <Options>/GenerateClientInterfaces:true /ClientClassAccessModifier:internal /GenerateExceptionClasses:false /GenerateOptionalParameters:true /AdditionalNamespaceUsages:JsonApiDotNetCore.OpenApi.Client.Exceptions</Options>
```
bkoelman, line 32: I'd rather not have this switch added in test projects. While it leads to shorter code, it also hides potential errors when a new parameter gets added that wasn't intended.

src/JsonApiDotNetCore.OpenApi/SwaggerComponents/JsonApiOperationDocumentationFilter.cs
```diff
@@ -462,4 +554,21 @@ private static void AddQueryStringParameters(OpenApiOperation operation, bool is
...
+            Required = false,
```
bkoelman, line 565: Isn't `false` the default? If so, this line can be omitted.

src/JsonApiDotNetCore.OpenApi/SwaggerComponents/JsonApiOperationDocumentationFilter.cs
```diff
@@ -462,4 +554,21 @@ private static void AddQueryStringParameters(OpenApiOperation operation, bool is
             Description = isRelationshipEndpoint ? RelationshipQueryStringParameters : ResourceQueryStringParameters
         });
     }
+
+    private static void AddHeaderParameterIfNoneMatch(OpenApiOperation operation)
```
bkoelman, line 558: Should probably be named `AddRequestHeaderIfNoneMatch` for consistency with the other added methods).

src/JsonApiDotNetCore.OpenApi/SwaggerComponents/JsonApiOperationDocumentationFilter.cs
```diff
@@ -420,6 +459,59 @@ private static void SetRequestBodyDescription(OpenApiRequestBody requestBody, st
...
+    {
+        OpenApiResponse response = GetOrAddResponse(responses, statusCode);
+
+        response.Headers[HeaderNames.Location] = new OpenApiHeader
+        {
+            Description = "Location of the newly created resource.",
+            Required = true,
+            Schema = new OpenApiSchema
+            {
+                Type = "string"
```
bkoelman, line 509: I wonder if anything would break if we added `format: uri` (see https://swagger.io/docs/specification/data-models/data-types/#format). Note that the value we emit today is always a relative URI, but that may change. 

Actually, I think it would be more correct to take `JsonApiOptions.UseRelativeLinks` into account, but that change needs to happen in the master branch, so out of scope of this PR.

bkoelman:
Thanks for looking into all these details. You're on the right track. Sorry if this turned out more complicated than I initially imagined, I appreciate you're not giving up.

> > I don't understand, can you clarify?
> 
> NSwag clients are generated using a command line from the csproj. How can I pass an implementation of IParameterNameGenerator there 🤔 I think a custom generator needs to be created like in here https://github.com/daveaglick/NetlifySharp/blob/main/tools/NetlifySharp.Generator/Program.cs.

Thanks, I understand now. It goes a bit too far to build our own generator for this, so let's just stick with `if_None_Match`. Feel free to open an issue in the NSwag repo with a request to add a switch to adhere to the C# naming conventions.

> > We have https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/openapi/src/JsonApiDotNetCore.OpenApi.Client/ApiResponse.cs for that; probably needs to be adapted to handle 403 as well.
> 
> Though I'm not sure what to return in that case. I've generated a response object to be able to get the response headers (e.g. ETag) so maybe it could return `{ status: 403, headers: /* ... */, data: null }` and the user should know that if status is 403 then data will be null.

I propose to extend the `ApiResponse` class in JADNC.OpenApi.Client project in the following way:
- Add classes `ApiResponse` and `ApiResponse<>`, similar to what NSwag generates. I don't think the name JsonApiResponse is appropriate, as it contains nothing specific to JSON:API.
- Extend existing methods to respond to `304` and return `null`.
- Add new methods (for returning void and T) for wrapped responses (a user may choose to use them or not), ie:
    ```c#
    public static async Task<ApiResponse<T?>> TranslateWrappedAsync<T>(Func<Task<ApiResponse<T>>> operation)
        where T : class
    {
        try
        {
            ApiResponse<T> response = await operation();
            return new ApiResponse<T?>(response.StatusCode, response.Headers, response.Result);
        }
        catch (ApiException exception) when (exception.StatusCode == (int)HttpStatusCode.NotModified)
        {
            return new ApiResponse<T?>(exception.StatusCode, exception.Headers, null);
        }
    }
    ```
- Move `ApiException` outside of the `.Exceptions` namespace, so we'll only need a single `/AdditionalNamespaceUsages` switch.

bkoelman (changes requested):
Aside from comments, remaining things to address:
- https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/1459#issuecomment-1949582165
- End-to-end tests (subset of what's in `ETagTests.cs`)
- Possibly update sample to show how ETag works, ie: call wrapped `GetPersonCollectionAsync`, capture ETagfrom header, call again with that ETag, and observe the result is `null` with status 304.
- Document the (optional) use of `/WrapResponses:true /GenerateResponseClasses:false` in docs/usage/openapi.md
- Try if it works with Kiota (no commit needed, but would be nice if you can add some working code in a PR comment so I can integrate it later)

src/Examples/JsonApiDotNetCoreExampleClient/Program.cs
```diff
@@ -47,6 +52,19 @@
 Console.WriteLine("Press any key to close.");
 Console.ReadKey();
 
+#pragma warning disable AV1553
+static Task<JsonApiResponse<PersonCollectionResponseDocument>> GetPersonCollectionAsync(ExampleApiClient apiClient, string? ifNoneMatch = null)
```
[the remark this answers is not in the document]
verdie-g: I'm calling it twice so it's just to avoid copying the code.

verdie-g:
> Document the (optional) use of /WrapResponses:true /GenerateResponseClasses:false in docs/usage/openapi.md

Hmm I was expecting to see a NSwag section there, not sure what to add.

> Try if it works with Kiota (no commit needed, but would be nice if you can add some working code in a PR comment so I can integrate it later)

I don't think I'll have the time/energy to test with Kiota. Though this library looks promising.

test/OpenApiEndToEndTests/QueryStrings/ETagTests.cs
```diff
@@ -0,0 +1,76 @@
...
+using JsonApiDotNetCore.OpenApi.Client;
+using Microsoft.Net.Http.Headers;
+using OpenApiEndToEndTests.QueryStrings.GeneratedCode;
+using OpenApiTests;
+using OpenApiTests.QueryStrings;
+using TestBuildingBlocks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace OpenApiEndToEndTests.QueryStrings;
```
bkoelman, line 12: These tests don't really belong here. Ideally, they would live in OpenApiEndToEndTests.Headers, with their own separate model, DbContext and fakers, unrelated to query string client tests. I'm not so happy to see this PR requires changes to all the query string test cases in this project. Those should not use wrap-response syntax.

Can you come up with a separate sample model and isolate it?

bkoelman (changes requested):
Please update this PR to the latest version of the base branch.

test/OpenApiTests/Headers/HeaderTests.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        testContext.UseController<CountriesController>();
+
+        testContext.SwaggerDocumentOutputDirectory = "test/OpenApiEndToEndTests/Headers";
+    }
+
+    [Fact]
+    public Task Dummy()
+    {
+        return _testContext.GetSwaggerDocumentAsync();
```
verdie-g, line 21: There is probably a better way to do that

bkoelman:
For future reference, here's what the sample looks like using Kiota:
```c#
var headerInspector = new HeadersInspectionHandlerOption
{
    InspectResponseHeaders = true
};

PersonCollectionResponseDocument? getResponse1 = await client.Api.People.GetAsync(
    configuration => configuration.Options.Add(headerInspector));

string eTag = headerInspector.ResponseHeaders["ETag"].Single();

try
{
    PersonCollectionResponseDocument? getResponse2 = await client.Api.People.GetAsync(
        configuration => configuration.Headers.Add("If-None-Match", eTag));
}
catch (ApiException exception) when (exception.ResponseStatusCode == (int)HttpStatusCode.NotModified)
{
    Console.WriteLine("The HTTP response hasn't changed, so no response body was returned.");
}
```
The try/catch is needed due to a bug in Kiota: it fails to map the 304 status, despite being in the OAS.

bkoelman (changes requested):

bkoelman:
Please add `HttpStatusCode.NotModified` to the appropriate entries in `OpenApiEndpointConvention.GetSuccessStatusCodesForEndpoint`, so that status codes are ordered alphabetically in the OAS document.

test/OpenApiEndToEndTests/OpenApiEndToEndTests.csproj
```diff
@@ -23,12 +23,19 @@
   </ItemGroup>
 
   <ItemGroup>
+    <OpenApiReference Include="Headers\swagger.g.json">
+      <Namespace>OpenApiEndToEndTests.Headers.GeneratedCode</Namespace>
+      <ClassName>HeadersClient</ClassName>
+      <OutputPath>HeadersClient.cs</OutputPath>
+      <CodeGenerator>NSwagCSharp</CodeGenerator>
+      <Options>/ClientClassAccessModifier:internal /GenerateExceptionClasses:false /WrapResponses:true /GenerateResponseClasses:false /ResponseClass:ApiResponse /AdditionalNamespaceUsages:JsonApiDotNetCore.OpenApi.Client /GenerateOptionalParameters:true /GenerateNullableReferenceTypes:true</Options>
```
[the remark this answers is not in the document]
verdie-g: It seems like the default value is `SwaggerResponse`.

test/OpenApiTests/Headers/HeaderTests.cs
verdie-g: Isn't it what `DocCommentsTests.cs` is already testing?

bkoelman: Partly, yes. You probably noticed that simple changes cause a lot of churn in many OAS files. We're set up like that _currently_ because we'd like to be notified of any impact (and potential undesired side effects) of structural changes. As the number of tests grows, this becomes a burden. The vision is to move these files under `.gitignore` in the future, after we have established good test coverage.

So we need test coverage that asserts on all the bits and pieces we expect to be produced. For headers, this means producing the right headers at the expected places, but also that they don't occur at places where they shouldn't be. `DocCommentsTests` focuses on the documentation only (and in order to do that, it needs to reference a location in the OAS). If we were to turn off doc-comments, headers still need to be produced at the right places and with the right details. That's what these tests are for.

I think that `DocCommentsTests` will go away in the future, as well as (most of) `JsonApiOperationDocumentationFilter`. It was my first contribution after picking up from Maurits, and I'm still learning. In hindsight, I think the documentation should be generated at the places where nodes are produced, not afterward in a post-processing step as it is now. Producing documentation in all the right places isn't much of a feature itself, but merely an aspect. As test coverage grows, we should probably move these assertions there, as those tests already assert on the location where these nodes need to be.

This is my current thinking, which may change over time. This is not all set in stone and I'm open to new ideas. It's hard to know upfront how OpenAPI tests are best organized. So unfortunately, we may be doing work today that's going to disappear eventually. That's frustrating sometimes, but I don't see any other way that enables us to confidently refactor internals, knowing that the required test coverage is in place.

verdie-g: Alright, I'll work do that later in the day.

test/OpenApiTests/Headers/HeaderTests.cs
```diff
@@ -0,0 +1,146 @@
...
+    [Theory]
+    [InlineData("/countries.get")]
+    [InlineData("/countries.head")]
+    [InlineData("/countries/{id}.get")]
+    [InlineData("/countries/{id}.head")]
+    [InlineData("/countries/{id}/languages.get")]
+    [InlineData("/countries/{id}/languages.head")]
+    [InlineData("/countries/{id}/relationships/languages.get")]
+    [InlineData("/countries/{id}/relationships/languages.head")]
+    public async Task Get_and_head_endpoints_have_caching_headers(string endpointPath)
```
bkoelman, line 30: I don't think the test name should include the inputs it is being used with.
```suggestion
    public async Task Endpoints_have_caching_headers(string endpointPath)
```

test/OpenApiTests/Headers/HeaderTests.cs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    [Theory]
+    [InlineData("/countries.post")]
+    [InlineData("/countries/{id}.patch")]
+    [InlineData("/countries/{id}.delete")]
+    [InlineData("/countries/{id}/relationships/languages.post")]
+    [InlineData("/countries/{id}/relationships/languages.patch")]
+    [InlineData("/countries/{id}/relationships/languages.delete")]
+    public async Task Post_patch_and_delete_endpoints_do_not_have_caching_headers(string endpointPath)
```
bkoelman, line 80:
```suggestion
    public async Task Endpoints_do_not_have_caching_headers(string endpointPath)
```

test/OpenApiTests/Headers/HeaderTests.cs
```diff
@@ -0,0 +1,146 @@
...
+            parametersElement.EnumerateArray().Should().NotContain(parameterElement => parameterElement.GetProperty("name").ValueEquals("If-None-Match"));
+        });
+    }
+
+    [Theory]
+    [InlineData("/countries.head")]
+    [InlineData("/countries/{id}.head")]
+    [InlineData("/countries/{id}/languages.head")]
+    [InlineData("/countries/{id}/relationships/languages.head")]
+    public async Task Head_endpoints_have_content_length_response_header(string endpointPath)
```
bkoelman, line 106:
```suggestion
    public async Task Endpoints_have_content_length_response_header(string endpointPath)
```

test/OpenApiTests/Headers/HeaderTests.cs
```diff
@@ -0,0 +1,146 @@
...
+
+    [Theory]
+    [InlineData("/countries.post")]
+    public async Task Post_endpoints_have_location_response_header(string endpointPath)
+    {
+        // Act
+        JsonElement document = await _testContext.GetSwaggerDocumentAsync();
+
+        // Assert
+        document.Should().ContainPath($"paths.{endpointPath}.responses.201.headers.Location").With(locationElement =>
```
bkoelman, line 133:
```suggestion
        document.Should().ContainPath("paths./countries.post.responses.201.headers.Location").With(locationElement =>
```

test/OpenApiTests/Headers/HeaderTests.cs
```diff
@@ -0,0 +1,146 @@
...
+            contentLengthElement.Should().ContainPath("schema").With(schemaElement =>
+            {
+                schemaElement.Should().HaveProperty("type", "integer");
+            });
+        });
+    }
+
+    [Theory]
+    [InlineData("/countries.post")]
+    public async Task Post_endpoints_have_location_response_header(string endpointPath)
```
bkoelman, line 127:
```suggestion
    public async Task Post_resource_endpoint_has_location_response_header()
```

bkoelman (changes requested):
Looks good, minor comments aside.

src/JsonApiDotNetCore.OpenApi/SwaggerComponents/JsonApiOperationDocumentationFilter.cs
```diff
@@ -420,6 +459,58 @@ private static void SetRequestBodyDescription(OpenApiRequestBody requestBody, st
...
+        };
+    }
+
+    private static void SetResponseHeaderLocation(OpenApiResponses responses, HttpStatusCode statusCode)
+    {
+        OpenApiResponse response = GetOrAddResponse(responses, statusCode);
+
+        response.Headers[HeaderNames.Location] = new OpenApiHeader
+        {
+            Description = "The URL at which the newly created JSON:API resource can be retrieved.",
```
bkoelman, line 503: I suddenly realized this looks inconsistent:

image

To fix this, this method needs to take the `singularName` value from the caller as a parameter.

test/OpenApiTests/LegacyOpenApiIntegration/swagger.json
```diff
@@ -4781,7 +6251,7 @@
           },
           "meta": {
             "type": "object",
-            "additionalProperties": { },
+            "additionalProperties": {},
```
bkoelman, line 6254: I wonder why this changed? Don't copy/paste the file contents in your IDE to sync up, because that reformats the file. Instead, copy the file itself.

bkoelman (changes requested):
Final round of review. Thanks for all the hard work!
I'll propose something to add to the docs; with that included, this looks ready for merge.
Meanwhile, can you update the PR description, with an updated SwaggerUI screenshot?

test/OpenApiTests/LegacyOpenApiIntegration/swagger.json
verdie-g: I copy the actual result from the test failure output. How do you do it?

bkoelman: I have a desktop shortcut to a .bat file containing:
```
copy /Y C:\Source\Repos\JsonApiDotNetCore\test\OpenApiClientTests\LegacyClient\swagger.g.json C:\Source\Repos\JsonApiDotNetCore\test\OpenApiTests\LegacyOpenApiIntegration\swagger.json
```

verdie-g:
I'm not committing the suggestions one by one ever again :D

bkoelman:
> I'm not committing the suggestions one by one ever again :D

I'd welcome that. You can batch them into a single commit from the **Files changed** tab.

I've created a suggestion to update the docs, see the diff at https://github.com/json-api-dotnet/JsonApiDotNetCore/compare/openapi...openapi-header-docs. Probably just easiest for you to cherry-pick the commit, assuming you agree with the change.

verdie-g:
> You can batch them into a single commit from the Files changed tab.

Good to know.

> Probably just easiest for you to cherry-pick the commit

Done

test/OpenApiTests/Headers/HeaderFakers.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    private readonly Lazy<Faker<Country>> _lazyCountryFaker = new(() => new Faker<Country>()
+        .UseSeed(GetFakerSeed())
+        .RuleFor(country => country.Name, faker => faker.Address.Country())
+        .RuleFor(country => country.Population, faker => faker.Random.Long(0, 2_000_000_000)));
+
+    private readonly Lazy<Faker<Language>> _lazyLanguageFaker = new(() => new Faker<Language>()
+        .UseSeed(GetFakerSeed())
+        .RuleFor(textLanguage => textLanguage.Name, faker => faker.PickRandom(LazyLanguageNames.Value))
+        .RuleFor(textLanguage => textLanguage.Code, faker => faker.PickRandom(LazyLanguageCodes.Value)));
```
bkoelman, line 32: ~textLanguage~ language

test/OpenApiTests/Headers/HeaderFakers.cs
```diff
@@ -0,0 +1,36 @@
...
+// @formatter:wrap_chained_method_calls chop_if_long
+// @formatter:wrap_before_first_method_call true
+
+namespace OpenApiTests.Headers;
+
+[UsedImplicitly(ImplicitUseTargetFlags.Members)]
+public sealed class HeaderFakers : FakerContainer
+{
+    private static readonly Lazy<string[]> LazyLanguageNames = new(() => CultureInfo
+        .GetCultures(CultureTypes.NeutralCultures)
```
bkoelman, line 15: Executing multiple times is okay, these appear to be cached internally.
```
CultureInfo.GetCultures ...  0:00:00:00.0158958
CultureInfo.GetCultures ...  0:00:00:00.0001816
CultureInfo.GetCultures ...  0:00:00:00.0001518
```

test/OpenApiEndToEndTests/Headers/ETagTests.cs
```diff
@@ -0,0 +1,184 @@
...
+        ApiException<ErrorResponseDocument> exception = (await action.Should().ThrowExactlyAsync<ApiException<ErrorResponseDocument>>()).Which;
+        exception.StatusCode.Should().Be((int)HttpStatusCode.NotFound);
+        exception.Headers.Should().NotContainKey(HeaderNames.ETag);
+    }
+
+    [Fact]
+    public async Task Returns_no_ETag_for_POST_request()
+    {
+        // Arrange
+        Country country = _fakers.Country.Generate();
```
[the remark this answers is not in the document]
bkoelman: Oh, I was partially wrong here. Don't extract into separate variables, but do use the `new` prefix. Sorry for the confusion.

test/OpenApiEndToEndTests/Headers/ETagTests.cs
```diff
@@ -0,0 +1,183 @@
...
+        ApiException<ErrorResponseDocument> exception = (await action.Should().ThrowExactlyAsync<ApiException<ErrorResponseDocument>>()).Which;
+        exception.StatusCode.Should().Be((int)HttpStatusCode.NotFound);
+        exception.Headers.Should().NotContainKey(HeaderNames.ETag);
+    }
+
+    [Fact]
+    public async Task Returns_no_ETag_for_POST_request()
+    {
+        // Arrange
+        Country newCountry = _fakers.Country.Generate();
```
bkoelman, line 98:
```suggestion
        Country newCountry = _fakers.Country.Generate();

```

bkoelman (approved):

Issue #1299, opened 2023-08-18
AllowClientGeneratedIds for single resource


verdie-g:
**Is your feature request related to a problem? Please describe.**

I would like to create a user for a steam game and reuse their steam id as the user id in my database. 

The user is defined as
```csharp
[Resource]
public class User : Identifiable<long>
{
}
```
but if I do
```
POST http://localhost:5203/users
Content-Type: application/vnd.api+json

{
  "data": {
    "type": "users",
    "id": "76561197960279927"
  }
}
```
I'll get this error
```
HTTP/1.1 403 Forbidden
Content-Type: application/vnd.api+json
Date: Fri, 18 Aug 2023 20:07:58 GMT
Server: Kestrel
Transfer-Encoding: chunked

{
  "links": {
    "self": "http://localhost:5203/users"
  },
  "errors": [
    {
      "id": "31676f4a-725e-44a3-9bd7-8013c0cf2f88",
      "status": "403",
      "title": "Failed to deserialize request body: The use of client-generated IDs is disabled.",
      "source": {
        "pointer": "/data/id"
      }
    }
  ]
}
```

I could use the option `AllowClientGeneratedIds` but it's applied to all resources and I only want it scoped to the user resource.

**Describe the solution you'd like**

```diff
- [Resource]
+ [Resource(AllowClientGeneratedIds = true)]
public class User : Identifiable<long>
{
}
```

**Describe alternatives you've considered**

- Define the steam id as an attribute but every time a steam id is mentioned in-game, I would have to do the mapping to my internal id to understand which users we're talking about

bkoelman:
I always wondered why this was only possible globally, and why nobody ever asked for this. It's a legitimate ask, I'm just not sure yet how to best provide that. Probably needs to be stored in the resource graph, settable by an attribute parameter. 

Until this is implemented, it would be easiest for you to define a generic resource definition class and register it with open generics. Override the OnPrepareWriteAsync method, checking for the endpoint kind (post) and resource type (user), otherwise throwing when Id is non-default. I'm on vacation right now, so can't provide working code.

bkoelman:
Hi @verdie-g, this is implemented in #1305. Can you give it a try? Feedback welcome.

verdie-g:
Looks good to me 👌

Issue #139, opened 2017-06-27
TypeHelper doesn't handle empty string


jaredcnance:
something like:
```csharp
var stringValue = value.ToString();
if(stringValue == string.Empty)
  return default(T);
```

PR #1475, opened 2024-02-23
openapi: support client-generated id

### LINKED ISSUE #1054: Support for client-generated IDs
Currently the OpenApi integration takes into account the `JsonApiOptions.AllowClientGeneratedIds` option when generating schemas for resource objects, but this hasn't been thoroughly tested.

verdie-g:
Closes #1054 

#### QUALITY CHECKLIST
- [x] Adapted tests
- [ ] Documentation updated


test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        using HttpClient httpClient = _testContext.Factory.CreateClient();
+        ClientGeneratedIdClient apiClient = new(httpClient);
+
+        // Act
+        Func<Task<PlayerPrimaryResponseDocument>> action = () => apiClient.PostPlayerAsync(null, new PlayerPostRequestDocument
+        {
+            Data = new PlayerDataInPostRequest
+            {
+                Id = null!, // FIXME: passing "" here works fine 🤔
```
verdie-g, line 37: Is that expect?

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
```diff
@@ -0,0 +1,133 @@
...
+                {
+                    Name = player.Name
+                }
+            }
+        });
+
+        // Assert
+        var exception = (await action.Should().ThrowAsync<Exception>()).Subject.First();
+        // Exception is Newtonsoft.Json.JsonSerializationException: Cannot write a null value for property 'id'. Property requires a value. Path 'data'.
+        // Probably not what we want.
```
verdie-g, line 48: Any idea how to write that test correctly?

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        using HttpClient httpClient = _testContext.Factory.CreateClient();
+        ClientGeneratedIdClient apiClient = new(httpClient);
+
+        // Act
+        Func<Task<GamePrimaryResponseDocument>> action = () => apiClient.PostGameAsync(null, new GamePostRequestDocument
+        {
+            Data = new GameDataInPostRequest
+            {
+                Id = game.StringId!, // FIXME: StringId is null, how to generate an id?
```
verdie-g, line 92: How can I generate an id here?

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        using HttpClient httpClient = _testContext.Factory.CreateClient();
+        ClientGeneratedIdClient apiClient = new(httpClient);
+
+        // Act
+        Func<Task<GamePrimaryResponseDocument>> action = () => apiClient.PostGameAsync(null, new GamePostRequestDocument
+        {
+            Data = new GameDataInPostRequest
+            {
+                Id = null!, // FIXME: incorrect nullability here
```
verdie-g, line 120: Because `ClientIdGenerationMode.Allowed` is used, I think `Id` should be nullable.

verdie-g:
@bkoelman could you help with the those comments 👆

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
bkoelman: Interesting! I've debugged it down to `RuntimeTypeConverter`, which returns the target type's default value for an empty string. And there are tests for that behavior. However, I have no idea _why_ it works like that. It was introduced in #139, all the way back in 2017.

When removing that behavior, all the tests are still green. However this is a very risky thing to change.

@maurei does this ring any bells? What's happening here is that a client sends `"data": { "id": "", ... }` in a post resource request, with the configuration that client-generated IDs are required.

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
bkoelman: This is Newtonsoft intentionally failing, because OAS defines the property as required, so NSwag generates the appropriate Newtonsoft attributes to reflect that. Similar to https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/openapi/test/OpenApiClientTests/ResourceFieldValidation/NullableReferenceTypesOn/ModelStateValidationOn/CreateResourceTests.cs#L158.

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
bkoelman: By assigning a value right after the faker created an instance. Similar to https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/test/JsonApiDotNetCoreTests/IntegrationTests/ReadWrite/Creating/CreateResourceWithClientGeneratedIdTests.cs#L36.

However, it's very unlikely to use client-generated IDs with `long` identifiers. More representative would be to use GUIDs.

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
bkoelman: No, because sending `"id": null` would cause an error. It is optional, but not nullable, from the OAS perspective. How this shows up in the generated client depends on the usage of `/GenerateNullableReferenceTypes` and `/GenerateOptionalParameters`.

bkoelman:
Tests are needed to cover:
- Sending without ID
- Sending with ID

for the three modes `Forbidden`, `Allowed` and `Required`.

The easiest setup would be to have three resource types with different `[Resource(ClientIdGeneration = ...)]` for the three modes.

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
verdie-g: Actually I reused my use-case from https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/1299, i.e. creating a user for a given steam id.

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
verdie-g: Is the difference that it's allowed to omit the property but not setting it explicit to null?

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
verdie-g: But let's use a Guid to ease the test.

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
bkoelman: Yes

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
```diff
@@ -0,0 +1,164 @@
...
+                    Name = group.Name
+                }
+            }
+        }));
+
+        // Assert
+        GroupPrimaryResponseDocument? doc = (await action.Should().NotThrowAsync()).Subject;
+        doc?.Data.Id.Should().NotBeNullOrEmpty();
+    }
+}
```
verdie-g, line 164: I can't write a `Pass_forbidden_id` because the id property is generated.

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
```diff
@@ -0,0 +1,164 @@
...
+using FluentAssertions.Specialized;
+using JsonApiDotNetCore.OpenApi.Client;
+using Newtonsoft.Json;
+using OpenApiEndToEndTests.ClientGeneratedId.GeneratedCode;
+using OpenApiTests;
+using OpenApiTests.ClientGeneratedId;
+using TestBuildingBlocks;
+using Xunit;
+
+namespace OpenApiEndToEndTests.ClientGeneratedId;
```
bkoelman, line 11: The topic is `ClientIdGenerationModes`, so the namespace/directory should reflect that. Test class, fakers, DbContext, etc. should be named accordingly.

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
```diff
@@ -0,0 +1,164 @@
...
+    {
+        _testContext = testContext;
+
+        testContext.UseController<PlayersController>();
+        testContext.UseController<GamesController>();
+        testContext.UseController<GroupsController>();
+    }
+
+    [Fact]
+    public async Task Omit_required_id()
```
bkoelman, line 28: Test names should be proper English sentences that state something functional. See the existing tests outside OpenAPI for examples. This could be: Cannot_create_resource_without_ID_when_required.

test/OpenApiTests/ClientGeneratedId/Group.cs
```diff
@@ -0,0 +1,17 @@
+using JetBrains.Annotations;
+using JsonApiDotNetCore.Configuration;
+using JsonApiDotNetCore.Resources;
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace OpenApiTests.ClientGeneratedId;
+
+[UsedImplicitly(ImplicitUseTargetFlags.Members)]
+[Resource(ControllerNamespace = "OpenApiTests.ClientGeneratedId", ClientIdGeneration = ClientIdGenerationMode.Forbidden)]
+public sealed class Group : Identifiable<Guid>
```
bkoelman, line 10: PlayerGroup? or Team with Members?

test/OpenApiTests/ClientGeneratedId/Player.cs
```diff
@@ -0,0 +1,20 @@
...
+
+[UsedImplicitly(ImplicitUseTargetFlags.Members)]
+[Resource(ControllerNamespace = "OpenApiTests.ClientGeneratedId", ClientIdGeneration = ClientIdGenerationMode.Required)]
+public sealed class Player : Identifiable<Guid>
+{
+    [Attr]
+    public string Name { get; set; } = null!;
+
+    [HasMany]
+    public List<Game> Games { get; set; } = [];
```
bkoelman, line 16: What's this, purchased games, recently played games, ...?

test/OpenApiTests/ClientGeneratedId/Player.cs
```diff
@@ -0,0 +1,20 @@
...
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace OpenApiTests.ClientGeneratedId;
+
+[UsedImplicitly(ImplicitUseTargetFlags.Members)]
+[Resource(ControllerNamespace = "OpenApiTests.ClientGeneratedId", ClientIdGeneration = ClientIdGenerationMode.Required)]
+public sealed class Player : Identifiable<Guid>
+{
+    [Attr]
+    public string Name { get; set; } = null!;
```
bkoelman, line 13: Perhaps NickName, to avoid having a Name property everywhere?

test/OpenApiTests/ClientGeneratedId/Player.cs
```diff
@@ -0,0 +1,20 @@
...
+
+    [HasMany]
+    public List<Game> Games { get; set; } = [];
+
+    [HasMany]
+    public List<Group> Groups { get; set; } = [];
```
bkoelman, line 19: If these are the groups this player is a member of, then the relationship name should reflect that. But I'm just guessing what "groups" is supposed to mean.

test/OpenApiTests/ClientGeneratedId/Game.cs
```diff
@@ -0,0 +1,17 @@
...
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace OpenApiTests.ClientGeneratedId;
+
+[UsedImplicitly(ImplicitUseTargetFlags.Members)]
+[Resource(ControllerNamespace = "OpenApiTests.ClientGeneratedId", ClientIdGeneration = ClientIdGenerationMode.Allowed)]
+public sealed class Game : Identifiable<Guid>
+{
+    [Attr]
+    public string Name { get; set; } = null!;
```
bkoelman, line 13: Title?

test/OpenApiTests/ClientGeneratedId/Game.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    [Attr]
+    public decimal Price { get; set; }
```
bkoelman, line 16: PurchasePrice, MonthlyCost, ...?

test/OpenApiTests/ClientGeneratedId/ClientGeneratedIdFakers.cs
```diff
@@ -0,0 +1,29 @@
...
+public sealed class ClientGeneratedIdFakers : FakerContainer
+{
+    private readonly Lazy<Faker<Player>> _lazyPlayerFaker = new(() => new Faker<Player>()
+        .UseSeed(GetFakerSeed())
+        .RuleFor(player => player.Name, faker => faker.Person.UserName));
+
+    private readonly Lazy<Faker<Game>> _lazyGameFaker = new(() => new Faker<Game>()
+        .UseSeed(GetFakerSeed())
+        .RuleFor(player => player.Name, faker => faker.Commerce.ProductName())
+        .RuleFor(player => player.Price, faker => decimal.Parse(faker.Commerce.Price())));
```
bkoelman, line 20: these are not players

test/OpenApiTests/ClientGeneratedId/ClientGeneratedIdTests.cs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    [Fact]
+    public async Task Post_data_should_not_have_id()
+    {
+        // Act
+        JsonElement document = await _testContext.GetSwaggerDocumentAsync();
+
+        // Assert
+        document.Should().ContainPath("components.schemas.groupDataInPostRequest").With(resourceDataInPostRequestElement =>
```
bkoelman, line 65: `resourceDataInPostRequestElement` is quite long, I think `dataElement` would be sufficiently clear.

bkoelman (changes requested):
See comments.

test/OpenApiTests/ClientIdGenerationModes/Player.cs
```diff
@@ -0,0 +1,20 @@
...
+public sealed class Player : Identifiable<Guid>
+{
+    [Attr]
+    public string UserName { get; set; } = null!;
+
+    [HasMany]
+    public List<Game> OwnedGames { get; set; } = [];
+
+    [HasMany]
+    public List<PlayerGroup> GroupMemberships { get; set; } = [];
```
bkoelman, line 19: Or just MemberOf?

test/OpenApiTests/ClientIdGenerationModes/PlayerGroup.cs
```diff
@@ -0,0 +1,17 @@
+using JetBrains.Annotations;
+using JsonApiDotNetCore.Configuration;
+using JsonApiDotNetCore.Resources;
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace OpenApiTests.ClientIdGenerationModes;
+
+[UsedImplicitly(ImplicitUseTargetFlags.Members)]
+[Resource(ControllerNamespace = "OpenApiTests.ClientIdGenerationModes", ClientIdGeneration = ClientIdGenerationMode.Forbidden)]
+public sealed class PlayerGroup : Identifiable<Guid>
```
bkoelman, line 10: Probably more realistic to make this `Identifiable<long>`, as setting the ID is forbidden.

test/OpenApiTests/ClientIdGenerationModes/ClientGeneratedIdFakers.cs
```diff
@@ -0,0 +1,29 @@
+using Bogus;
```
bkoelman, line 1: Please rename files to match their containing type. There's a refactoring for that at the project/solution level.

test/OpenApiTests/ClientIdGenerationModes/ClientGeneratedIdFakers.cs
```diff
@@ -0,0 +1,29 @@
...
+public sealed class ClientIdGenerationModesFakers : FakerContainer
+{
+    private readonly Lazy<Faker<Player>> _lazyPlayerFaker = new(() => new Faker<Player>()
+        .UseSeed(GetFakerSeed())
+        .RuleFor(player => player.UserName, faker => faker.Person.UserName));
+
+    private readonly Lazy<Faker<Game>> _lazyGameFaker = new(() => new Faker<Game>()
+        .UseSeed(GetFakerSeed())
+        .RuleFor(game => game.Title, faker => faker.Commerce.ProductName())
+        .RuleFor(game => game.PurchasePrice, faker => decimal.Parse(faker.Commerce.Price())));
```
bkoelman, line 20: Use `faker.Finance.Amount()`, and set a reasonable range.

test/OpenApiEndToEndTests/ClientIdGenerationModes/PostTests.cs
```diff
@@ -0,0 +1,165 @@
...
+    {
+        _testContext = testContext;
+
+        testContext.UseController<PlayersController>();
+        testContext.UseController<GamesController>();
+        testContext.UseController<PlayerGroupsController>();
+    }
+
+    [Fact]
+    public async Task Cannot_create_resource_without_ID_when_mode_is_required()
```
bkoelman, line 28: This is an integration test (as opposed to a unit test), so it should be functional instead of describing implementation details. The use case is that a user is trying to create a resource without specifying an ID, while the system requires an ID to be provided. That there's an internal "mode" is irrelevant to the user. If we rename that, the scenario does not change.

Therefore I suggested: Cannot_create_resource_without_ID_when_required.
Perhaps better: Cannot_create_resource_without_ID_when_supplying_ID_is_required.

test/OpenApiEndToEndTests/ClientIdGenerationModes/PostTests.cs
```diff
@@ -0,0 +1,165 @@
...
+using Newtonsoft.Json;
+using OpenApiEndToEndTests.ClientIdGenerationModes.GeneratedCode;
+using OpenApiTests;
+using OpenApiTests.ClientIdGenerationModes;
+using TestBuildingBlocks;
+using Xunit;
+
+namespace OpenApiEndToEndTests.ClientIdGenerationModes;
+
+public sealed class PostTests : IClassFixture<IntegrationTestContext<OpenApiStartup<ClientIdGenerationModesDbContext>, ClientIdGenerationModesDbContext>>
```
bkoelman, line 13: ClientIdGenerationModesTests

test/OpenApiEndToEndTests/ClientIdGenerationModes/PostTests.cs
```diff
@@ -0,0 +1,165 @@
...
+                Attributes = new PlayerAttributesInPostRequest
+                {
+                    UserName = player.UserName
+                }
+            }
+        }));
+
+        // Assert
+        PlayerPrimaryResponseDocument? doc = (await action.Should().NotThrowAsync()).Subject;
+        doc.Should().BeNull();
```
bkoelman, line 79: We generally don't assert that something doesn't throw, because anything could throw at any time, which would make the test fail already.

We normally assert on the response (status code and body, and occasionally on headers if relevant) and what's changed in the database. So this should assert on the returned document (don't abbreviate), and verify the database was updated (see `CreateResourceWithClientGeneratedIdTests` for examples).

test/OpenApiTests/ClientIdGenerationModes/ClientGeneratedIdTests.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        testContext.UseController<PlayersController>();
+        testContext.UseController<GamesController>();
+        testContext.UseController<PlayerGroupsController>();
+
+        testContext.SwaggerDocumentOutputDirectory = "test/OpenApiEndToEndTests/ClientIdGenerationModes";
+    }
+
+    [Fact]
+    public async Task Post_data_should_have_required_id()
```
bkoelman, line 24: Suggestion: Schema_property_for_ID_is_required_in_post_request (based on names used in `NullabilityTests`)

test/OpenApiTests/ClientIdGenerationModes/ClientGeneratedIdTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                requiredElement.Should().ContainArrayElement("id");
+            });
+
+            dataElement.Should().ContainPath("properties.id");
+        });
+    }
+
+    [Fact]
+    public async Task Post_data_should_have_non_required_id()
```
bkoelman, line 42: Suggestion: Schema_property_for_ID_is_optional_in_post_request

test/OpenApiTests/ClientIdGenerationModes/ClientGeneratedIdTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                requiredElement.Should().NotContainArrayElement("id");
+            });
+
+            dataElement.Should().ContainPath("properties.id");
+        });
+    }
+
+    [Fact]
+    public async Task Post_data_should_not_have_id()
```
bkoelman, line 60: Suggestion: Schema_property_for_ID_is_omitted_in_post_request

test/OpenApiEndToEndTests/ClientIdGenerationModes/swagger.g.json
```diff
@@ -0,0 +1,4630 @@
+{
```
bkoelman, line 1: This file can be shortened a lot by adding:
```
GenerateControllerEndpoints = JsonApiEndpoints.Post
```
to the `[Resource]` attribute on models. I don't expect we'll ever need to assert on other endpoints here.

bkoelman (changes requested):
See comments.

test/OpenApiEndToEndTests/ClientGeneratedId/PostTests.cs
maurei: Read into it a little bit. It sounds familiar for sure, but it doesn't _really_ ring a concrete bell at this moment... 🤷

bkoelman: Closing, tracked at https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/1485.

test/OpenApiEndToEndTests/ClientIdGenerationModes/ClientIdGenerationModesTests.cs
```diff
@@ -0,0 +1,191 @@
...
+                Attributes = new GameAttributesInPostRequest
+                {
+                    Title = game.Title,
+                    PurchasePrice = (double)game.PurchasePrice
+                }
+            }
+        }));
+
+        // Assert
+        doc?.Data.Id.Should().NotBeNullOrEmpty();
```
bkoelman, line 113: This assertion is conditional, ie when null is returned, nothing is asserted on.
Instead, first assert the response isn't null using `.ShouldNotBeNull()`. This extension method is a workaround for https://github.com/fluentassertions/fluentassertions/issues/1115.

bkoelman (changes requested):
Few minor comments.

bkoelman (approved):
LGTM

If you don't mind, I'd like to merge Kiota support first, so Kiota tests can be added to this PR before merging it.

verdie-g:
Thanks! I couldn't find the time to do the Kiota part.

Issue #1470, opened 2024-02-18
OpenAPI: Required method parameters become nullable with /GenerateOptionalParameters:true


bkoelman:
When adding the `/GenerateOptionalParameters:true` option in NSwag generated clients, the request body parameter becomes nullable. For example:
```c#
public virtual async Task<ApiResponse<CountryPrimaryResponseDocument>> PatchCountryAsync(string id,
    IDictionary<string, string?>? query = null, CountryPatchRequestDocument? body = null,
    CancellationToken cancellationToken = default(CancellationToken))
```
This is incorrect because it always fails with HTTP 422.

The OAS for this method looks correct, as far as I'm aware:
```json
"requestBody": {
  "description": "The attributes and relationships of the country to update. Omitted fields are left unchanged.",
  "content": {
    "application/vnd.api+json": {
      "schema": {
        "allOf": [
          {
            "$ref": "#/components/schemas/countryPatchRequestDocument"
          }
        ]
      }
    }
  }
}
```

I've opened this issue to investigate why this is happening.

bkoelman:
Closing, this was fixed in https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/1507.

PR #1477, opened 2024-02-24
openapi: support validation attributes

### LINKED ISSUE #1055: Reflect model state validation attributes in resource object schemas
The usage of `RequiredAttribute` on a resource field translates to a property needing to be listed as `required`, see [required properties](https://swagger.io/docs/specification/data-models/data-types/). This is currently already supported in the OpenApi integration, but requires more thorough testing.

Further investigation is required to see whether other model state validation attributes are associated to similar options in an OAS.

verdie-g:
Closes #1055

image

| .NET | Applicable types | OpenAPI | NSwag | Kiota | MVC | EF Core | Notes |
|---|---|---|---|---|---|---|---|
| `[Length]` | `string`, `ICollection` | `minLength`/`maxLength`/`minItems`/`maxItems` | Y | Y | Y | N | .NET 8+ |
| `[MinLength]` | `string`, `ICollection` | `minLength`/`minItems` | Y | Y | Y | N | .NET 8+ |
| `[MaxLength]` | `string`, `ICollection` | `maxLength`/`maxItems` | Y | Y | Y | N | .NET 8+ |
| `[Compare]` | any | - | Y | Y | Y | N | Works on any type that implements `Equals` |
| `[Required(AllowEmptyStrings = true)]` | `string` | `required: []` | Y | Y | Y | N | Only applies to JSON:API create resource |
| `[RegularExpression]` | `string` | `pattern: XX` | Y | Y | Y | N | Potential differences in regex dialects, don't care. `Convert.ToString` is used on the value so could theoretically be used on other types than `string` |
| `[StringLength]` | `string` | `minLength` / `maxLength` | Y | Y | Y | Y | Omitting `MinimumLength` implies `0`. `character varying` with PostgreSQL |
| `[CreditCard]` or `[DataType(DataType.CreditCard)]` | `string` | `format: credit-card` | Y | Y | Y | N | |
| `[EmailAddress]` or `[DataType(DataType.EmailAddress)]` | `string` | `format: email` | Y | Y | Y | N | |
| `[Base64String]` | `string` | - | Y | Y | Y | N | .NET 8+ |
| `[Phone]` or `[DataType(DataType.PhoneNumber)]` | `string` | `format: tel` | Y | Y | Y | N | |
| `[Range]` | `int`, `double`, `IComparable` | `minimum`, `maximum` (both inclusive) | Y | Y | Y | N | .NET 8 adds `MinimumIsExclusive`, `MaximumIsExclusive` but it's not supported by Swashbuckle |
| `[Url]` | `string` | `format: uri` | Y | Y | Y | N | Only http/https/ftp, non-relative for NSwag. Kiota generates a string and any absolute or relative uris are accepted |
| `[AllowedValues]` | any | - | Y | Y | Y | N | .NET 8+ |
| `[DeniedValues]` | any | - | Y | Y | Y | N | .NET 8+ |
| `[Required]` | any | `required: []`, `minLength: 1` | Y | Y | Y | Y | Only applies to JSON:API create resource. Makes a db column `NOT NULL`, has priority over `[NotNull]` |
| `[Remote]` | - | - | N | N | Y | N | Not part of System.ComponentModel.DataAnnotations. Ignored by Swashbuckle |
| `[ValidateNever]` | - | - | N | N | Y | N | Not part of System.ComponentModel.DataAnnotations. Ignored by Swashbuckle |
| `[DataType(DataType.Currency)]` | ? | `format: currency` | ? | ? | ? | ? | |
| `[DataType(DataType.Date)]` | ? | `format: date` | ? | ? | ? | ? | |
| `[DataType(DataType.DateTime)]` | ? | `format: date-time` | ? | ? | ? | ? | |
| `[DataType(DataType.Duration)]` | ? | `format: duration` | ? | ? | ? | ? | |
| `[DataType(DataType.Html)]` | ? | `format: html` | ? | ? | ? | ? | |
| `[DataType(DataType.ImageUrl)]` | ? | `format: uri` | ? | ? | ? | ? | |
| `[DataType(DataType.MultilineText)]` | ? | `format: multiline` | ? | ? | ? | ? | |
| `[DataType(DataType.Password)]` | ? | `format: password` | ? | ? | ? | ? | |
| `[DataType(DataType.PostalCode)]` | ? | `format: postal-code` | ? | ? | ? | ? | |
| `[DataType(DataType.Time)]` | ? | `format: time` | ? | ? | ? | ? | |
| `[DataType(DataType.Upload)]` | ? | `format: binary` | ? | ? | ? | ? | |
`[DataType(DataType.Url)]` | ? | `format: uri` | ? | ? | ? | ? | |
| `Guid` | - | `format: uuid` | Y | Y | Y | Y | |
| `HashSet` | - | `uniqueItems` | Y | ? | Y | N | |
| `Uri` | - | `format: uri` | Y | N | Y | Y | Kiota generates a string and any absolute or relative uris are accepted. No DB constraint |
| `DateOnly` | - | `format: date` | Y | Y | Y | Y | Kiota generates a Microsoft.Kiota.Abstractions.Date |
| `TimeOnly` | - | `format: time` | Y | Y | Y | Y | Kiota generates a Microsoft.Kiota.Abstractions.Time |
| - | `multipleOf` | | | | | | |
| - | `hostname` | | | | | | |
| - | `minItems`, `maxItems` | | | | | | |
| - | `ipv4`, `ipv6` | | | | | | |
| `get;` | - | `readOnly` | ? | ? | ? | ? | |
| `set;` | - | `writeOnly` | ? | ? | ? | ? | |

Legend:
- NSwag/Kiota means a working C# client is generated (with test coverage for valid/invalid input, part of this PR).
- MVC means the expected server-side validation error is produced. Sometimes NSwag already catches bad input, so it never hits the server. We don't need to add tests for that in this PR, but it should be tried out. For example, if adding [StringLength] does NOT validate server-side, but only NSwag catches it, we should document that. Because only client-side validation is insecure.
- EF Core means checking what SQL data type and SQL constraints are being generated, and whether they make sense. For example, does [Required(AllowEmptyStrings = false)] generate just a non-nullable column, or does it also add the minimum length constraint? If not, should it do so, if possible at all? Google for it, are there open issues, StackOverflow questions, etc? We need to understand how it works and is supposed to be used, and reason about the feature from an API developer perspective. Is there anything to watch out for? Similar to the previous bullet, this is exploration.

#### QUALITY CHECKLIST
- [x] Adapted tests
- [ ] Documentation updated


test/OpenApiTests/ModelValidation/ModelValidationTests.cs
```diff
@@ -0,0 +1,256 @@
...
+        JsonElement document = await _testContext.GetSwaggerDocumentAsync();
+
+        // Assert
+        document.Should().ContainPath($"components.schemas.{modelName}.properties.tags").With(tagsEl =>
+        {
+            tagsEl.Should().HaveProperty("type", "array");
+            tagsEl.Should().ContainPath("items").With(itemsEl =>
+            {
+                itemsEl.Should().HaveProperty("type", "string");
+                // TODO: no length constraint?
```
verdie-g, line 149: It seems like the library used to generate the OpenAPI document doesn't support `Length` and `Base64String` which were both introduced in .NET 8.

test/OpenApiTests/ModelValidation/ModelValidationTests.cs
```diff
@@ -0,0 +1,256 @@
...
+    [MemberData(nameof(ModelNames))]
+    public async Task Next_revalidation(string modelName)
+    {
+        // Act
+        JsonElement document = await _testContext.GetSwaggerDocumentAsync();
+
+        // Assert
+        document.Should().ContainPath($"components.schemas.{modelName}.properties.nextRevalidation").With(nextRevalidationEl =>
+        {
+            // TODO: TimeSpan format is an akward object with all the TimeSpan public properties.
```
verdie-g, line 180: What is the user expected to do with user to avoid the default TimeSpan JSON serialization?

test/OpenApiTests/ModelValidation/ModelValidationTests.cs
```diff
@@ -0,0 +1,256 @@
...
+        _testContext = testContext;
+
+        testContext.UseController<FingerprintsController>();
+
+        testContext.SwaggerDocumentOutputDirectory = "test/OpenApiEndToEndTests/ModelValidation";
+    }
+
+    [Theory]
+    [MemberData(nameof(ModelNames))]
+    public async Task String(string modelName)
```
verdie-g, line 22: Any opinion on the test names?

verdie-g:
@bkoelman could you give a first review on just that?

test/OpenApiTests/ModelValidation/ModelValidationTests.cs
bkoelman: Swashbuckle's type mappings are defined [here](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/8f363f7359cb1cb8fa5de5195ec6d97aefaa16b3/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs#L189-L215). ModelState attributes are mapped [here](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/8f363f7359cb1cb8fa5de5195ec6d97aefaa16b3/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs#L91-L108).

I suppose we could fill the gap ourselves, as Swashbuckle isn't maintained anymore. Or just postpone adding support for the .NET 8 annotations. Either way, `[StringLength(10)]` can still be used on both frameworks.

It all depends on what .NET 8 adds and how useful that is, how much effort it takes to implement, and how messy the codebase becomes when doing so.

We _could_ adapt the write logic to emit a swagger.g.json per target framework (not saying we _should_, but something to consider).

bkoelman: It's also unclear at this point if our OpenAPI support is going to support .NET 6, which EOLs in less than a year.

test/OpenApiTests/ModelValidation/ModelValidationTests.cs
bkoelman: I'm not sure I understand, can you clarify?

test/OpenApiTests/ModelValidation/ModelValidationTests.cs
bkoelman: How about this?

Phone_annotation_on_resource_property_produces_expected_schema
EmailAddress_annotation_on_resource_property_produces_expected_schema
RegularExpression_annotation_on_resource_property_produces_expected_schema
StringLength_annotation_on_resource_property_produces_expected_schema
etc...

test/OpenApiTests/ModelValidation/ModelValidationTests.cs
verdie-g: The default TimeSpan serialization is
```json
"timeSpan": {
        "type": "object",
        "properties": {
          "ticks": {
            "type": "integer",
            "format": "int64"
          },
          "days": {
            "type": "integer",
            "format": "int32",
            "readOnly": true
          },
          "hours": {
            "type": "integer",
            "format": "int32",
            "readOnly": true
          },
          "milliseconds": {
            "type": "integer",
            "format": "int32",
            "readOnly": true
          },
          "microseconds": {
            "type": "integer",
            "format": "int32",
            "readOnly": true
          },
...
```
So I was just wondering, should I validate that the type is a reference to this object?

It's so sad that there is no good standard to represent a duration as a string (except the horrible `P3Y6M4DT12H30M5S` in ISO 8601).

bkoelman: Can you google a bit? You're surely not the first person wondering about this.

verdie-g: Yeah it's a known issue. I was just wondering if you had a preference about it.

A custom converter can be used to do a TimeSpan.ToString() or TimeSpan.TotalMilliseconds.

bkoelman:
@verdie-g Do you still intend to complete this?

verdie-g:
I pushed some old changes that were still on my machine in case someone wants to continue the PR.

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
```diff
@@ -0,0 +1,450 @@
...
+                    UserName = fingerprint.UserName,
+                    CreditCard = fingerprint.CreditCard,
+                    Email = fingerprint.Email,
+                    Phone = fingerprint.Phone,
+                    Age = fingerprint.Age,
+                    ProfilePicture = fingerprint.ProfilePicture,
+                    BackgroundPicture = new Uri(fingerprint.BackgroundPicture!),
+                    NextRevalidation = "02:00:00",
+                    ValidatedAt = fingerprint.ValidatedAt!.Value.ToUniversalTime(),
+                    // TODO: ValidatedDateAt = new DateTimeOffset(fingerprint.ValidatedDateAt!.Value.ToDateTime(new TimeOnly()).ToUniversalTime()),
```
verdie-g, line 441: If uncommented, fails with
```
 "JsonApiDotNetCore.Repositories.DataStoreUpdateException: Failed to persist changes in the underlying data store.",
          " ---> Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while saving the entity changes. See the inner exception for details.",
          " ---> System.ArgumentException: Cannot write DateTime with Kind=Local to PostgreSQL type 'timestamp with time zone', only UTC is supported. Note that it's not possible to mix DateTimes with different Kinds in an array, range, or multirange. (Parameter 'value')",
          "   at PgConverterResolution? Npgsql.Internal.Converters.DateTimeConverterResolver<T>.Get(DateTime value, PgTypeId? expectedPgTypeId, bool validateOnly)",
          "   at DateTimeConverterResolver<DateTime> Npgsql.Internal.Converters.DateTimeConverterResolver.CreateResolver(PgSerializerOptions options, PgTypeId timestampTz, PgTypeId timestamp, bool dateTimeInfinityConversions)+(DateTimeConverterResolver<DateTime> resolver, DateTime value, PgTypeId? expectedPgTypeId) => { }",
          "   at PgConverterResolution? Npgsql.Internal.Converters.DateTimeConverterResolver<T>.Get(T value, PgTypeId? expectedPgTypeId)",
          "   at PgConverterResolution? Npgsql.Internal.PgConverterResolver<T>.GetAsObjectInternal(PgTypeInfo typeInfo, object value, PgTypeId? expectedPgTypeId)",
          "   at PgConverterResolution? Npgsql.Internal.PgResolverTypeInfo.GetResolutionAsObject(object value, PgTypeId? expectedPgTypeId)",
          "   at PgConverterResolution Npgsql.Internal.PgTypeInfo.GetObjectResolution(object value)",
          "   at PgConverterResolution Npgsql.NpgsqlParameter.ResolveConverter(PgTypeInfo typeInfo)",
          "   at void Npgsql.NpgsqlParameter.ResolveTypeInfo(PgSerializerOptions options)",
          "   at void Npgsql.NpgsqlParameterCollection.ProcessParameters(PgSerializerOptions options, bool validateValues, CommandType commandType)",
          "   at async ValueTask<NpgsqlDataReader> Npgsql.NpgsqlCommand.ExecuteReader(bool async, CommandBehavior behavior, CancellationToken cancellationToken) x 2",
          "   at async Task<DbDataReader> Npgsql.NpgsqlCommand.ExecuteDbDataReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)",
          "   at async Task<RelationalDataReader> Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken) x 2",
          "   at async Task Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.ExecuteAsync(IRelationalConnection connection, CancellationToken cancellationToken)",
          "   --- End of inner exception stack trace ---",
          "   at async Task Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.ExecuteAsync(IRelationalConnection connection, CancellationToken cancellationToken)",
          "   at async Task<int> Microsoft.EntityFrameworkCore.Update.Internal.BatchExecutor.ExecuteAsync(IEnumerable<ModificationCommandBatch> commandBatches, IRelationalConnection connection, CancellationToken cancellationToken) x 3",
          "   at async Task<int> Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChangesAsync(IList<IUpdateEntry> entriesToSave, CancellationToken cancellationToken) x 2",
          "   at async Task<TResult> Npgsql.EntityFrameworkCore.PostgreSQL.Storage.Internal.NpgsqlExecutionStrategy.ExecuteAsync<TState, TResult>(TState state, Func<DbContext, TState, CancellationToken, Task<TResult>> operation, Func<DbContext, TState, CancellationToken, Task<ExecutionResult<TResult>>> verifySucceeded, CancellationToken cancellationToken)",
          "   at async Task<int> Microsoft.EntityFrameworkCore.DbContext.SaveChangesAsync(bool acceptAllChangesOnSuccess, CancellationToken cancellationToken) x 2",
          "   at async Task JsonApiDotNetCore.Repositories.EntityFrameworkCoreRepository<TResource, TId>.SaveChangesAsync(CancellationToken cancellationToken) in JsonApiDotNetCore/src/JsonApiDotNetCore/Repositories/EntityFrameworkCoreRepository.cs:line 651",
          "   --- End of inner exception stack trace ---",
          "   at async Task JsonApiDotNetCore.Repositories.EntityFrameworkCoreRepository<TResource, TId>.SaveChangesAsync(CancellationToken cancellationToken) in JsonApiDotNetCore/src/JsonApiDotNetCore/Repositories/EntityFrameworkCoreRepository.cs:line 657",
          "   at async Task JsonApiDotNetCore.Repositories.EntityFrameworkCoreRepository<TResource, TId>.CreateAsync(TResource resourceFromRequest, TResource resourceForDatabase, CancellationToken cancellationToken) in JsonApiDotNetCore/src/JsonApiDotNetCore/Repositories/EntityFrameworkCoreRepository.cs:line 232",
          "   at void CallSite.Target(Closure, CallSite, object)",
          "   at void System.Dynamic.UpdateDelegates.UpdateAndExecuteVoid1<T0>(CallSite site, T0 arg0)",
          "   at async Task JsonApiDotNetCore.Repositories.ResourceRepositoryAccessor.CreateAsync<TResource>(TResource resourceFromRequest, TResource resourceForDatabase, CancellationToken cancellationToken)",
          "   at async Task<TResource> JsonApiDotNetCore.Services.JsonApiResourceService<TResource, TId>.CreateAsync(TResource resource, CancellationToken cancellationToken) in JsonApiDotNetCore/src/JsonApiDotNetCore/Services/JsonApiResourceService.cs:line 217",
          "   at async Task<TResource> JsonApiDotNetCore.Services.JsonApiResourceService<TResource, TId>.CreateAsync(TResource resource, CancellationToken cancellationToken) in JsonApiDotNetCore/src/JsonApiDotNetCore/Services/JsonApiResourceService.cs:line 223",
          "   at async Task<IActionResult> JsonApiDotNetCore.Controllers.BaseJsonApiController<TResource, TId>.PostAsync(TResource resource, CancellationToken cancellationToken) in JsonApiDotNetCore/src/JsonApiDotNetCore/Controllers/BaseJsonApiController.cs:line 207",
          "   at async Task<IActionResult> JsonApiDotNetCore.Controllers.JsonApiController<TResource, TId>.PostAsync(TResource resource, CancellationToken cancellationToken) in JsonApiDotNetCore/src/JsonApiDotNetCore/Controllers/JsonApiController.cs:line 78",
          "   at async ValueTask<IActionResult> Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor+TaskOfIActionResultExecutor.Execute(ActionContext actionContext, IActionResultTypeMapper mapper, ObjectMethodExecutor executor, object controller, object[] arguments)",
          "   at async Task Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.InvokeActionMethodAsync()+Logged(?)",
          "   at async Task Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.InvokeNextActionFilterAsync()+Awaited(?)",
          "   at void Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Rethrow(ActionExecutedContextSealed context)",
          "   at Task Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Next(ref State next, ref Scope scope, ref object state, ref bool isCompleted)",
          "   at Task Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.InvokeInnerFilterAsync()",
          "   at async Task Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeNextExceptionFilterAsync()+Awaited(?)"
```

It sounds to me that JADNC pass the DateOnly as a DateTime to Npgsql?

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
```diff
@@ -0,0 +1,450 @@
...
+
+        ErrorObject errorObject = document.Errors.First();
+        errorObject.Title.Should().Be("Input validation failed.");
+        errorObject.Detail.Should().Be("The field NextRevalidation must be between 01:00:00 and 05:00:00.");
+        errorObject.Source.ShouldNotBeNull();
+        errorObject.Source.Pointer.Should().Be("/data/attributes/nextRevalidation");
+    }
+
+    [Fact]
+    public async Task Invalid_datetime_should_return_an_error()
```
verdie-g, line 316: Date, DateTime, and Uri tests all fail because the generated code disallows me to pass an invalid value. Should I bypass the generated client and craft a request using the HttpClient?

verdie-g:
Hi @bkoelman, I'll try finishing that PR. Could you have a pass on its current state, I'll add tests for Kiota once the one for NSwag start looking okay. Also, is there anything worth adding to the doc regarding this change?

bkoelman:
> Hi @bkoelman, I'll try finishing that PR. Could you have a pass on its current state, I'll add tests for Kiota once the one for NSwag start looking okay. Also, is there anything worth adding to the doc regarding this change?

Cool! Sure, I'll take a look. I don't expect the docs need to be updated, unless there are pitfalls users should know about. For kiota, it's worth removing the `--log-level Error` switch temporarily to see what comes up. However, kiota is pretty rigid and inflexible, so don't worry too much about its quirks.

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
bkoelman: Fakers generally contain `.ToUniversalTime()`, so PostgreSQL won't block non-UTC values. Though it's odd you're getting this error on a `DateOnly`. You can check the generated SQL by running a server-side test in debug mode, which writes to the Output window. Perhaps this is a bug in the PostgreSQL EF Core provider?

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
bkoelman: Then it's not an OpenAPI test anymore. If it can't be tested through OpenAPI, just leave them out.

test/OpenApiNSwagEndToEndTests/OpenApiNSwagEndToEndTests.csproj
```diff
@@ -30,6 +30,13 @@
       <CodeGenerator>NSwagCSharp</CodeGenerator>
       <Options>/ClientClassAccessModifier:internal /GenerateExceptionClasses:false /AdditionalNamespaceUsages:JsonApiDotNetCore.OpenApi.Client.NSwag /GenerateNullableReferenceTypes:true</Options>
     </OpenApiReference>
+    <OpenApiReference Include="..\OpenApiTests\ModelValidation\GeneratedSwagger\swagger.g.json">
+      <Namespace>OpenApiNSwagEndToEndTests.ModelValidation.GeneratedCode</Namespace>
+      <ClassName>ModelValidationClient</ClassName>
+      <OutputPath>ModelValidationClient.cs</OutputPath>
+      <CodeGenerator>NSwagCSharp</CodeGenerator>
+      <Options>/ClientClassAccessModifier:internal /GenerateExceptionClasses:false /AdditionalNamespaceUsages:JsonApiDotNetCore.OpenApi.Client.NSwag /GenerateNullableReferenceTypes:true</Options>
```
bkoelman, line 38: #1470 was fixed, so `/GenerateOptionalParameters:true` should work properly now.

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
```diff
@@ -0,0 +1,450 @@
...
+using Newtonsoft.Json;
+using OpenApiNSwagEndToEndTests.ModelValidation.GeneratedCode;
+using OpenApiTests;
+using OpenApiTests.ModelValidation;
+using TestBuildingBlocks;
+using Xunit;
+
+namespace OpenApiNSwagEndToEndTests.ModelValidation;
+
+public sealed class ModelValidationTests : IClassFixture<IntegrationTestContext<OpenApiStartup<ModelValidationDbContext>, ModelValidationDbContext>>
```
bkoelman, line 13: Meanwhile, `XUnitLogHttpMessageHandler` was added to all other tests, which writes the JSON request/response to the xUnit test output window. Please adapt accordingly, I've found it quite helpful in debugging.

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
```diff
@@ -0,0 +1,450 @@
...
+
+    public ModelValidationTests(IntegrationTestContext<OpenApiStartup<ModelValidationDbContext>, ModelValidationDbContext> testContext)
+    {
+        _testContext = testContext;
+
+        testContext.UseController<FingerprintsController>();
+    }
+
+    [Fact]
+    public async Task Omitting_a_required_attribute_should_return_an_error()
```
bkoelman, line 26: We've generally used statement form in test names, ie:
- Cannot_omit_required_attribute (note this applies to POST only, so it would be better to include that in the name: Cannot_create_resource_with_omitted_required_property)
- Cannot_exceed_length_constraint
- Cannot_violate_regular_expression_constraint
- Cannot_use_invalid_email
- Cannot_use_TimeSpan_outside_of_valid_range
- ...
- Can_create_resource_with_valid_properties

test/OpenApiTests/ModelValidation/Fingerprint.cs
```diff
@@ -0,0 +1,60 @@
...
+using JetBrains.Annotations;
+using JsonApiDotNetCore.Controllers;
+using JsonApiDotNetCore.Resources;
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace OpenApiTests.ModelValidation;
+
+[UsedImplicitly(ImplicitUseTargetFlags.Members)]
+[Resource(ControllerNamespace = "OpenApiTests.ModelValidation", GenerateControllerEndpoints = JsonApiEndpoints.Post | JsonApiEndpoints.Patch)]
+public sealed class Fingerprint : Identifiable<Guid>
```
bkoelman, line 11: Perhaps the name `SocialMediaAccount` is more appropriate?

test/OpenApiTests/ModelValidation/Fingerprint.cs
```diff
@@ -0,0 +1,60 @@
...
+    public TimeSpan? NextRevalidation { get; set; }
+
+    [Attr]
+    public DateTime? ValidatedAt { get; set; }
+
+    [Attr]
+    public DateOnly? ValidatedDateAt { get; set; }
+
+    [Attr]
+    public TimeOnly? ValidatedTimeAt { get; set; }
```
bkoelman, line 59: `ValidatedAtDate` / `ValidatedAtTime` sounds more like proper English.

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
```diff
@@ -0,0 +1,450 @@
...
+using FluentAssertions.Specialized;
+using JsonApiDotNetCore.OpenApi.Client.NSwag;
+using Newtonsoft.Json;
+using OpenApiNSwagEndToEndTests.ModelValidation.GeneratedCode;
+using OpenApiTests;
+using OpenApiTests.ModelValidation;
+using TestBuildingBlocks;
+using Xunit;
+
+namespace OpenApiNSwagEndToEndTests.ModelValidation;
```
bkoelman, line 11: Please rename to ModelStateValidation, to match with existing non-OpenAPI tests.

test/OpenApiTests/ModelValidation/ModelValidationTests.cs
```diff
@@ -0,0 +1,255 @@
...
+        JsonElement document = await _testContext.GetSwaggerDocumentAsync();
+
+        // Assert
+        document.Should().ContainPath($"components.schemas.{modelName}.properties.lastName").With(lastNameEl =>
+        {
+            lastNameEl.Should().HaveProperty("minLength", 1);
+            lastNameEl.Should().HaveProperty("type", "string");
+            lastNameEl.Should().NotContainPath("nullable");
+        });
+    }
```
bkoelman, line 49: Nullability and requiredness are extensively covered by the tests in `ResourceFieldValidation`, so I think they can be left out here. But one case to include here is using `[Required(AllowEmptyStrings = false)]`.

test/OpenApiTests/ModelValidation/ModelValidationTests.cs
```diff
@@ -0,0 +1,255 @@
...
+    {
+        // Act
+        JsonElement document = await _testContext.GetSwaggerDocumentAsync();
+
+        // Assert
+        document.Should().ContainPath($"components.schemas.{modelName}.properties.userName").With(userNameEl =>
+        {
+            userNameEl.Should().HaveProperty("pattern", "^[a-zA-Z]+$");
+            userNameEl.Should().HaveProperty("type", "string");
+            userNameEl.Should().HaveProperty("nullable", true);
```
bkoelman, line 80: I wouldn't expect nullability to be of concern in these tests. Am I missing something?

bkoelman (commented):
I've added some comments on things that jumped out to me. Before focussing on kiota, we should get the proposed scope table in better shape.

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
verdie-g: Make sense

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
verdie-g: That's cool
```
Failed executing DbCommand (34ms) [Parameters=[@p0='43debb5f-abe4-4e44-bad5-b852124e8dd4', @p1='95' (Nullable = true), @p2='https://loremflickr.com/320/240?lock=1798481699', @p3='6759-7554-5043-3270', @p4='Kellie15@gmail.com', @p5='Kellie', @p6='Wilderman' (Nullable = false), @p7='02:00:00' (Nullable = true) (DbType = Object), @p8='(546) 274-9304 x65491', @p9='https://loremflickr.com/320/240?lock=475250130', @p10='ogxiroilgkft', @p11='2019-12-30T20:45:27.0196183-05:00' (Nullable = true) (DbType = DateTime), @p12='12/31/2019' (Nullable = true) (DbType = Date), @p13='19:31' (Nullable = true) (DbType = Time)], CommandType='Text', CommandTimeout='30']
      INSERT INTO "Fingerprints" ("Id", "Age", "BackgroundPicture", "CreditCard", "Email", "FirstName", "LastName", "NextRevalidation", "Phone", "ProfilePicture", "UserName", "ValidatedAt", "ValidatedDateAt", "ValidatedTimeAt")
      VALUES (@p0, @p1, @p2, @p3, @p4, @p5, @p6, @p7, @p8, @p9, @p10, @p11, @p12, @p13);
```
Looks fine
```
@p12='12/31/2019' (Nullable = true) (DbType = Date)
```
I'll dig more

test/OpenApiNSwagEndToEndTests/OpenApiNSwagEndToEndTests.csproj
verdie-g: I don't remember exactly what this is about but I understand you would want me to add it to this client

verdie-g: Oh I understand now, it allows me to write
```
apiClient.PostFingerprintAsync(new FingerprintPostRequestDocument
```
instead of
```
apiClient.PostFingerprintAsync(null, new FingerprintPostRequestDocument
```

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
verdie-g: That's indeed pretty neat

test/OpenApiTests/ModelValidation/Fingerprint.cs
verdie-g: Agreed

test/OpenApiTests/ModelValidation/ModelValidationTests.cs
verdie-g: I've added the table to the PR description and will complete it later.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    [Attr]
+    [Url]
+    public string? BackgroundPicture { get; set; }
+
+    [Attr]
+#if NET8_0_OR_GREATER
+    [AllowedValues("tag1", "tag2")]
+    [DeniedValues("tag3")]
+    [Length(0, 10)]
```
verdie-g, line 68: Even if Length is `0, 10`, a POST with `Tags = null` will fail with `The Tags field does not equal any of the values specified in AllowedValuesAttribute.` :(

verdie-g: Oh AllowedValues probably doesn't work on collections

test/OpenApiNSwagEndToEndTests/OpenApiNSwagEndToEndTests.csproj
bkoelman: I'm fine either way. I just mentioned it because you added it initially and it didn't work properly.

test/OpenApiTests/ModelValidation/ModelValidationTests.cs
bkoelman: My mistake: `AllowEmptyStrings` is already false by default. We should test the case with `true`.

test/OpenApiTests/ModelValidation/ModelValidationTests.cs
bkoelman: I've re-pasted the table in the PR description from the above comment's source to preserve formatting. I noticed the `DataType` entries were removed, was that intentional? They _do_ affect the `openapi.json` output when used.

verdie-g: >  I noticed the DataType entries were removed, was that intentional?

I did remove them because of your comment

> so perhaps the [DataType] entries are better left out of scope


I'm not sure how to fill that table. What does "yes" means for NSwag, MVC, and EF Core?

bkoelman: > > I noticed the DataType entries were removed, was that intentional?
> 
> I did remove them because of your comment

The point of the table is to document the outcome of experiments and decisions. For example, if we decide that `[DataType(DataType.Html)]` is out of scope because it doesn't affect validation, the notes column should indicate so. This makes it possible to distinguish between "there's intentionally no test coverage" vs "we simply forgot it".

> > so perhaps the [DataType] entries are better left out of scope
> 
> I'm not sure how to fill that table. What does "yes" means for NSwag, MVC, and EF Core?

That it behaves the way JADNC users would expect, without oddities/workarounds. It indicates if we need to warn users in the docs, log warnings, etc.
- NSwag/Kiota means a working C# client is generated (with test coverage for valid/invalid input, part of this PR).
- MVC means the expected server-side validation error is produced. Sometimes NSwag already catches bad input, so it never hits the server. We don't need to add tests for that in this PR, but it should be tried out. For example, if adding `[StringLength]` does NOT validate server-side, but only NSwag catches it, we should document that. Because only client-side validation is insecure.
- EF Core means checking what SQL data type and SQL constraints are being generated, and whether they make sense. For example, does `[Required(AllowEmptyStrings = false)]` generate just a non-nullable column, or does it also add the minimum length constraint? If not, should it do so, if possible at all? Google for it, are there open issues, StackOverflow questions, etc? We need to understand how it works and is supposed to be used, and reason about the feature from an API developer perspective. Is there anything to watch out for? Similar to the previous bullet, this is exploration.

bkoelman: According to https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.validationattribute?view=net-8.0, there's also `[MinLength]` and `[MaxLength]`. They should be included in the table.

test/OpenApiTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,309 @@
...
+
+    [Theory]
+    [MemberData(nameof(ModelNames))]
+    public async Task HashSet_type_produces_expected_schema(string modelName)
+    {
+        // Act
+        JsonElement document = await _testContext.GetSwaggerDocumentAsync();
+
+        // Assert
+        document.Should().ContainPath($"components.schemas.{modelName}.properties.tags").With(tagsEl =>
```
bkoelman, line 217: Please rename `tagsEl` to `tagsElement`, as well as similar variables in this file.

test/OpenApiTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,309 @@
...
+        // Assert
+        document.Should().ContainPath($"components.schemas.{modelName}.properties.validatedAtTime").With(validatedTimeAtEl =>
+        {
+            validatedTimeAtEl.Should().HaveProperty("type", "string");
+            validatedTimeAtEl.Should().HaveProperty("format", "time");
+        });
+    }
+
+    public static TheoryData<string> ModelNames =>
+        // ReSharper disable once UseCollectionExpression
```
bkoelman, line 302: This looks like a Resharper bug. Is there an open issue at https://youtrack.jetbrains.com/issues/RSRP we can link to? If not, can you create one and link to it from here?

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
```diff
@@ -0,0 +1,94 @@
...
+    public string? CountryCode { get; set; }
+
+    [Attr]
+#if NET8_0_OR_GREATER
+    [DeniedValues("pluto")]
+#endif
+    public string? Planet { get; set; }
+
+    [Attr]
+    [Range(typeof(TimeSpan), "01:00", "05:00")]
```
bkoelman, line 83: Can you check if `TimeSpan` is culture-sensitive? If so, should add `ConvertValueInInvariantCulture = true, ParseLimitsInInvariantCulture = true`.

test/OpenApiTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,309 @@
...
+
+        // Assert
+        document.Should().ContainPath($"components.schemas.{modelName}.properties.validatedAtTime").With(validatedTimeAtEl =>
+        {
+            validatedTimeAtEl.Should().HaveProperty("type", "string");
+            validatedTimeAtEl.Should().HaveProperty("format", "time");
+        });
+    }
+
+    public static TheoryData<string> ModelNames =>
```
bkoelman, line 301:
```suggestion
    public static readonly TheoryData<string> ModelNames =
```

bkoelman (commented):
See comments. The test coverage needs adjustment, based on the table (that isn't finished yet). I would expect to see some net80-only tests too.

test/OpenApiTests/ModelStateValidation/GeneratedSwagger/net6.0/swagger.g.json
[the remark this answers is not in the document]
verdie-g: I think the .NET 8 attributes are currently not supported by OpenAPI.

test/OpenApiTests/ModelStateValidation/ModelStateValidationFakers.cs
```diff
@@ -0,0 +1,39 @@
...
+[UsedImplicitly(ImplicitUseTargetFlags.Members)]
+public sealed class ModelStateValidationFakers
+{
+    private readonly Lazy<Faker<SocialMediaAccount>> _lazySocialMediaAccountFaker = new(() => new Faker<SocialMediaAccount>()
+        .MakeDeterministic()
+        .RuleFor(socialMediaAccount => socialMediaAccount.AlternativeId, faker => faker.Random.Guid())
+        .RuleFor(socialMediaAccount => socialMediaAccount.FirstName, faker => faker.Person.FirstName)
+        .RuleFor(socialMediaAccount => socialMediaAccount.GivenName, (_, socialMediaAccount) => socialMediaAccount.FirstName)
+        .RuleFor(socialMediaAccount => socialMediaAccount.LastName, faker => faker.Person.LastName)
+        .RuleFor(socialMediaAccount => socialMediaAccount.UserName, faker => faker.Random.String2(3, 18))
```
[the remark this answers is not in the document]
verdie-g: I missed the `Random.UserName`

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
verdie-g: It seems like it only accepts invariant culture. I added a test to check that a comma is not allowed.

test/OpenApiTests/ModelStateValidation/ModelStateValidationFakers.cs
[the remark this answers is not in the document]
verdie-g: Ah never mind. I used `String2` because I added a only-letters constraint on UserName.

test/OpenApiNSwagEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,555 @@
...
+    public async Task Cannot_violate_compare_constraint()
+    {
+        // Arrange
+        SocialMediaAccount socialMediaAccount = _fakers.SocialMediaAccount.Generate();
+
+        using HttpClient httpClient = _testContext.Factory.CreateDefaultClient(_logHttpMessageHandler);
+        ModelStateValidationClient apiClient = new(httpClient);
+
+        // Act
+        SocialMediaAccountPostRequestDocument requestBody = new()
```
bkoelman, line 37: This doesn't look right. Building the request body is not part of the act phase.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
bkoelman: I believe it is culture sensitive. According to https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-timespan-format-strings, the decimal separator is used for milliseconds.

test/OpenApiTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,308 @@
...
+    public async Task Compare_annotation_on_resource_property_produces_expected_schema(string modelName)
+    {
+        // Act
+        JsonElement document = await _testContext.GetSwaggerDocumentAsync();
+
+        // Assert
+        document.Should().ContainPath($"components.schemas.{modelName}.properties.givenName").With(surnameElement =>
+        {
+            surnameElement.Should().HaveProperty("type", "string");
+        });
```
bkoelman, line 52:
```suggestion
        document.Should().ContainPath($"components.schemas.{modelName}.properties.givenName").With(givenNameElement =>
        {
            givenNameElement.Should().HaveProperty("type", "string");
        });
```

test/OpenApiTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,309 @@
...
+        _testContext = testContext;
+
+        testContext.UseController<SocialMediaAccountsController>();
+
+        const string targetFramework =
+#if NET8_0_OR_GREATER
+            "net8.0";
+#else
+            "net6.0";
+#endif
```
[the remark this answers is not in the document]
bkoelman: This was updated in this file, but it occurs in multiple places. Please adapt accordingly.

test/OpenApiTests/ModelStateValidation/ModelStateValidationFakers.cs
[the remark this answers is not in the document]
bkoelman: Ok, that makes sense. But it doesn't apply to the other places where `String2` is used, right?

test/OpenApiTests/ModelStateValidation/GeneratedSwagger/net6.0/swagger.g.json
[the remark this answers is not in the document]
bkoelman: At least `Length` _is_ supported, see https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2882. I haven't checked the other ones.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
verdie-g: Ah maybe my computer is using US culture. I'll try setting the culture explicitly in the program.

test/OpenApiNSwagEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,555 @@
...
+        errorObject.Detail.Should().Be("Failed to convert attribute 'validatedAtTime' with value '-00:00:01' of type 'String' to type 'Nullable<TimeOnly>'.");
+        errorObject.Source.ShouldNotBeNull();
+        errorObject.Source.Pointer.Should().Be("/data/attributes/validatedAtTime");
+    }
+
+    [Fact]
+    public async Task Can_create_resource_with_valid_properties()
+    {
+        // Arrange
+        SocialMediaAccount socialMediaAccount = _fakers.SocialMediaAccount.Generate();
```
bkoelman, line 519: The tests in the codebase use the convention to prefix variables that represent unstored data with `new`, and existing data (at the time of _act_) with `existing`. For example, see https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/test/JsonApiDotNetCoreTests/IntegrationTests/ReadWrite/Updating/Resources/UpdateResourceTests.cs#L86-L110. I'd prefer to apply the same convention here for consistency.

Please rename to `newAccount` here, and use `newLastName` in the other tests.

test/OpenApiTests/ModelStateValidation/ModelStateValidationFakers.cs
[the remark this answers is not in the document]
verdie-g: For `Planet` and `Tags` I could not find a better alternative

test/OpenApiNSwagEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
verdie-g: > and use newLastName in the other tests.

I assume something like this `string newLastName = _fakers.SocialMediaAccount.Generate().LastName;`

I think I would prefer keeping `SocialMediaAccount newAccount = _fakers.SocialMediaAccount.Generate();` in case another property from `newAccount` is needed later.

test/OpenApiTests/ModelStateValidation/GeneratedSwagger/net6.0/swagger.g.json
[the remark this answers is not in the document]
verdie-g: Even after setting `SwashbuckleVersion` to 6.6.2, no special fields seem to be set on `firstName`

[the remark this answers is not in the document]
verdie-g: Oh nvm I missed the `SwashbuckleFrozenVersion` that was set to 6.6.1

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
verdie-g: My bad, it's `ValidatedAt` that causes some issues. I was able to reproduce the issue
```csharp
var builder = WebApplication.CreateBuilder(args);
builder.Services.AddDbContext<AppDbContext>(options => options.UseNpgsql("Host=localhost:5432;Database=test;Username=postgres;Password=root"));
builder.Services.AddJsonApi<AppDbContext>();
var app = builder.Build();

using (var scope = app.Services.CreateScope())
{
    var db = scope.ServiceProvider.GetRequiredService<AppDbContext>();
    db.Database.EnsureCreated();
}

app.UseRouting();
app.UseJsonApi();
app.MapControllers();

app.Run();

[Resource]
public class Author : Identifiable<Guid>
{
    [Attr]
    public DateTime Date { get; set; }
}

public class AppDbContext : DbContext
{
    public DbSet<Author> Authors => Set<Author>();

    public AppDbContext(DbContextOptions<AppDbContext> options)
        : base(options)
    {
    }
}
```

```
curl -X POST http://localhost:18574/authors -H "Content-Type: application/vnd.api+json" --data '{"data":{"type":"authors","attributes":{"date": "2019-12-31T01:23:25.2607037+00:00"}}}'
```

After some debugging I found out that a date like `2019-12-31T01:23:25.2607037+00:00` will be converted by ASP.NET to `30-Dec-19 20:23:25 (Kind=Local)`. However, `2019-12-31T01:23:25.2607037Z` is converted to `31-Dec-19 01:23:25 (Kind=Utc)`.

The reason why a timezone offset is used in the test is that NSwag generates a `DateTimeOffset` for `DateTime` attributes.

So this doesn't seem to bug. It's super awkward for the user to get a 500 for that though. Also, I'm not sure how to fix the test. One work-around is to replace the `DateTime` with `DateTimeOffset` in the resource definition.

verdie-g:
I've completed the table except the EF Core column which I think is out-of-scope for this PR.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
verdie-g: I was a bit confused about what these properties do so I copied the `RangeAttribute` code to play with it
```csharp
CultureInfo.CurrentCulture = new CultureInfo("fr-FR");
RangeAttribute2 a1 = new(typeof(TimeSpan), "01:00", "05:00");
RangeAttribute2 a2 = new(typeof(TimeSpan), "01:00", "05:00")
{
    ConvertValueInInvariantCulture = true,
    ParseLimitsInInvariantCulture = true,
};
foreach (string s in new[] { "02:00:00", "02:00:00.123", "02:00:00,123" })
{
    Console.WriteLine($"{s} -> {a1.IsValid(s)}");
    Console.WriteLine($"{s} -> {a2.IsValid(s)}");
}
```
this prints
```
02:00:00 -> True
02:00:00 -> True
02:00:00.123 -> True
02:00:00.123 -> True
02:00:00,123 -> True
02:00:00,123 -> False
```

`ParseLimitsInInvariantCulture` is to parse `01:00` and `05:00` as invariant. In our case since they are constants, it's not super useful. Though, `ConvertValueInInvariantCulture` means that the input will be parsed using InvariantCulture. This one is important. My computer uses US culture so I could not see the difference.

verdie-g:
> For kiota, it's worth removing the --log-level Error switch temporarily to see what comes up.
```
#/paths/~1socialMediaAccounts/post/responses/201/headers/Location/schema - The format uri is not supported by Kiota and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPatchRequest/properties/creditCard - The format credit-card is not supported by Kiota for the type string and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPatchRequest/properties/email - The format email is not supported by Kiota and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPatchRequest/properties/phone - The format tel is not supported by Kiota for the type string and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPatchRequest/properties/profilePicture - The format uri is not supported by Kiota and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPatchRequest/properties/backgroundPicture - The format uri is not supported by Kiota and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPatchRequest/properties/nextRevalidation - The format date-span is not supported by Kiota for the type string and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPostRequest/properties/creditCard - The format credit-card is not supported by Kiota for the type string and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPostRequest/properties/email - The format email is not supported by Kiota and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPostRequest/properties/phone - The format tel is not supported by Kiota for the type string and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPostRequest/properties/profilePicture - The format uri is not supported by Kiota and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPostRequest/properties/backgroundPicture - The format uri is not supported by Kiota and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInPostRequest/properties/nextRevalidation - The format date-span is not supported by Kiota for the type string and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInResponse/properties/creditCard - The format credit-card is not supported by Kiota for the type string and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInResponse/properties/email - The format email is not supported by Kiota and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInResponse/properties/phone - The format tel is not supported by Kiota for the type string and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInResponse/properties/profilePicture - The format uri is not supported by Kiota and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInResponse/properties/backgroundPicture - The format uri is not supported by Kiota and the string type will be used.
#/components/schemas/socialMediaAccountAttributesInResponse/properties/nextRevalidation - The format date-span is not supported by Kiota for the type string and the string type will be used.
```

Kiota doesn't support the formats `uri` and `date-span` :/

test/OpenApiTests/ModelStateValidation/ModelStateValidationFakers.cs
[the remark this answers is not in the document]
bkoelman: Oh, I see now. It's because a string of a certain length is needed. And .`String()` potentially generates malformed UTF-16.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
bkoelman: Yes, that's how it works. Thanks for looking into it.

The `ConvertValueInInvariantCulture = true` on `Age` should be removed, because it's meaningless. See the documentation at https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.rangeattribute.convertvalueininvariantculture?view=net-8.0:

> Gets or sets a value that determines whether any conversions of the value being validated to [OperandType](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.rangeattribute.operandtype?view=net-8.0#system-componentmodel-dataannotations-rangeattribute-operandtype) as set by the type parameter of the [RangeAttribute(Type, String, String)](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.rangeattribute.-ctor?view=net-8.0#system-componentmodel-dataannotations-rangeattribute-ctor(system-type-system-string-system-string)) constructor use the invariant culture or the current culture.
> 
> This property has no effect on the constructors with [Int32](https://learn.microsoft.com/en-us/dotnet/api/system.int32?view=net-8.0) or [Double](https://learn.microsoft.com/en-us/dotnet/api/system.double?view=net-8.0) parameters, which always use the invariant culture for any conversions of the validated value.

test/OpenApiNSwagEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,560 @@
...
+                {
+                    FirstName = newAccount.FirstName,
+                    GivenName = "something else",
+                    LastName = newAccount.LastName
+                }
+            }
+        };
+
+        // Act
+        Func<Task<SocialMediaAccountPrimaryResponseDocument>> action = () => apiClient.PostSocialMediaAccountAsync(requestBody);
```
bkoelman, line 51: This can be shortened to:
```c#
Func<Task> action = async() => await apiClient.PostSocialMediaAccountAsync(requestBody);
```

test/OpenApiNSwagEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
bkoelman: Sounds good. I'm seeing failed tests because `Age` is not in a valid range.

test/OpenApiNSwagEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,560 @@
...
+                    Email = newAccount.Email,
+                    Phone = newAccount.Phone,
+                    Age = newAccount.Age,
+                    ProfilePicture = newAccount.ProfilePicture,
+                    BackgroundPicture = new Uri(newAccount.BackgroundPicture!),
+                    Tags = newAccount.Tags,
+                    Planet = newAccount.Planet,
+                    NextRevalidation = newAccount.NextRevalidation!.Value.ToString(),
+                    ValidatedAt = newAccount.ValidatedAt!,
+                    ValidatedAtDate = new DateTimeOffset(newAccount.ValidatedAtDate!.Value.ToDateTime(new TimeOnly()).ToUniversalTime()),
```
bkoelman, line 548: This looks ugly. Can it be addressed more elegantly? See tips at https://github.com/RicoSuter/NSwag/issues/4062. A similar approach possibly also solves the `ValidatedAt` UTC issue.

This is why we write tests: to see if/how we can provide a smooth experience to our users, instead of just hacking around the quirks we encounter.

bkoelman (commented):
I'm not seeing any end-to-end tests for the .NET 8 attributes.

Oh well, I expected to find `#if` lines, but end-to-end tests can't multi-target so it's .NET 8 only.

bkoelman:
> I've completed the table except the EF Core column which I think is out-of-scope for this PR.

I've mentioned earlier that the goal of writing end-to-end tests is to assess the overall experience our users will face. This doesn't mean we need to write tests for EF Core features in this PR, but we do need to check for oddities and possibly create separate issues to address them, or document things to watch out for. So I disagree that it's out of scope. Otherwise, I wouldn't have added the EF Core column in the first place.

test/OpenApiTests/ModelStateValidation/ModelStateValidationFakers.cs
```diff
@@ -0,0 +1,39 @@
...
+        .RuleFor(socialMediaAccount => socialMediaAccount.ProfilePicture, faker => new Uri(faker.Image.LoremFlickrUrl()))
+        .RuleFor(socialMediaAccount => socialMediaAccount.BackgroundPicture, faker => faker.Image.LoremFlickrUrl())
+        .RuleFor(socialMediaAccount => socialMediaAccount.Tags, faker => faker.Make(faker.Random.Number(1, 10), () => faker.Random.String2(2, 10)))
+        .RuleFor(socialMediaAccount => socialMediaAccount.CountryCode, faker => faker.Random.ListItem([
+            "NL",
+            "FR"
+        ]))
+        .RuleFor(socialMediaAccount => socialMediaAccount.Planet, faker => faker.Random.String2(2, 8))
+        .RuleFor(socialMediaAccount => socialMediaAccount.NextRevalidation, faker => TimeSpan.FromHours(faker.Random.Number(1, 5)))
+        .RuleFor(socialMediaAccount => socialMediaAccount.ValidatedAt, faker => faker.Date.Recent().ToUniversalTime())
```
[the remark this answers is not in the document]
verdie-g: > Is there a way for the client to send with "Z" suffix instead?

Yeah I mentioned that issue there https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/1477#discussion_r1632492916 :/ I didn't find a good solution yet except adding a custom json converter.

test/OpenApiNSwagEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
verdie-g: The ugliness comes from the fact that there is no direct conversion from DateOnly to DateTimeOffset. I'm not sure a custom serializer will help here.

verdie-g:
Here is SQL statements to create the table
```sql
CREATE TABLE "SocialMediaAccounts" (
    "Id" uuid NOT NULL,
    "AlternativeId" uuid,
    "FirstName" text,
    "LastName" text NOT NULL,
    "UserName" character varying(18),
    "CreditCard" text,
    "Email" text,
    "Password" text,
    "Phone" text,
    "Age" double precision,
    "ProfilePicture" text,
    "BackgroundPicture" text,
    "Tags" text[],
    "CountryCode" text,
    "Planet" text,
    "NextRevalidation" interval,
    "ValidatedAt" timestamp with time zone,
    "ValidatedAtDate" date,
    "ValidatedAtTime" time without time zone,
    CONSTRAINT "PK_SocialMediaAccounts" PRIMARY KEY ("Id")
);
```
I've updated the table accordingly.

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
verdie-g: I added custom JSON converter for DateTime to always convert them to UTC. The code copied between the two tests projects. It could probably be shared in another assembly.

test/OpenApiNSwagEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
bkoelman: I don't see this change was applied. Did you push it?

test/OpenApiTests/ModelStateValidation/ModelStateValidationFakers.cs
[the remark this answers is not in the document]
bkoelman: The custom converter is a nice solution.

test/OpenApiTests/ModelStateValidation/ModelStateValidationFakers.cs
```diff
@@ -0,0 +1,39 @@
...
+        .RuleFor(socialMediaAccount => socialMediaAccount.ValidatedAtDate, faker => DateOnly.FromDateTime(faker.Date.Recent()))
+        .RuleFor(socialMediaAccount => socialMediaAccount.ValidatedAtTime, faker => TimeOnly.FromDateTime(faker.Date.Recent())));
```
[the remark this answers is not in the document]
bkoelman: I don't see this changed. Did you know you can batch-up suggestions and commit them in one go? If you take the suggestion, it auto-resolves the conversation.

test/OpenApiNSwagEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
bkoelman: Probably not, let's just leave it like this then.

bkoelman (commented):
Great work on filling out the feature table.

Please update the following entries, replacing the '?':
- | [ValidateNever] | - | N | N | N |
- | [Remote] | - | N | N | N |

Also, `readOnly` and `writeOnly` need to move one column to the right. I've tried adding properties for them, but set-only causes a crash inside `Microsoft.AspNetCore.Mvc.ModelBinding.GetMetadataForProperty`, so let's forget about them.

I'm working on a PR to support atomic:operations in OpenAPI, which has heavy changes throughout all tests that I don't want to disturb you with. So if feasible, I'd like to get this PR merged first. With some luck, mine will be completed somewhere next week. No pressure, but it would be nice if we can get this one finalized in the next few days.

test/OpenApiNSwagEndToEndTests/ModelValidation/ModelValidationTests.cs
bkoelman: Nice. We don't have a central place for that yet, so I'm fine with the duplication.

test/OpenApiNSwagEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
verdie-g: Maybe on the first file only. Done.

test/OpenApiKiotaEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,582 @@
...
+        errorObject.Source.ShouldNotBeNull();
+        errorObject.Source.Pointer.Should().Be("/data/attributes/email");
+    }
+
+    [Theory]
+    [InlineData(-1)]
+    [InlineData(-0.56)]
+    [InlineData(123.98)]
+    [InlineData(124)]
+    public async Task Cannot_use_double_outside_of_valid_range(int age)
```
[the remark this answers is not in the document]
verdie-g: Oh that's sad that values are silently casted to int :/

test/OpenApiKiotaEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,582 @@
...
+
+        ErrorObject errorObject = document.Errors.First();
+        errorObject.Title.Should().Be("Input validation failed.");
+        errorObject.Detail.Should().Be("The field Age must be between 0.1 and 122.9.");
+        errorObject.Source.ShouldNotBeNull();
+        errorObject.Source.Pointer.Should().Be("/data/attributes/age");
+    }
+
+    [Fact]
+    public async Task Cannot_use_invalid_url()
```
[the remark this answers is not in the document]
verdie-g: Yes, with NSwag ProfilePicture is a `System.Uri`. Here we get an opportunity to test the server-side url validation

test/OpenApiKiotaEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,582 @@
...
+
+        SocialMediaAccountPostRequestDocument requestBody = new()
+        {
+            Data = new SocialMediaAccountDataInPostRequest
+            {
+                Type = SocialMediaAccountResourceType.SocialMediaAccounts,
+                Attributes = new SocialMediaAccountAttributesInPostRequest
+                {
+                    LastName = newAccount.LastName,
+                    Tags = Enumerable.Repeat("-", length).ToList()
```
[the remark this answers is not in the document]
verdie-g: Tags is a List with Kiota and an ICollection with NSwag. I can change them both to ToList for consistency.

test/OpenApiKiotaEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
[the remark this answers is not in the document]
verdie-g: Reading your other comments, this doesn't JADNC so I'll remove it.

test/OpenApiTests/ModelStateValidation/ModelStateValidationFakers.cs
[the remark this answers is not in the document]
verdie-g: I'll do that from now on

test/OpenApiTests/ModelStateValidation/ModelStateValidationTests.cs
```diff
@@ -0,0 +1,333 @@
...
+        "socialMediaAccountAttributesInPostRequest",
+        "socialMediaAccountAttributesInPatchRequest",
+        "socialMediaAccountAttributesInResponse"
+    };
+
+    private readonly OpenApiTestContext<OpenApiStartup<ModelStateValidationDbContext>, ModelStateValidationDbContext> _testContext;
+
+    public ModelStateValidationTests(OpenApiTestContext<OpenApiStartup<ModelStateValidationDbContext>, ModelStateValidationDbContext> testContext)
+    {
+        CultureInfo.CurrentCulture = new CultureInfo("fr-FR");
```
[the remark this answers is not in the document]
verdie-g: Ope, I didn't mean to leave that here.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
```diff
@@ -0,0 +1,90 @@
...
+    [Base64String]
+#endif
+    public string? Password { get; set; }
+
+    [Attr]
+    [Phone]
+    public string? Phone { get; set; }
+
+    [Attr]
+    [Range(0.1, 122.9)]
```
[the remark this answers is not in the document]
verdie-g: Done. It doesn't seem to be supported by the OpenAPI library though.

test/OpenApiKiotaEndToEndTests/ModelStateValidation/ModelStateValidationTests.cs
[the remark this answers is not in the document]
bkoelman: Ah, that explains it. Yes, please do.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
```diff
@@ -0,0 +1,90 @@
...
+using JetBrains.Annotations;
+using JsonApiDotNetCore.Controllers;
+using JsonApiDotNetCore.Resources;
+using JsonApiDotNetCore.Resources.Annotations;
+
+namespace OpenApiTests.ModelStateValidation;
+
+[UsedImplicitly(ImplicitUseTargetFlags.Members)]
+[Resource(ControllerNamespace = "OpenApiTests.ModelStateValidation", GenerateControllerEndpoints = JsonApiEndpoints.Post | JsonApiEndpoints.Patch)]
+public sealed class SocialMediaAccount : Identifiable<Guid>
```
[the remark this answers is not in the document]
bkoelman: The above bullets are based on the assumption that Swashbuckle understands them, which I didn't verify. If that's not the case, there's no point in adding tests, but then please update the overview table with N for OpenAPI.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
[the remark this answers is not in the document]
bkoelman: My comment was based on the assumption that Swashbuckle understands them, which I didn't verify. If that's not the case, there's no point in adding tests, but please update the overview table with a comment they're not supported.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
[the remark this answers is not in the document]
verdie-g: Done for Min/Max.

HashSet is trickier because it's not supported by EF.

The PR is already pretty large, I'll skip the DataType for now.

bkoelman:
That table update isn't correct. The Y/N columns need to move one position to the right.
image

The same applies to: `[AllowedValues]`, `[DeniedValues]`, `[Remote]`, and `[ValidateNever]`.

bkoelman (changes requested):
I've created https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2957 for the missing entries.

verdie-g:
> The Y/N columns need to move.

I used `N` to specify that it's not supported. I can use `-` instead.

test/OpenApiTests/ModelStateValidation/ModelStateValidationTests.cs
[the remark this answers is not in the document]
bkoelman: This still hasn't been fully addressed.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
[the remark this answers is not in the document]
bkoelman: Why were `[MinimumIsExclusive]` and `[MaximumIsExclusive]` fully removed in the latest commit? Please revert, I didn't ask for that. There's value in having them in end-to-end tests. If you insist on removing them, do so correctly by also removing the assertions on `exclusiveMaximum`/`exclusiveMinimum`.

bkoelman (changes requested):
See comments. Not all changes have been applied yet, and unexpected parts got deleted.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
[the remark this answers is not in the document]
verdie-g: >  there's no point in adding tests

I'm receiving some mixed signals here. I can add them back.

verdie-g:
> Both (remove culture and change assertions) haven't been addressed yet.

I'm not able to access the comment to answer but I could you clarify what you are expecting here.

bkoelman:
> > Both (remove culture and change assertions) haven't been addressed yet.
> 
> I'm not able to access the comment to answer but I could you clarify what you are expecting here.

Using the dollar sign and putting literal values in there, so it becomes culture-insensitive.

test/OpenApiTests/ModelStateValidation/SocialMediaAccount.cs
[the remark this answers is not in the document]
bkoelman: To clarify: they don't matter for OpenApiTests, but do provide value in EndToEndTests.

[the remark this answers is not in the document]
verdie-g: Got it, makes sense

bkoelman (approved):
Thanks, this is in great shape now!

Issue #1472, opened 2024-02-20
Benchmark tool


verdie-g:
**Is your feature request related to a problem? Please describe.**

In the [FAQ](https://www.jsonapi.net/getting-started/faq.html), it states
> JsonApiDotNetCore [scales pretty well](https://github.com/json-api-dotnet/PerformanceReports) under high load and/or large database tables

Unfortunately the linked repository was not updated in the last 3 years and is still targeting .NET Core 3.1 which is 5 years old and reached end of support in 2022.

Additionally, this tool was made to confirm that JADNC has a decent response time when reaching 50 QPS (which is kind of low btw) but it doesn't allow the developer to easily test that a local/recent change improve/worsen the performance.

Finally, the tool runs the JADNC in a docker container which makes profiling challenging and I'm also concerned about the performance impact of running the app in WSL (on Windows).

**Describe the solution you'd like**

These 3 issues can be easily fix by:
1. moving the repository PerformanceReports in the main repo which would enforce the maintainers to keep the benchmarked project up-to-date and that project could also directly reference (with a ProjectReference) JADNC to test a local/recent change.
2. running the benchmarked project outside of Docker to avoid any overhead from WSL (on Windows) and also easily profile the application using dotnet-trace, DotTrace, perf, ...

**Describe alternatives you've considered**

The solution proposed will help to easily measure the change of performance but because it doesn't give any internal metrics about the application, it will not help to find bottlenecks easily. One alternative is to use a Grafana stack:
- [k6](https://grafana.com/products/cloud/k6) for load testing
- [Mimir](https://grafana.com/oss/mimir) as a metrics database
- [Grafana](https://grafana.com/grafana) for visualization

and the app would push metrics (ASP.NET Core, CLR, etc.) to Mimir so after that a load test is run, one could get some hints about where the bottlenecks are.

bkoelman:
Thanks for the proposal. It's great to hear you have experience with this. I'm looking forward to what you can come up with.

[...]

PR #1486, opened 2024-02-27
Optimize query string parsing

verdie-g:
#### QUALITY CHECKLIST
- [x] Adapted tests
- [ ] Documentation updated

While profiling the query string parsing, 40% of the time was spent in the GC 😬 so I focused on optimizing allocations.

Before
| Method  | argument            | Mean      | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|-------- |-------------------- |----------:|----------:|----------:|-------:|-------:|----------:|
| ReadAll | filt(...)name [121] | 12.347 us | 0.0965 us | 0.2099 us | 0.0610 | 0.0153 |  19.92 KB |

After
| Method  | argument            | Mean      | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|-------- |-------------------- |----------:|----------:|----------:|-------:|-------:|----------:|
| ReadAll | filt(...)name [121] | 10.014 us | 0.1852 us | 0.4143 us | 0.0458 | 0.0153 |  15.84 KB |

You can find the result of each commit in their message.

### DIFF
--- a/src/JsonApiDotNetCore.Annotations/Configuration/ResourceType.cs
+++ b/src/JsonApiDotNetCore.Annotations/Configuration/ResourceType.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using JetBrains.Annotations;
 using JsonApiDotNetCore.Resources.Annotations;
 
@@ -11,6 +12,8 @@ public sealed class ResourceType
 {
     private readonly Dictionary<string, ResourceFieldAttribute> _fieldsByPublicName = [];
     private readonly Dictionary<string, ResourceFieldAttribute> _fieldsByPropertyName = [];
+    private readonly ConcurrentDictionary<string, IReadOnlySet<AttrAttribute>> _attributesInTypeOrDerivedByPublicName = [];
+    private readonly ConcurrentDictionary<string, IReadOnlySet<RelationshipAttribute>> _relationshipsInTypeOrDerivedByPublicName = [];
     private readonly Lazy<IReadOnlySet<ResourceType>> _lazyAllConcreteDerivedTypes;
 
     /// <summary>
@@ -259,10 +262,10 @@ public sealed class ResourceType
 
     internal IReadOnlySet<AttrAttribute> GetAttributesInTypeOrDerived(string publicName)
     {
-        return GetAttributesInTypeOrDerived(this, publicName);
+        return _attributesInTypeOrDerivedByPublicName.GetOrAdd(publicName, GetAttributesInTypeOrDerived, this);
     }
 
-    private static IReadOnlySet<AttrAttribute> GetAttributesInTypeOrDerived(ResourceType resourceType, string publicName)
+    private static IReadOnlySet<AttrAttribute> GetAttributesInTypeOrDerived(string publicName, ResourceType resourceType)
     {
         AttrAttribute? attribute = resourceType.FindAttributeByPublicName(publicName);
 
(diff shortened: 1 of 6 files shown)


src/JsonApiDotNetCore.Annotations/Configuration/ResourceType.cs
```diff
@@ -11,6 +12,8 @@ public sealed class ResourceType
 {
     private readonly Dictionary<string, ResourceFieldAttribute> _fieldsByPublicName = [];
     private readonly Dictionary<string, ResourceFieldAttribute> _fieldsByPropertyName = [];
+    private readonly ConcurrentDictionary<string, IReadOnlySet<AttrAttribute>> _attributesInTypeOrDerivedByPublicName = [];
+    private readonly ConcurrentDictionary<string, IReadOnlySet<RelationshipAttribute>> _relationshipsInTypeOrDerivedByPublicName = [];
```
verdie-g, line 16: It's a bit sad to use a `ConcurrentDictionary` here but I couldn't find the right way to get all public names to eagerly fill those caches.

bkoelman:
I don't mean to discourage you, but these micro-optimizations do not make much sense to me. Aside from the logging (2ms), we're talking about a max gain of 1ms, which is neglectable in a typical API request involving a network call to a database server. We're not CoreCLR.

Also, the benchmarks in the codebase are "whatever somebody was once curious about", like a playground. It's not a suite of scenarios that represent typical usage. The PerformanceReports repo is better aligned with that.

For logging, its source generator should be used instead, which avoid formatting the message. It's on my list of potential performance improvements. Though it's low priority, compared to other candidates with likely much higher gain. One of the biggest is to look into `dynamic` and reflection usage, replacing them with cached precompiled delegates.

For `Get[Attributes/Relationships]InTypeOrDerived`, it could be evaluated lazily instead of using a concurrent dictionary. That optimizes throughput at the cost of higher startup time, which is a good trade-off.

I'm not against replacing `.First()` with `[0]` when we know there's at least one element, or `.Any()` with `.Count == 0`, but the codebase of full of them, so fixing just one isn't valuable. Again, this is an improvement barely measurable. I'd rather focus on more important things.

The same for `dictionary.TryGetValue`: unless it's reasonable to assume there may be thousands of entries, it doesn't pay off. Inside Dictionary, there's a threshold where it switches from an array to a tree. As array lookups are vectorized these days, the change may not even result in a measurable gain.

verdie-g:
> I don't mean to discourage you, but these micro-optimizations do not make much sense to me. Aside from the logging (2ms), we're talking about a max gain of 1ms, which is neglectable in a typical API request involving a network call to a database server. We're not CoreCLR.

The optimizations made don't target the execution time but the pressure the code puts on the GC. That JADNC middleware execution time will always be negligible compared to the SQL query. However, by allocating too much you can very easily impact the tail latencies of an application and allocating 20KB to parse 100 bytes string is **huge**. This is what I'm targeting here, and it easily reduces the allocations by 20% of the query parsing.

> For Get[Attributes/Relationships]InTypeOrDerived, it could be evaluated lazily instead of using a concurrent dictionary. That optimizes throughput at the cost of a higher latency for the first request, which is a good trade-off.

I actually wanted to eagerly evaluate them but doing it lazily is fine too. The proposed implementation using the ConcurrentDictionary is doing that.

> I'm not against replacing .First() with [0] when we know there's at least one element, or .Any() with .Count == 0, but the codebase of full of them, so fixing just one isn't valuable. Again, this is an improvement barely measurable. I'd rather focus on more important things.

For those it just hurts me to see an enumerator allocated just to get a length and they are so easy to fix.

> The same for dictionary.TryGetValue

Same idea, it was in the way, it doesn't save an allocation but it was easy to fix.

> Inside Dictionary, there's a threshold where it switches from an array to a tree

I double-checked but I don't think there is such complex algorithm in [Dictionary](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs). It's a simple hash table algorithm.

verdie-g:
Though I agree I'm definitely not taking the problem by the right end. The steps were just:
1. Oh there is actually a benchmark project?
2. 20KB allocated that sounds off
3. Easy quick wins

I haven't got the time to look at #1472 but I will some day.

bkoelman:
Ok, so the 20 KB allocation is reduced down to 16 KB, saving 4 KB. As I said, there are many things more important to worry about. JADNC is stateless, it's easy to scale up extra instances. We never had performance complaints, so working on this is not a priority. It requires effort from both of us that is better spent on things that users _do_ care about.

ConcurrentDictionary lookups are slower than lookups from a regular dictionary, so what you did is not the same. But I don't want to even discuss that.

I'm not going to review this PR, so I suggest you close it and not waste time on it.

PR #1487, opened 2024-02-28
Document PR workflow

bkoelman:
This PR documents the responsibilities and mutual expectations for a smooth PR review workflow.


bkoelman:
@verdie-g This is what worked well between Maurits and me. Any thoughts?

Issue #1046, opened 2021-09-01
Umbrella: OpenAPI integration


maurei:
# OpenAPI integration

This document provides an overview of the work items involved in creating an OpenAPI integration for JsonApiDotNetCore.

> Disclaimer: this is an open source project. The available time of our contributors varies and therefore we do not plan release dates. This document expresses our current intent, which may change over time.

## Approach

Creating an OpenAPI integration for JADNC requires setting up a mechanism that generates an [OpenAPI Specification (OAS) file](https://swagger.io/specification/) from application code. This file allows for automatic generation of API docs (see eg. [SwaggerUI](https://editor.swagger.io/)) and client code (see eg. [NSwag code generator](https://www.c-sharpcorner.com/article/generate-client-code-with-nswag-studio/)). In the design choices for this mechanism precedence will be given to compatibility with client code generation tools because we believe this to be of the biggest impact to the JSON:API community (see eg. [this issue](https://discuss.jsonapi.org/t/api-description-language-client-library-generator-related-top-level-document-keys-spec/172) and [others](https://discuss.jsonapi.org/search?q=api%20client)).

In a private repository we have experimented with integrating [Swashbuckles SwaggerGen](https://github.com/domaindrivendev/Swashbuckle.AspNetCore#swashbuckleaspnetcoreswaggergen) in a JADNC application. We have decided to extract this code into this repository and move forward from there.

## Work items

The order of the list below reflects the order in which we intend to address the work items.

- Will follow shortly ...

## Feedback

The best way to give feedback is to create new issues or upvote/downvote existing ones.
Please give us feedback that will give us insight on the following points:

* Existing features that are missing some capability or otherwise don't work well enough.
* Missing features that should be added to the product.
* Design choices for a feature that is currently in-progress.

wunki:
This is so exciting! This will catapult JADNC to the next level 🚀

sharky98:
> For anyone that wants to try out the latest bits: pick a version with "openapi" in the name from https://github.com/json-api-dotnet/JsonApiDotNetCore#trying-out-the-latest-build.

Hello, I wanted to try this out, however I am facing a bug, as shown below.

[...]

PR #1518, opened 2024-03-27
OpenAPI: Hide inaccessible links

### LINKED ISSUE #1062: Generalized handling of link objects
The current state of the OpenApi integration has a hardcoded configuration for where and when links are included in schemas.

In JADNC this can be configured by:
-  ~toggling `JsonApiOptions.DefaultPageSize` (enabling/disabling paging links)~
   - This wouldn't be correct. Even when pagination is turned off by default, it can still be activated per request using the `page[size]` query string parameter.
-  `LinkTypes` cascading configuration

These settings should be taken into account when links are included or excluded in the generation of schemas.

bkoelman:
This PR updates OpenAPI to take into account whether JSON:API links are available.

My first take on implementing this was to declare nearly all links as required in the component schemas. Then at runtime remove the links we know can't be returned, which can be determined from the cascading links configuration.

However, because links can be configured per resource type, this requires expansion of the link schema types. For example, `linksInResourceData` becomes `linksInTodoItemData`, `linksInPersonData`, `linksInTagData`, etc. This introduces many new schemas for questionable gain, as using OpenAPI typically makes users care less about links. But it gets worse. Relationship links can be overruled per relationship within a resource type. This means we need to expand link schema types for them as well. But the relationship links are defined on the relationship left-type, whereas the currently produced schemas for relationships reflect their right-type. For example, `todoItems` contains relationships `owner` and `assignee`, both of type `people`. Let's assume both relationships are required. In that case, a single schema `toOnePersonInResponse` is produced for both relationships' right-side types, because they have the same data structure. Now taking expansion for relationship links into account, we'd need to generate separate schemas: `ownerToOnePersonInResponse` and `assigneeToOnePersonInResponse`. As a result, consumers can't write the following code anymore, because different schemas would be produced:
```c#
var todoItemsResponse = await _apiClient.GetTodoItemCollectionAsync(null, null);
var firstTodoItem = todoItemsResponse.Result.Data.First();

PrintPersonRelationship(firstTodoItem.Relationships.Owner);
PrintPersonRelationship(firstTodoItem.Relationships.Assignee);

private void PrintPersonRelationship(ToOnePersonInResponse person)
{
    // ....
}
```

For the reasons mentioned above, I chose not to implement schema expansion at all. The next best thing is to determine which links can be eliminated by taking a union of the configured link types. So if all links are turned off, but `todoItems` turns on the top-level `self` link, then _all_ resource types get only the top-level `self` link. Likewise, if the `owner` relationship turns on the `related` relationship-level link, while the `assignee` relationship turns on the `self` link, then all relationship schemas get both links. If no link overrules are used and links are turned off in options, then no links appear in the component schemas.

As a result, all links must be declared as non-required in the component schemas, because now the schemas may contain links that won't actually be returned at runtime (and we haven't eliminated them from the schema).

Closes #1062.

#### QUALITY CHECKLIST
- [x] Adapted tests


xontab:
@bkoelman Will this feature be released soon?

bkoelman:
This is available for use from the GitHub NuGet feed, see https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/README.md#trying-out-the-latest-build. The overall progress of adding OpenAPI support is tracked at #1046. It'll take a while to complete that, after which it'll be published to NuGet. To speed that up, you can help if you want by working on open issues.

PR #1550, opened 2024-05-25
Multi-target against .NET 9 RC2

bkoelman:
Explores multi-targeting against .NET 9 with EF Core 9 pre-release versions. Based on #1619.

> [!CAUTION]
> This PR is experimental and not intended to be merged.

Blocked by:
- https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1841
- ~https://github.com/NuGet/Home/issues/13855~
- ~https://github.com/dotnet/efcore/issues/34728~
- ~https://youtrack.jetbrains.com/issue/RSRP-498478/Breaking-primary-constructors-with-readonly-field~
- ~https://github.com/dotnet/roslyn-analyzers/issues/7421~

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "jetbrains.resharper.globaltools": {
-      "version": "2024.2.7",
+      "version": "2024.3.0",
       "commands": [
         "jb"
       ],
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,6 +48,7 @@ jobs:
         dotnet-version: |
           6.0.x
           8.0.x
+          9.0.x
     - name: Show installed versions
       shell: pwsh
       run: |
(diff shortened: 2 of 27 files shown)


wayne-o:
Hi - is there a preview Nuget of this anywhere we can try against our own code?

bkoelman:
I don't think there a need for that. EF Core 9 targets .NET 8, so the current JADNC version should work already with EF Core 9 on both .NET 8 and 9.

bkoelman:
I've created #1619 to verify all tests are green when running on .NET 9 RC2 with EF Core 9 RC2, while still referencing JADNC that targets only .NET 8 and 6.

While there are breaking changes in the EF Core driver API (so you'll need matching versions for PostgreSQL, MySQL, SQL Server, etc.), it appears that JADNC itself is already compatible. This means we don't need to release a new version of JADNC when .NET 9 ships.

bkoelman:
Superseded by #1656. We currently have no need to multi-target the JsonApiDotNetCore library itself against .NET 9.

PR #1619, opened 2024-10-14
Run tests and examples against .NET 9

bkoelman:
Demonstrates that all tests are green when running on .NET 9 with EF Core 9, while still referencing JADNC that targets only .NET 8 and 6.

While there are breaking changes in the EF Core driver API (so you'll need matching EF Core versions for PostgreSQL, MySQL, SQL Server, etc.), the current stable version of JADNC is already compatible. This means there's no need to release a new version.

#### QUALITY CHECKLIST
- [x] Adapted tests

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "jetbrains.resharper.globaltools": {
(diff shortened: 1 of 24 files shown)


bkoelman:
Superseded by #1656.

Issue #1511, opened 2024-03-20
Source generation when models are in a separate project


wayne-o:
We have the following structure:

 - CompanyName.Models (only has reference to JADNC.Annotations)
     - Model.cs
     - DbContext.cs
     - Migrations
     - etc etc
 - CompanyName.Web (has reference to full JADNC package)
     - Program.cs

I was hoping that the source gen would happen in CompanyName.Web since it has a reference to CompanyName.Models but that doesn't seem to be the case?

We don't want to have to add a reference to the full package in the Models project and we would prefer to not pulute that project with controllers. Is there any way around this?

bkoelman:
Understandable. This is a limitation of using source generators. They operate at the project level. I'm not aware of a way to fix that.

In your case, it requires to turn it off and define controllers manually.

PR #1676, opened 2025-02-13
Note single project req. in common pitfalls

bjornharrtell:
Might be worth mentioning that source code generation doesn't work if the code is split out into separate project, as noted in https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/1511.

### DIFF
--- a/docs/usage/common-pitfalls.md
+++ b/docs/usage/common-pitfalls.md
@@ -143,3 +143,7 @@ A better way is to inject your `DbContext` in a [Resource Definition](~/usage/ex
 While we'd love to support such [tokens for optimistic concurrency](https://learn.microsoft.com/en-us/ef/core/saving/concurrency?tabs=data-annotations),
 it turns out that the implementation is far from trivial. We've come a long way, but aren't sure how it should work when relationship endpoints and atomic operations are involved.
 If you're interested, we welcome your feedback at https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/1119.
+
+#### Source generation features doesn't work on code in separated projects
+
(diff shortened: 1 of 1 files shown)


docs/usage/common-pitfalls.md
```diff
@@ -143,3 +143,7 @@ A better way is to inject your `DbContext` in a [Resource Definition](~/usage/ex
 While we'd love to support such [tokens for optimistic concurrency](https://learn.microsoft.com/en-us/ef/core/saving/concurrency?tabs=data-annotations),
 it turns out that the implementation is far from trivial. We've come a long way, but aren't sure how it should work when relationship endpoints and atomic operations are involved.
 If you're interested, we welcome your feedback at https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/1119.
+
+#### Source generation features doesn't work on code in separated projects
+
+See https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/1511.
```
bkoelman, line 149: Thanks for the suggestion. This is already documented at https://www.jsonapi.net/usage/extensibility/controllers.html#auto-generated-controllers, which we should link to. Please move this topic after "Do not use [ApiController] on JSON:API controllers".

Suggested header name:
```
Don't use auto-generated controllers with shared models
```
Suggested text:
```
When model classes are defined in a separate project, the controllers are generated in that project as well, which is probably not what you want. For details, see [here](~/usage/extensibility/controllers.html#auto-generated-controllers).
```

bkoelman:
The master branch now contains a fix for the broken build, please merge or rebase this PR on top of that.

bkoelman:
Superseded by #1679.

bjornharrtell:
Thanks @bkoelman 🙂

PR #1785, opened 2025-08-31
Add tests for non-primitive id

verdie-g:
Closes #{ISSUE_NUMBER}

#### QUALITY CHECKLIST
- [ ] Complies with our [contributing guidelines](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/.github/CONTRIBUTING.md)
- [ ] Adapted tests
- [ ] Documentation updated

Hi @bkoelman, I'm trying to have non-primitive ids but it's not working as expected. The use-case is to display a UUID in a compact form using base 58. The Guid is wrapped in a custom struct to do that. That wrapper should be used as far as possible until it hits the db. Using JADNC, creating a resource returns a 204 even though I'm not using a client-generated id. Any idea?

### DIFF
--- a/test/JsonApiDotNetCoreTests/IntegrationTests/IdCompaction/CompactGuid.cs
+++ b/test/JsonApiDotNetCoreTests/IntegrationTests/IdCompaction/CompactGuid.cs
@@ -1,0 +1,251 @@
+using System.Buffers;
+using System.Buffers.Text;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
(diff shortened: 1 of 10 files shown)


verdie-g:
This is not a JADNC issue but how values are generated with EF. The fix is to do a `property.SetDefaultValueSql("gen_random_uuid")`. So not sure it's worth adding tests in JADNC.

bkoelman:
This looks very similar to [ID Obfuscation](https://www.jsonapi.net/usage/advanced/id-obfuscation.html). Thanks for bringing it up (and for finding the fix). I suspect the fix varies per database provider, so I don't think it's worth adding it to the codebase.

PR #1879, opened 2025-12-10
Replace int with long for ID

### LINKED ISSUE #1876: Replace `int` with `long` as ID type in samples/docs/tests
This project originally used `int` (32-bit) as the ID type in resources. While that's not wrong, I believe we should default to `long` (64-bit) in docs, samples and most tests. With many database transactions, the 2 billion limit can be reached earlier than expected, requiring a costly database migration and possible outage. A 64-bit auto-incrementing number doesn't have this limitation. Users can always choose their own ID type, but for anyone who doesn't care, this sounds like the safer default.

We should keep a few tests around, though, that still use `int`, so it remains covered. Models in `JsonApiDotNetCoreTests/IntegrationTests/ReadWrite` and `JsonApiDotNetCoreTests/IntegrationTests/AtomicOperations` should not be updated, as they cover scenarios with varying ID types.

The change is basically from, for example:
```c#
public sealed class TelevisionBroadcast : Identifiable<int>
```
to:
```c#
public sealed class TelevisionBroadcast : Identifiable<long>
```
Likewise, tests that parse the ID should be updated accordingly, from:
```c#
int newShopId = int.Parse(responseDocument.Data.SingleValue.Id);
```
to:
```c#
long newShopId = long.Parse(responseDocument.Data.SingleValue.Id);
```
As well as usages to unknown IDs in tests, such as from:
```c#
int id = Unknown.StringId.Int32;
```
to:
```c#
long id = Unknown.StringId.Int64;
```

niteshsinghal85:
Closes #1876 

#### QUALITY CHECKLIST
- [x] Adapted tests
- [x] Documentation updated

### DIFF
--- a/docs/usage/errors.md
+++ b/docs/usage/errors.md
@@ -34,9 +34,9 @@ This handler is also the place to choose the log level and message, based on the
 ```c#
 public class ProductOutOfStockException : Exception
 {
-    public int ProductId { get; }
+    public long ProductId { get; }
 
-    public ProductOutOfStockException(int productId)
+    public ProductOutOfStockException(long productId)
     {
         ProductId = productId;
     }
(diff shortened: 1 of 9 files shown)


niteshsinghal85:
> Thanks, looks great!
> 
> There are still a few places not updated in /docs. Can you search for "int" (match whole word) and update accordingly?

Updated for relevant places

PR #1894, opened 2025-12-29
Add integration test for offloading requests to background process

### LINKED ISSUE #1727: Create integration test for offloading requests to a background process
Create an integration test for the scenario described at https://www.jsonapi.net/getting-started/faq.html#can-i-offload-requests-to-a-background-process.

snaanouh:
## Summary
   This PR adds an integration test for the scenario described in the FAQ about offloading requests to a background process.

   Resolves #1727

   ## Changes
   - Added `WorkItem` resource model
   - Added `BackgroundJobDbContext` for database context
   - Added `WorkItemsController` with custom POST method returning HTTP 202 Accepted with Location header
   - Added `BackgroundProcessingTests` with tests for:
     - Returning 202 Accepted with Location header when offloading to background
     - Retrieving result from Location after background processing
     - Capturing query string when offloading
     - Returning 404 when result not yet available

   ## Related
   - FAQ: https://www.jsonapi.net/usage/faq.html#can-i-offload-requests-to-a-background-process
   - Reference PR: #1144

### DIFF
--- a/test/JsonApiDotNetCoreTests/IntegrationTests/BackgroundProcessing/BackgroundJobDbContext.cs
+++ b/test/JsonApiDotNetCoreTests/IntegrationTests/BackgroundProcessing/BackgroundJobDbContext.cs
@@ -1,0 +1,16 @@
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore;
+using TestBuildingBlocks;
+
+namespace JsonApiDotNetCoreTests.IntegrationTests.BackgroundProcessing;
+
+[UsedImplicitly(ImplicitUseTargetFlags.Members)]
+public sealed class BackgroundJobDbContext : TestableDbContext
+{
+    public DbSet<WorkItem> WorkItems => Set<WorkItem>();
+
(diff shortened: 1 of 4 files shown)


bkoelman:
Hi @snaanouh,

Thanks for trying, but this doesn't quite represent the intent of the underlying open issue. I don't think this is something for external community members to work on unless they are already familiar with the internals of JsonApiDotNetCore.

snaanouh:
Hi @bkoelman,

Thanks for the feedback! I'm new to the project and would like to learn.

Could you please explain what specific aspects of the background processing scenario should be tested? I'd be happy to update the PR based on your guidance.

bkoelman:
I appreciate that, but as I said, this is not a beginner-level task. My time is currently limited, and I haven't thought much yet about how best to implement this. I don't want to waste your time giving you wrong guidance. The user I originally created this for is highly experienced with this project, so the partial implementation I provided was sufficient to pick up from.

We currently don't have any beginner-level open issues that aren't already being worked on.

PR #2003, opened 2026-05-12
Option to use PG row count estimation

### LINKED ISSUE #2002: Optionally use PostgreSQL row estimates
IncludeTotalResourceCount can be expensive since it will cause full table scan at least in PostgreSQL. PostgreSQL has row estimates that usually are correct and might work well for some uses. I would like suggest an option that will make use of the PostgreSQL estimated rows instead of real count.

bjornharrtell:
Closes #2002

#### QUALITY CHECKLIST
- [x] Adapted tests
- [ ] Documentation updated

### DIFF
--- a/src/JsonApiDotNetCore/Repositories/EntityFrameworkCoreRepository.cs
+++ b/src/JsonApiDotNetCore/Repositories/EntityFrameworkCoreRepository.cs
@@ -37,6 +37,7 @@ public class EntityFrameworkCoreRepository<TResource, TId> : IResourceRepository
     private readonly IResourceFactory _resourceFactory;
     private readonly IQueryConstraintProvider[] _constraintProviders;
     private readonly IResourceDefinitionAccessor _resourceDefinitionAccessor;
+    private readonly IJsonApiOptions _options;
     private readonly TraceLogWriter<EntityFrameworkCoreRepository<TResource, TId>> _traceWriter;
 
     /// <inheritdoc />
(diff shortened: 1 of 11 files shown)


src/JsonApiDotNetCore/Repositories/EntityFrameworkCoreRepository.cs
```diff
@@ -37,14 +37,15 @@ public class EntityFrameworkCoreRepository<TResource, TId> : IResourceRepository
...
+    private readonly IJsonApiOptions _options;
     private readonly TraceLogWriter<EntityFrameworkCoreRepository<TResource, TId>> _traceWriter;
 
     /// <inheritdoc />
     public virtual string? TransactionId => _dbContext.Database.CurrentTransaction?.TransactionId.ToString();
 
     public EntityFrameworkCoreRepository(ITargetedFields targetedFields, IDbContextResolver dbContextResolver, IResourceGraph resourceGraph,
         IResourceFactory resourceFactory, IEnumerable<IQueryConstraintProvider> constraintProviders, ILoggerFactory loggerFactory,
-        IResourceDefinitionAccessor resourceDefinitionAccessor)
+        IResourceDefinitionAccessor resourceDefinitionAccessor, IJsonApiOptions options)
```
bkoelman, line 48: This breaking change is only permitted in a major release, according to our [versioning policy](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/VERSIONING_POLICY.md#minor-updates).

Aside from logging, the existing `CountAsync` method contains only 4 statements, which can be easily copied into a third-party-derived class. And JADNC is fully database-agnostic; adding a Postgres-specific setting breaks that.

Overall, this doesn't meet the bar to introduce an extensibility point.

PR #2005, opened 2026-05-14
Add virtual CanIncludeTotalResourceCount to JsonApiResourceService

### LINKED ISSUE #2002: Optionally use PostgreSQL row estimates

bjornharrtell:
Closes #2002

## Problem

`IncludeTotalResourceCount = true` applies globally. There was no per-resource way to disable the total-count query, which on PostgreSQL with large tables is O(table) due to MVCC.

## Solution

Adds `protected virtual bool CanIncludeTotalResourceCount()` to `JsonApiResourceService<TResource, TId>`. The default implementation returns `_options.IncludeTotalResourceCount`, so behavior is unchanged for all existing consumers.

A derived service can override it to conditionally disable counting based on `TResource`, request context, a custom query string parameter, or any other criteria:

```csharp
// Register:
services.AddResourceService<MyArticleService>();

// Implementation:
public sealed class MyArticleService(/* same ctor params */)
    : JsonApiResourceService<Article, long>(/* ... */)
{
    protected override bool CanIncludeTotalResourceCount()
    {
        // Disable count for this resource type regardless of global option
        return false;
    }
}
```

For the estimate use case (returning a cheaper approximation instead of an exact `COUNT(*)`), the existing `virtual CountAsync` on `EntityFrameworkCoreRepository` is the right seam — override it in a custom repository to execute an `EXPLAIN`-based estimate and return the planner row count.

## Changes

- `src/JsonApiDotNetCore/Services/JsonApiResourceService.cs` — adds `protected virtual bool CanIncludeTotalResourceCount()`; the three count guard conditions now call it instead of `_options.IncludeTotalResourceCount` directly
- `test/JsonApiDotNetCoreTests/IntegrationTests/Meta/OverrideTotalResourceCountTests.cs` — integration test: overrides `CanIncludeTotalResourceCount()` to return `false` while the global option is `true`, asserts no `meta.total` in the response


bjornharrtell:
@bart-degreed do you think this is an acceptable extension point?

src/JsonApiDotNetCore/Services/JsonApiResourceService.cs
```diff
@@ -24,19 +24,21 @@ public class JsonApiResourceService<TResource, TId> : IResourceService<TResource
...
     private readonly IJsonApiOptions _options;
     private readonly TraceLogWriter<JsonApiResourceService<TResource, TId>> _traceWriter;
     private readonly IJsonApiRequest _request;
     private readonly IResourceChangeTracker<TResource> _resourceChangeTracker;
     private readonly IResourceDefinitionAccessor _resourceDefinitionAccessor;
 
     public JsonApiResourceService(IResourceRepositoryAccessor repositoryAccessor, IQueryLayerComposer queryLayerComposer, IPaginationContext paginationContext,
-        IJsonApiOptions options, ILoggerFactory loggerFactory, IJsonApiRequest request, IResourceChangeTracker<TResource> resourceChangeTracker,
-        IResourceDefinitionAccessor resourceDefinitionAccessor)
+        IResourceCountStrategy resourceCountStrategy, IJsonApiOptions options, ILoggerFactory loggerFactory, IJsonApiRequest request,
```
bkoelman: Same concerns as in https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/2003#pullrequestreview-4304519163.

If the _only_ reason to move this from the repository to the resource service layer is so you can override `options.IncludeTotalResourceCount`, then it would make more sense to introduce a virtual `CanIncludeTotalResourceCount` method here. A third-party derived type can then override its logic based on `TResource`, a custom query string parameter, or another source.

bkoelman:
> @bart-degreed do you think this is an acceptable extension point?

I don't use that account anymore, just saw the email by accident. Please use @bkoelman instead.

src/JsonApiDotNetCore/Services/JsonApiResourceService.cs
bjornharrtell: I see, and I agree. I've reworked this PR to implement accordingly.

src/JsonApiDotNetCore/Services/JsonApiResourceService.cs
```diff
@@ -212,6 +212,16 @@ public virtual async Task<TResource> GetAsync([DisallowNull] TId id, Cancellatio
         return rightValue;
     }
 
+    /// <summary>
+    /// Indicates whether the total resource count should be retrieved and included in the response. This is called for paginated collection endpoints when
+    /// <see cref="IJsonApiOptions.IncludeTotalResourceCount" /> is <c>true</c>. Override this method to conditionally disable counting for
+    /// <typeparamref name="TResource" />, for example based on the request context.
+    /// </summary>
+    protected virtual bool CanIncludeTotalResourceCount()
```
bkoelman:
```suggestion
    protected virtual bool CanIncludeTotalResourceCount()
```

I prefer not to add comments describing the implementation, as they can quickly become outdated. Anyone considering overriding such a protected method should look at how it is used in the code.

test/JsonApiDotNetCoreTests/IntegrationTests/Meta/OverrideTotalResourceCountTests.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        _testContext = testContext;
+
+        testContext.UseController<SupportTicketsController>();
+
+        testContext.ConfigureServices(services =>
+        {
+            services.AddResourceService<CountDisabledSupportTicketService>();
+            services.AddScoped(typeof(IResourceChangeTracker<>), typeof(NeverSameResourceChangeTracker<>));
+        });
```
bkoelman:
```suggestion
        testContext.ConfigureServices(services => services.AddResourceService<CountDisabledSupportTicketService>());
```

`NeverSameResourceChangeTracker` is used in post/patch tests to ensure a response body is always returned. It's redundant here.

bkoelman (changes requested):
Thanks! Minor feedback, but basically looks good.

bjornharrtell:
@bkoelman adressed those minor things so hopefully read to go now :)

bkoelman (approved):
LGTM, thanks!

PR #2017, opened 2026-05-27
Remove Microsoft.Extensions.Caching.Memory reference since it is platform provided

nkolev92:
Follow up to https://devblogs.microsoft.com/dotnet/nuget-package-pruning-in-dotnet-10/?commentid=22975#comment-22975.

#### QUALITY CHECKLIST
- [ ] ~Changes implemented in code~
- [x] ~Adapted tests~ Existing tests should pass just fine. Not audit issues reported.
- [ ] ~Documentation updated~

### DIFF
--- a/src/JsonApiDotNetCore.OpenApi.Swashbuckle/JsonApiDotNetCore.OpenApi.Swashbuckle.csproj
+++ b/src/JsonApiDotNetCore.OpenApi.Swashbuckle/JsonApiDotNetCore.OpenApi.Swashbuckle.csproj
@@ -33,6 +33,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <!-- Workaround: Microsoft.EntityFrameworkCore 8.0.0 depends on Microsoft.Extensions.Caching.Memory 8.0.0, which is vulnerable. -->
     <FrameworkReference Include="Microsoft.AspNetCore.App" Condition="'$(TargetFramework)' == 'net8.0'" />
   </ItemGroup>
   
--- a/src/JsonApiDotNetCore.OpenApi.Swashbuckle/JsonApiDotNetCore.OpenApi.Swashbuckle.csproj
+++ b/src/JsonApiDotNetCore.OpenApi.Swashbuckle/JsonApiDotNetCore.OpenApi.Swashbuckle.csproj
@@ -33,6 +33,10 @@
   </ItemGroup>
 
   <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+  
+  <ItemGroup>
     <PackageReference Include="SauceControl.InheritDoc" Version="$(InheritDocVersion)" PrivateAssets="All" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="$(SwashbuckleFrozenVersion)" />
   </ItemGroup>
--- a/src/JsonApiDotNetCore.OpenApi.Swashbuckle/JsonApiDotNetCore.OpenApi.Swashbuckle.csproj
+++ b/src/JsonApiDotNetCore.OpenApi.Swashbuckle/JsonApiDotNetCore.OpenApi.Swashbuckle.csproj
@@ -33,7 +33,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <FrameworkReference Include="Microsoft.AspNetCore.App" Condition="'$(TargetFramework)' == 'net8.0'" />
   </ItemGroup>
   
   <ItemGroup>
--- a/src/JsonApiDotNetCore/JsonApiDotNetCore.csproj
+++ b/src/JsonApiDotNetCore/JsonApiDotNetCore.csproj
@@ -40,12 +40,6 @@
     <PackageReference Include="Humanizer.Core" Version="$(HumanizerFrozenVersion)" />
     <PackageReference Include="Microsoft.EntityFrameworkCore" Version="$(EntityFrameworkCoreFrozenVersion)" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" Version="$(EntityFrameworkCoreFrozenVersion)" />
-    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="8.0.1" Condition="'$(TargetFramework)' == 'net8.0'" NoWarn="$(NoWarn);NU1510">
-      <!--
-        Microsoft.EntityFrameworkCore 8.0.0 depends on Microsoft.Extensions.Caching.Memory 8.0.0, which is vulnerable.
-        This package reference silences the vulnerability warning and suppresses the NU1510 warning that the dependency will not appear in our NuGet package.
-      -->
-    </PackageReference>
     <PackageReference Include="SauceControl.InheritDoc" Version="$(InheritDocVersion)" PrivateAssets="All" />
   </ItemGroup>
 </Project>


bkoelman:
That breaks the build.

nkolev92:
I see the problem. 

The ASP.NET framework ref is what actually provides `Microsoft.Extensions.Caching.Memory` and because it's transitive and to make restore repeatable, the framework refs the SDK uses are the ones specified in the project.

By adding the framework ref explicitly, you get a broader set of prunable packages.
This should now fix the build. 

I'll file some issues to improve this behavior. We had talked about a warning to help identify this scenario at build time. I'll file some issues and improve the documentation.

bkoelman:
Okay, I think that's a reasonable change _for this project_, because `JsonApiDotNetCore.OpenApi.Swashbuckle` depends on `JsonApiDotNetCore`, which already has the framework ref, so we can safely add it there, too. Though it should be redundant, because [`Swashbuckle.AspNetCore`](https://www.nuget.org/packages/Swashbuckle.AspNetCore/10.1.7#dependencies-body-tab) (which our `JsonApiDotNetCore.OpenApi.Swashbuckle` depends on) already [pulls in the framework ref](https://github.com/search?q=repo%3Adomaindrivendev%2FSwashbuckle.AspNetCore%20frameworkreference&type=code), and `JsonApiDotNetCore` pulls that in, too.

Though I wonder what the guidance would be for low-level libraries. For example, a hypothetical YAML parser that uses only [Base64UrlTextEncoder](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.webutilities.base64urltextencoder?view=aspnetcore-10.0), which is defined in the `Microsoft.AspNetCore.WebUtilities` NuGet package. Once that package becomes vulnerable, should it also add a framework ref to `Microsoft.AspNetCore.App`, just to silence the warning? That seems overkill. There are many such general-purpose libraries that should not require a dependency on ASP.NET (think of configuration providers, logging frameworks, database providers and test assertion libraries). Once the framework ref is added, all users are required to install the ASP.NET runtime in production, instead of just the .NET runtime, right? That's a breaking change. It sounds better to me for packages to express the minimal dependencies they require, keeping consumption flexible.

The NU1510 warning that is emitted when I _remove_ the warning suppression:
```xml
<PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="8.0.1"
    Condition="'$(TargetFramework)' == 'net8.0'">
```

Is:
> PackageReference Microsoft.Extensions.Caching.Memory will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary.

This looks questionable, because:
1.  It isn't actually correct, because the reference [_does_ get pruned](https://www.nuget.org/packages/JsonApiDotNetCore/5.11.0#dependencies-body-tab) (the framework ref overrides it).
2. The warning is emitted for both `net8.0` and `net10.0`, while it should apply to `net8.0` only:
   {5678DB7C-AE1D-4C9D-833C-77A72C534E14}

The implication of how this warning works today is that package owners need to choose between:
1. Don't care about vulnerabilities in dependencies. It's the responsibility of consuming apps to decide whether to bump to a non-vulnerable version. Consumers have no way to know whether a higher non-vulnerable version of the dependency is compatible. (uncommon)
2. Remove the dependency and add a framework ref instead (a breaking change). The warning doesn't suggest doing so.
3. Bump to a non-vulnerable version, as I did. The warning must be suppressed. 

I think many package authors choose 3 for good reasons. The warning should be a suggestion because its message reads "likely unnecessary," and low-level libraries are a legitimate use case to suppress it. In general, keeping the reference should be fine, since it will be pruned anyway and is harmless.

I'd love to hear your thoughts on this.

bkoelman:
Perhaps I'm misunderstanding all of this. https://stackoverflow.com/questions/64478618/what-does-frameworkreference-include-microsoft-aspnetcore-app-actually-do, suggests that `FrameworkReference` _only_ pulls in packages without a pinned version. Doing so doesn't require the ASP.NET runtime to be installed. And, apparently, it isn't transitive. So that would mean all low-level libraries could safely add a framework reference (or multiple). But the same link states it results in larger binaries when trimming is used. I'm a bit lost on the implications and how it all works.

I also noticed that nuget.org doesn't list framework dependencies, although they do appear in the `.nuspec` file.

nkolev92:
FrameworkReferences are transitive already. 

So we're not really adding anything new by explicitly specifying the FrameworkReference. 
There's no net change for build/publish by doing that, it's just restore.

bkoelman:
> FrameworkReferences are transitive already.

Then why do I need to add it in `JsonApiDotNetCore.OpenApi.Swashbuckle`, which depends on two libraries that already pull in the `Microsoft.AspNetCore.App` framework reference?

nkolev92:
Transitive framework reference application and pruning are both restore applied. 

Currently, pruning affects framework reference application by virtue of picking packages.
The opposite can't happen cause it may lead to an indeterministic restore in some cases.  

https://github.com/nuget/home/issues/14920 option 1 touches on it.

bkoelman:
Thanks for the clarification and for opening an issue.

Could you please address my concerns in https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/2017#issuecomment-4559662334, especially about low-level libraries? At work, I'm part of a team that builds many of those, and discussions keep coming up about whether we should keep references to System/Microsoft libraries or add framework references instead. Some examples [here](https://github.com/search?q=repo%3ASteeltoeOSS%2FSteeltoe+Include%3D%22System.&type=code) and [here](https://github.com/search?q=repo%3ASteeltoeOSS%2FSteeltoe+Include%3D%22Microsoft.Extensions.&type=code). My point of view has always been to keep these explicit references, but now I'm not so sure anymore. Would it be recommended to replace them all with framework references? Is there any guidance/docs, or are there limitations I should be aware of before doing so?

And are the oddities regarding NU1510 tracked anywhere? Particularly, I find the message "will _not_ be pruned" _very_ confusing. As a developer, my understanding of the concept of pruning is: what will be removed from the set of libraries my project depends on, not what the restore command does internally to validate the graph.

nkolev92:
No, I wouldn't recommend for lower level libraries to add extra framework references. I'd just bump the version then. 
These lower level libraries aren't really depending on the framework provided verison, they are bringing their own. 

I would just say, add an explicit direct framework ref only if you have a framework ref you were getting transitively anyways.

We have improved the wording of NU1510 in a recent release, https://github.com/Nuet/NuGet.Client/pull/7380.

I'm documenting the various scenarios in the NU1510 docs page.

bkoelman:
@nkolev92 Thanks for the clarification. I've pushed a commit to scope the `PackageReference` to only `net8.0`, so I can remove it once .NET 8 reaches end of life.

Regarding https://github.com/NuGet/NuGet.Client/pull/7380, the new message doesn't help at all, because:
1. It still says the package will _not_ be pruned. Which is wrong because the package doesn't appear in the resulting nuspec.
2. It still says to remove the package reference. Which is wrong because then the vulnerability comes back.
3. Maybe, in some cases, the package reference is redundant, but more often it is not. So it should not be a warning, because a warning typically indicates you're doing something wrong.

It seems that warnings produced from `nuget restore` don't align with what's actually happening during build/run/pack. So the need to suppress them remains.

nkolev92:
Thanks for the feedback, we appreciate it.

1. The privatization is something that's intentional to not propagate unnecessary references. But it is not completely pruned in that, it's not removed. I'd argue it's still correct. You may get vulnerability or any other resolver warnings for these packages.
2. For the project in question, it's still correct. There's improvements we can make, but the scenario you are hitting is not the majority case. I do recognize there's room for improvement here though. 
3. In the common scenario, framework refs aren't transitive. They match the direct ones.

I'll try to see if I can make the error message more explicit in these transitive scenarios.

bkoelman:
@nkolev92 When I compare the resulting .nupkg from this PR with the one on nuget.org, it adds the following section:
```xml
    <frameworkReferences>
      <group targetFramework="net10.0" />
      <group targetFramework="net8.0">
        <frameworkReference name="Microsoft.AspNetCore.App" />
      </group>
    </frameworkReferences>
```

That looks like a breaking change to me. So I prefer to stick with the original code, which works and will be removed anyway once .NET 8 goes out of support.

nkolev92:
The transitive package is already bringing in the FrameworkReference, so they're net the same once restored.

It's your choice ofc. Appreciate all your feedback on this topic.

